PR #6, opened 2015-03-27
add modules.

jp311:


maxwellb:
Having modules within this repository may change the spirit of the repository. As is, the project serves as a launch pad and educational resource to explore .NET Core. The goal of the project does not appear to be building the Framework or CLR, and thus stands fine without these additions, imo.

PR #16, opened 2015-09-06
.NET Core roadmap.

richlander:
This is a start.

/cc: @kangaroo, @terrajobst

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -1,11 +1,13 @@
 .NET Core Roadmap
 =================
 
-.NET Core is a general purpose, modular, cross-platform and open source implementation of .NET. The .NET Core team (anyone contributing to .NET Core) aspires to enable it to work for many interesting scenarios. Some of those are already part-way to completion. Nice work, team! 
+.NET Core is a general purpose, modular, cross-platform and open source implementation of .NET. It includes a set of [framework libraries](https://github.com/dotnet/corefx) and a [runtime](https://github.com/dotnet/coreclr). Both together support a set of OS platforms and chip architectures, enabling a broad set of .NET apps to be built and run. Much of this support is in progress.
+
+The team (anyone contributing to .NET Core) is composed of many individuals and groups. They (separately) have been adding support for the scenarios they are most interested in. That ranges from Windows to FreeBSD. Some of the projects are half-way to completion. Nice work, team! Some contributors, likely companies, will create official distributions (AKA "distros") that they support and potentially sell as part of some offering. A distro is an official build of .NET Core (e.g. zip, apt-get, brew, msi), much like Linux distros.
+
+Microsoft makes multiple "official" .NET Core distributions (libraries and/or runtime) available to its customers. It is expected (and desired) that other groups (corporations or otherwise) will also become significantly invested in .NET Core, making other distros available.
 
 This roadmap is intended to communicate the team's priorities for evolving and extending the scope of .NET Core.
-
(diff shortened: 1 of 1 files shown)


roadmap.md
```diff
@@ -0,0 +1,87 @@
+.NET Core Roadmap
+=================
+
+.NET Core is a general purpose, modular, cross-platform and open source implementation of .NET. The .NET Core team (anyone contributing to .NET Core) aspires to enable it to work for many interesting scenarios. Some of those are already part-way to completion. Nice work, team! 
+
+This roadmap is intended to communicate the team's priorities for evolving and extending the scope of .NET Core.
+
+Microsoft is very invested in .NET Core and makes multiple "official Microsoft" .NET Core distributions available to its customers. Sweet! It is expected (and desired) that other groups (corporations or otherwise) will also become significantly invested in .NET Core, possibly making other distros available. As a result, this roadmap is written from a neutral perspective, not a Microsoft one (unless specifically called out). This document is therefore naturally open to PRs.
```
akoeplinger: I think there needs to be a section somewhere that defines what exactly a "distribution" is in this case.

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+
+- .NET Core code is high quality (important consideration for contributions).
+- .NET Core can be ported to a broad set of OS platforms and chip architectures.
+- .NET Core has compelling performance and is highly reliable.
+- Developers can acquire a .NET Core developer environment quickly and intuitively.
+- Developers can productively and intuitively build apps, using documentation, samples and NuGet components.
+
+Active ports:
+
+- OSes: Linux, OS X, FreeBSD, Rasberry Pi2 (Linux) and Windows.
```
akoeplinger: Rasberry Pi2 isn't an OS, it's ARM Linux (or Windows IoT?).

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+- Linux distros: Debian family, CentOS.
+- Windows versions: Win7+ (client and server).
+- Chips: X64, X86, ARM32.
```
akoeplinger: I was under the impression that ARM64 is further along at this point? Or was that just the community porting effort?

roadmap.md
```diff
@@ -0,0 +1,87 @@
...
+|---------|------------|-------|-------|
+|Beta8    |    Oct 2015|ASP.NET|Beta   |
+|RC1      |    Nov 2015|ASP.NET|RC     |
+|v5 RTM   |     Q1 2016|ASP.NET|RTM    |
+
+Notes:
+
+- The RC1 release candidate will be a supported and production ready cross-platform release. 
+- Depending on feedback from RC1, we will ship additional release candidates as necessary.
+- The RTM release is **CY** Q1 2016.
```
akoeplinger: Do we really need CY here? I don't think anyone would think of the fiscal year here. If you want to keep it, please write it out as the abbreviation isn't commonly known to non-native speakers.

roadmap.md
terrajobst: I'm sure we'd do that if that definition would already exist :-). We're still trying to figure out exactly what this would look like (NuGet metapackage, ZIP file, some mapping, etc).

For now, the closes thing we have is UWP and ASP.NET 4 which essentially provide a list of packages that we consider "the recommended" set of packages. Essentially, the notion of a distribution is implicit there; there is no explicit concept that models that (yet).

roadmap.md
terrajobst: That's fair. We should probably rename that section to "Platforms: ". Strictly speaking, Windows isn't a singular platform either, there is Win32 and AppContainer, for example.

roadmap.md
akoeplinger: Makes sense :smile:

PR #35, opened 2015-12-07
Add warning about needing to delete ilcRepro.zip

MattWhilden:
We don't currently check to see if the file is already there and if it is, we crash ungracefully after spending multiple minutes building.


karelz:
LGTM

tmeschter:
:+1:

MattWhilden:
Cool. I think we need to get one of our counterparts to jam merge. @richlander @richlander @terrajobst

PR #64, opened 2016-03-23
Add Target Packs on NuGet spec

cartermp:
This doc is the working spec for how we can get targeting packs up on NuGet and use them.

This aims to address the corefx issue here: https://github.com/dotnet/corefx/issues/3934 and the CLI issue here: https://github.com/dotnet/cli/issues/191

Timeline for this is RTM, and we'd like to do this as a component of the CLI.  Personally, I also want to see this as the mechanism by which targeting packs are handled in VS and vscode.

/cc @richlander @terrajobst @ericstj @weshaggard @davidfowl @piotrpMSFT @onovotny

### DIFF
--- a/Documentation/design-docs/targeting-packs-on-nuget.md
+++ b/Documentation/design-docs/targeting-packs-on-nuget.md
@@ -1,0 +1,314 @@
+# Targeting Packs as NuGet Packages for Fun and Profit
+
+We want to create NuGet packages for each of the targeting packs.  Here's a short list of why we'd like to do this:
+
+- Making the lives of library developers easier
+- Dealing with a build server for multitargeting code of any kind
+- Allowing for easily auto-referencing assemblies for traditional PCL targets
+- Providing a way to allow for supporting new targets with PCL profiles and not break everyone's code
+
+A long-term goal is to make this the mechanism Visual Studio uses, as well.
+
+## Let's talk solutions
+
+1. Author NuGet Packages adhering to a particular naming scheme:
+
+   - `Microsoft.TargetingPack.NETFramework.v3.5`
+   - `Microsoft.TargetingPack.NETFramework.v4.6.1`
+   - `Microsoft.TargetingPack.NETPortable.Profile259`
+   - etc.
+   
+2. Implement a convention where the CLI treats the NuGet package as the replacement directory for framework assemblies.
+
+   - Instead of looking in e.g., `ref/net46`, it looks in the package directory instead.
+   - Will prefer targeting packs already installed on the machine if they exist (so as not to screw with VS).
+  
+3. Have the targeting pack define metadata that informs the build system.
+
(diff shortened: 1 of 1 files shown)


Documentation/design-docs/targeting-packs-on-nuget.md
```diff
@@ -0,0 +1,314 @@
...
+                     "System.Xml.Linq"
+                  ]
+               }
+            }
+       },
+       "net40": {
+            "dependencies":{
+               "Microsoft.TargetingPack.NETFramework.v4.0":{
+                  "version": "1.0.0",
+                  "references": [
```
onovotny: Why do we need a `references` section for the desktop frameworks? Is there harm in passing all the libraries in? One current issue is that a few of the files in the targeting pack are not .NET assemblies, but I would think putting those in a `ref` directory would solve that?

Main point is to not make the user do all the work and to treat the targeting pack just like the `netstandard.library` where it pulls in the world.

onovotny:
Why do we need a `references` section for the desktop frameworks? Is there harm in passing all the libraries in? One current issue is that a few of the files in the targeting pack are not .NET assemblies, but I would think putting those in a `ref` directory would solve that?

Main point is to not make the user do all the work and to treat the targeting pack just like the `netstandard.library` where it pulls in the world.

cartermp:
An issue there would be compiler performance if we pulled in all reference assemblies for .NET Framework targets at build time.

What I'm intending here is to mimic existing VS behavior today.  Adding an entry to a `references` section would be analogous to *Right-Click -> Add Reference* in VS.

onovotny:
The thing is, today it already refs all of them anyway, doesn't it? It does for the PCLs and expands/adds the contract assms.

cartermp:
For PCLs it does, but not for full .NET Framework.  E.g., if you were to use `SqlClient` you'd have to add a reference to `System.Data` - I believe we'd like to preserve that same kind of behavior here.

For PCL targets here everything would be auto-referenced.

onovotny:
Right, i get that. I'm asking if there's an opportunity to make it better. I was pointing out that for perf, the num of assms is already large. And you have to keep in mind the implicit expand of contract assms that also happens for desktop.

ericstj:
> The thing is, today it already refs all of them anyway, doesn't it? It does for the PCLs and expands/adds the contract assms.

PCL is significantly smaller than all of desktop.  @jaredpar would you want us passing every desktop assembly to roslyn?

> implicit expand of contract assms that also happens for desktop.

Those are just type-forwards.  They don't do much in the compiler if they targets aren't provided.

One thing we haven't done but could do to mitigate this problem would be add frameworkAssembly entries to our packages which have inbox facades.  That'd be something special you'd get from CoreFx packages, random PCL's wouldn't benefit.

cartermp:
Closing this for now, as product changes have effectively made this obsolete.  This can be revisited later.

PR #116, opened 2016-06-02
Create prereqs.md

blackdwarf:
Creating an overall .NET Core pre-req native deps file to be able to point people to.

/cc @richlander @gkhanna79 @ellismg

### DIFF
--- a/Documentation/prereqs.md
+++ b/Documentation/prereqs.md
@@ -3,6 +3,8 @@
 
 This document outlines the dependencies needed to run .NET Core tools and applications. 
 
+For Operating System support, please refer to [the roadmap document](https://github.com/dotnet/core/blob/master/roadmap.md#technology-roadmaps). 
+
 ## Windows dependencies
 On Windows, the only dependency is the VC++ Redistributable. Depending on the version of Windows you are running on, the versions are changing.
 
@@ -13,6 +15,7 @@ On Windows, the only dependency is the VC++ Redistributable. Depending on the ve
     * [Visual C++ Redistributable for Visual Studio 2015](https://www.microsoft.com/en-us/download/details.aspx?id=48145)
 * Pre-Windows 10
     * Please make sure that your Windows installation is up-to-date and includes hotfix [KB2533623] (https://support.microsoft.com/en-us/kb/2533623) installed through Windows Update.
+    * [Universal CRT update](https://www.microsoft.com/en-us/download/details.aspx?id=48234) (you can get more info on what Universal CRT is in [this blog post](https://blogs.msdn.microsoft.com/vcblog/2015/03/03/introducing-the-universal-crt/))
     * [Visual C++ Redistributable for Visual Studio 2012 Update 4](https://www.microsoft.com/en-us/download/confirmation.aspx?id=30679)
 
 ## Ubuntu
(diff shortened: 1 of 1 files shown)


Documentation/prereqs.md
```diff
@@ -0,0 +1,51 @@
...
+On Windows, the only dependency is the VC++ Redistributable. Depending on the version of Windows you are running on, the versions are changing.
+
+> **Note:** these dependencies are chained in the installer, you only need to install them manually if you are using
+> the installer script (`dotnet-install.sh` or `dotnet-install.ps1`). 
+
+* Windows 10
+    * [Visual C++ Redistributable for Visual Studio 2015](https://www.microsoft.com/en-us/download/details.aspx?id=48145)
+* Pre-Windows 10
+    * Please make sure that your Windows installation is up-to-date and includes hotfix [KB2533623] (https://support.microsoft.com/en-us/kb/2533623) installed through Windows Update.
+    * [Visual C++ Redistributable for Visual Studio 2012 Update 4](https://www.microsoft.com/en-us/download/confirmation.aspx?id=30679)
```
gkhanna79: Why is this required?

CC @schellap

gkhanna79:
@weshaggard PTAL as well.

Documentation/prereqs.md
```diff
@@ -0,0 +1,51 @@
+.NET Core native prerequisites
+==============================
+
+This document outlines the dependencies needed to run .NET Core tools and applications. 
+
+## Windows dependencies
+On Windows, the only dependency is the VC++ Redistributable. Depending on the version of Windows you are running on, the versions are changing.
```
gkhanna79: CRT Dependency: 

On Pre-Win10 OS, the only dependency is uCRT Windows update that can be installed from https://www.microsoft.com/en-us/download/details.aspx?id=48234. You can point folks to https://blogs.msdn.microsoft.com/vcblog/2015/03/03/introducing-the-universal-crt/ for details. 

On Win10+, this is available out of the box.

OS Dependency:

You should call out that for curated OS SKUs (e.g. the ones that are not full server or client - like Nano), they need to have Reverse Forwarders in their OS image.

Documentation/prereqs.md
blackdwarf: I don't know if it is. Could be a remnant of something else. Though I'm pretty sure @leecow ran into something similar recently on Win7 vanilla machine.

Documentation/prereqs.md
```diff
@@ -0,0 +1,51 @@
+.NET Core native prerequisites
+==============================
+
+This document outlines the dependencies needed to run .NET Core tools and applications. 
+
+## Windows dependencies
```
gkhanna79: Where are we calling out the list of supported OS platforms and architectures since that is part of the pre-req as well?

blackdwarf: https://github.com/dotnet/core/blob/master/roadmap.md here, I will link, that's a good point.

Documentation/prereqs.md
blackdwarf: WTF are "reverse forwarders"??

Documentation/prereqs.md
```diff
@@ -0,0 +1,51 @@
...
+* gettext 
+* libcurl-devel 
+* openssl-devel 
+* zlib 
+* libicu-devel
+
+## OS X 
+OS X requires the following libraries and versions installed:
+
+* libssl >= 1.1
```
bartonjs: >= 1.0.1.

The current version from homebrew is 1.0.2, and 1.1 isn't quite released yet, and may not even be compatible.

Documentation/prereqs.md
leecow: Without the 2012 crt, dotnet build will toss an error that it cannot find Microsoft.DiaSymReader.Native.amd64.dll.

Documentation/prereqs.md
gkhanna79: Also, see:

https://social.technet.microsoft.com/Forums/en-US/e08daebc-c298-4932-a9d7-ff08fabf9536/nano-server-packages?forum=NanoServer

Some formal definition at https://blogs.technet.microsoft.com/windowsserver/2015/11/16/moving-to-nano-server-the-new-deployment-option-in-windows-server-2016/

PR #124, opened 2016-06-13
Adding new distros and updating release date

Petermarcu:


richlander:
LGTM

mairaw:
The right brand name would be OS X (until the new macOS comes :smile:). Also, are these supported in RC2 already @Petermarcu?
Because now we have a topic on this subject on the docs site too, so I'd like to keep up-to-date: https://dotnet.github.io/docs/getting-started/installing/system-reqs.html

PR #138, opened 2016-06-22
Update known-issues.md

ehajri:
``` console
ekh@imac ~ ls -la /usr/local/share/dotnet/dotnet
ls: /usr/local/share/dotnet/dotnet: No such file or directory
ekh@imac ~ ls -la /usr/local/share/dotnet/bin/dotnet
-rwxr-xr-x 1 root wheel 43292 Mar 1 09:35 /usr/local/share/dotnet/bin/dotnet
```


ehajri:
nvm, that was for an earlier version.

PR #142, opened 2016-06-24
1.0 Release notes and Known Issues

leecow:
Still working on the API Diff so don't look at that yet ;-) Contributors list is a placeholder for the moment.


release-notes/1.0/Release-Notes-1.0.0.md
```diff
@@ -1,2 +1,103 @@
...
+Details on work the WCF team has done since RC2 can be found in the [WCF Release Notes](https://github.com/dotnet/wcf/releases/tag/v1.0.0).
+
+## Closed issues
+
+The lists of issues closed for this and past releases can be found here:
+
+* [CLR issues](https://github.com/dotnet/coreclr/issues?q=is%3Aissue+no%3Amilestone+is%3Aclosed)
+* [BCL issues](https://github.com/dotnet/corefx/issues?q=is%3Aissue+no%3Amilestone+is%3Aclosed)
+
+## Commits for RC2
```
bartonjs: Presumably you mean "RTM" or "v1", or something other than "RC2".

release-notes/1.0/known-issues-1.0.0.md
```diff
@@ -0,0 +1,232 @@
...
+* On OSX, Gss.framework supports SPNEGO mechanism with Kerberos and NTLM as the security protocols.
+* On Linux, mit krb5 library supports SPENGO mechanism and Kerberos as the out of the box security protocol. The implenentation can be made to support NTLM too by installing [GSS-NTLMSSP plugin](https://fedorahosted.org/gss-ntlmssp/) or another plugin with similar functionality. The fallback to NTLM is thus dependent on runtime availability of such a plugin.
+
+On RHEL and CENTOS, the plugin GSS-NTLMSSP is available from the package managers. The same is also available on Ubuntu 16. It can be installed on the other flavors of linux.
+
+There's no compile time dependency on this plugin.
+
+## Windows 7
+
+**Issue:**
```
gkhanna79: Can we point this to the pre-req document @blackdwarf wrote so that folks can install the required dependencies (which is what these "issues" are about)?

release-notes/1.0/known-issues-1.0.0.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Bash on Ubuntu on Windows
+
+[Bash on Windows (WSL)](https://msdn.microsoft.com/en-us/commandline/wsl/about) is not yet supported by .NET Core. Attempting to run applications in the environment can experience interittent crashes.
+
+<https://github.com/Microsoft/BashOnWindows/issues/520>
+
+## NegotiateStream.Write interop issues on Linux/OSX
+
+There are some combinations of Kerberos/NTLM, SignOnly/EncryptAndSign etc. that cause the ```gss_wrap``` call on Unix client to fail against a Windows server using ```NegotiateStream```. This issue is for tracking the foll. failing combinations of credentials and protection level passed in ```AuthenticateAsClientAsync```
```
bartonjs: Style: Do we want to use "foll.", or expand it to "following"?

release-notes/1.0/known-issues-1.0.0.md
```diff
@@ -0,0 +1,232 @@
...
+- NTLM creds with EncryptAndSign: Server complains about message format
+- NTLM creds with Sign: gss_wrap fails on client side
+
+The fix has been pushed to MIT Kerberos source <https://github.com/krb5/krb5/pull/436> which will make the fix available for all unix platforms, starting krb5-1.15. Red Hat reports the fix will be backported to RHEL 7..
+
+<https://github.com/dotnet/corefx/issues/6767>
+
+## Crypto
+
+Adding certs to an X509Store fails when the current user's home directory is an NTFS volume on Linux/OSX
```
bartonjs: This line may or may not want to be promoted to something more titular, since someone could accidentally read that we always don't support NTFS, instead of that we don't support it on Linux and OS X.  (We might also want to normalize the document style for "OS X", "OSX", etc.

release-notes/1.0/known-issues-1.0.0.md
```diff
@@ -0,0 +1,232 @@
...
+
+If global.json file has an UTF-16 BOM, dotnet commands will fail due to not supporting UTF-16 BOM with the error:
+
+```
+A JSON parsing exception occurred: * Line 1, Column 2 Syntax error: Malformed token
+```
+
+If you save global.json with an UTF-16 BOM (Byte Order Mark), dotnet commands will error out with the following error:
+
+A JSON parsing exception occurred: * Line 1, Column 2 Syntax error: Malformed token
```
bartonjs: These preceding two lines are almost the same as the two lines before them.  Probably only one or the other set should be used.

release-notes/1.0/known-issues-1.0.0.md
```diff
@@ -0,0 +1,232 @@
...
+
+If you’re using Nano Server Technical Preview 5 with .NET Core CLI, due to a bug in the product, you will need to copy binaries from `c:\windows\system32\forwarders`. The destination depends on the type of deployment that you are choosing for your application.
+
+For portable applications, the forwarders need to be copied to the shared runtime location. The shared runtime can be found wherever you installed .NET Core 1.0.0 (by default, it is `C:\Program Files\dotnet`) under the following path: `shared\Microsoft.NETCore.App\<version>\`.
+
+For self-contained applications, the forwarders need to be copied over into the application folder, that is, whereever you put the published output.
+
+This process will ensure that that the dotnet host finds the appropriate APIs it needs. If your Nano Server Technical Preview 5 build is updated or serviced, please make sure to repeat this process, in case any of the DLLs have been updated as well.
+
+## dotnet does not support UTF-16 BOM
```
bartonjs: We might want to make the title more clear that we mean the utility, not the framework. Replace "dotnet" with "global.json", perhaps?  (Unless it also applies to project.json?)

release-notes/1.0/Release-Notes-1.0.0.md
```diff
@@ -1,2 +1,103 @@
...
+* System.Security.Cryptography
+    * ECDsa supports creating keys for arbitrary named and explicit curves (Windows 10, OS X, Linux)
+	* ECDsa supports importing and exporting keys via the platform-independent ECParameters type
+    * EnvelopedCms is now supported via the System.Security.Cryptography.Pkcs package
+* System.Security.SecureString
+* System.Text.RegularExpressions - precompiled regex
+
+### CLI
+
+* Offline support: now you don't have to be connected to the internet to write applications that target only the .NET Core runtime and libraries
```
akoeplinger: I'd probably clarify that the "have to be connected to the internet" part only applied to restoring packages, it's not like there's some sort of constant connection from the CLI to a server. This just caches a set of packages locally on first run.

release-notes/1.0/Release-Notes-1.0.0.md
joshfree: Also typos like "implenentation" below. The NegotiateStream section could probably be improved further by giving exact package names. /cc @terlochan

release-notes/1.0/Release-Notes-1.0.0.md
blackdwarf: This is a good point @akoeplinger.

leecow: @akoeplinger , @blackdwarf 

How about - 

* Offline support: now you don't have to be connected to the internet to write applications that target only the .NET Core runtime and libraries. This means the core libraries are cached locally after running various dotnet commands the first time. Restoring packages which are not part of the Shared Framework do generally require and internet connection to gather the packages from NuGet.

release-notes/1.0/known-issues-1.0.0.md
```diff
@@ -0,0 +1,232 @@
...
+## NegotiateStream
+
+```NegotiateStream``` relies on the gssapi implementation available on the platform.
+* On Linux flavors, the default gssapi implementation provided is [MIT's krb5 library](http://web.mit.edu/kerberos/) which is available on all the linux platforms
+* On OSX, the default implementation is heimdal-based GSS.framework
+
+On both Linux and OSX, NegotiateStream uses SPNEGO and relies on the underlying implementation for supporting Kerberos or NTLM as the underlying security protocol.
+
+* On OSX, Gss.framework supports SPNEGO mechanism with Kerberos and NTLM as the security protocols.
+* On Linux, mit krb5 library supports SPENGO mechanism and Kerberos as the out of the box security protocol. The implenentation can be made to support NTLM too by installing [GSS-NTLMSSP plugin](https://fedorahosted.org/gss-ntlmssp/) or another plugin with similar functionality. The fallback to NTLM is thus dependent on runtime availability of such a plugin.
```
joshfree: > implenentation 

typo still exists

release-notes/1.0/Release-Notes-1.0.0.md
```diff
@@ -81,7 +81,7 @@ New .NET Core APIs
 
 ### CLI
 
-* Offline support: now you don't have to be connected to the internet to write applications that target only the .NET Core runtime and libraries
+* Offline support: now you don't have to be connected to the internet to write applications that target only the .NET Core runtime and libraries. This means the core libraries are cached locally after running various dotnet commands the first time. Restoring packages which are not part of the Shared Framework do generally require and internet connection to gather the packages from NuGet.
```
akoeplinger: typo: ~~and~~ an internet connection

joshfree:
could you squash this into 1 commit when it's ready?

release-notes/1.0/known-issues-1.0.0.md
```diff
@@ -0,0 +1,232 @@
...
+```
+sudo dnf downgrade nspr nss-util nss-softokn-freebl nss-softokn nss-sysinit nss nss-tools


+```
+
+<https://github.com/dotnet/cli/issues/3676>
+
+
+
```
stephentoub: We appear to be missing both of these:
https://github.com/dotnet/corefx/issues/9235
https://github.com/dotnet/corefx/issues/9240
Not sure if that's by choice or an accident.

leecow: @stephentoub - got 'em

PR #143, opened 2016-06-25
Update known-issues.md

blackdwarf:
Add the https://github.com/NuGet/Home/issues/3038 as known issue.

/cc @leecow @naamunds


naamunds:
@blackdwarf I think the known issues should also mention that even with the workaround for `dotnet restore`, `dotnet run` will still fail for projects in the root of the file system, and I'm not aware of a workaround for that. See https://github.com/dotnet/cli/issues/3697

PR #146, opened 2016-06-26
Change OSX -> OS X

leecow:


stephentoub:
LGTM

mairaw:
We're actually changing all the references of OS X to macOs in the docs and just leaving the ones that specifically talk about an OS X version. Not sure if it's something you wanna do here too.

akoeplinger:
@leecow there seem to be some inaccuracies in the contributor list, e.g. [@VSadov](https://github.com/dotnet/corefx/commits/release/1.0.0?author=vsadov@microsoft.com) shows up with 1 commit to corefx but he did a lot more according to GitHub's commit list. Additionally I don't show up at all (not trying to put a plug in for myself, but it makes me wonder who else was missed?).

It might also be a good idea make the commit links for each author specific to the `release/1.0.0` branch instead of `master`.

leecow:
Hey @akoeplinger - I used the following to generate the list. Being a git neophyte, any suggestions on how to increase the accuracy would be great!

`git shortlog -sne --no-merges --since release/1.0.0@{2016-03-21}`

akoeplinger:
@leecow `--since release/1.0.0@{2016-03-21}` sounds wrong to me, this would only count commits to release/1.0.0 after March 21. Don't you want _all_ commits from the beginning of the repo for the 1.0 release?

leecow:
@akoeplinger - arg, of course! I've gotten a bit bleary headed today ;-) Re-running the scripts.

PR #160, opened 2016-06-27
Update README.md

aochagavia:


aochagavia:
I hope the updated information is accurate. Otherwise let me know so I can correct it.

PR #161, opened 2016-06-27
Correct indentation

geek:


Petermarcu:
Looks like this conflicts with the updates to use the 1.0.0 version which also addressed indentation to match the templates.

PR #162, opened 2016-06-27
Don't add .DS_Store to source

geek:


AustinWise:
There are some .DS_Store files checked in:
````
.DS_Store
release-notes/.DS_Store
````
Perhaps they should be removed?

leecow:
Yep, accidentally included and will watch more closely for that.

PR #163, opened 2016-06-27
Fix contributor list links for the dotnet/cli repository.

AustinWise:
The cli repository currently does not have a master branch. This fixes the links to point at the rel/1.0.0 branch instead. I did some spot checking with `git rev-list --no-mergs --count origin/rel/1.0.0` and it matches up with the commit counts on this page.

As an aside, I think it would make a little more sense to link to tag rather than the master branch for this page, as over time these links will start to show commits that did not ship in 1.0, but that's a separate issue.


leecow:
lgtm

PR #168, opened 2016-06-27
1.0.0-api-list

leecow:


leecow:
cc @ericstj , @terrajobst for the OK

ericstj:
Are we sure GH can actually render this?  All the views I click on say it is too large.

richlander:
That's an interesting problem.

One option is to break into two files:

- Standard Library
- Other .NET Core APIs

Would that help?

leecow:
Working on an update that will use separate files per namespace and work better than the single giant file.

PR #216, opened 2016-07-09
Remove all .DS_Store files

Senorsen:


akoeplinger:
:+1: might make sense to add .DS_Store to .gitignore

Senorsen:
`. DS_Store` is already in .gitignore, these two files might be introduced into the repo before added to .gitignore, I guess 😆

EtherTyper:
Also may have been introduced by an internal TFS Server.

PR #225, opened 2016-07-15
Update README.md

blackdwarf:
Add the note on creating issues.

/cc @joshfree @richlander @Petermarcu

### DIFF
--- a/ISSUE_TEMPLATE
+++ b/ISSUE_TEMPLATE
@@ -1,0 +1,7 @@
+Thanks for your interest in .NET Core. 
+
+Please note that this repo (dotnet/core) is not for filing product issues. Here are some repos where you can file an issue:
+ * dotnet/cli - for CLI tools and questions
+ * dotnet/corefx - for API issues and questions
+ * dotnet/coreclr - for runtime issues
+* nuget/home - for NuGet questions and issues
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The core repository is the starting point to engage in and learn about the
 ## Get Started
 
 If you're new to .NET Core and have 10 minutes to try it, start here: 
-- [What is .NET Core?](http://dotnet.github.io/about/)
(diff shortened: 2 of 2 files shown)


mairaw:
Maybe you guys would also benefit of having an issue template as @BillWagner did for core-docs:
https://github.com/dotnet/core-docs/blob/master/.github/ISSUE_TEMPLATE.md

Then this information would surface when the person is actually creating the issue.

blackdwarf:
@mairaw good point.

README.md
```diff
@@ -1,12 +1,21 @@
...
+.NET Core stack. 
+
+> **Note:** please note that this repository is not for filing product issues.
+> If you run into an issue using .NET Core, there are multiple repos where you can 
+> file an issue:
+> * dotnet/cli - for CLI tools and questions
+> * dotnet/corefx - for API issues and questions
+> * dotnet/coreclr - for runtime issues
+> * nuget/home - for NuGet questions and issues
+> We will be removing the ability to file issues on this repo in the near future.
```
svick: * Where should I file an issue if it does not fall clearly within one of these categories or if I'm unsure where it belongs?
* These should be links.
* Also mention aspnet/Home?

ISSUE_TEMPLATE
```diff
@@ -0,0 +1,7 @@
+Thanks for your interest in .NET Core. 
+
+Please note that this repo (dotnet/core) is not for filing product issues. Here are some repos where you can file an issue:
+ * dotnet/cli - for CLI tools and questions
+ * dotnet/corefx - for API issues and questions
+ * dotnet/coreclr - for runtime issues
+* nuget/home - for NuGet questions and issues
```
svick: Nit: since users will see the markdown here, I think the list should consistently have no indentation, and not inconsistently have one space indentation for the first three items and no indentation for the last item.

PR #238, opened 2016-07-31
Update known-issues.md to address issue on latest brew

noelitoa:
Update know-issues.md to publish a work around on brew refusing to link openssl from its latest version.


noelitoa:
Submitted an issue as well - https://github.com/dotnet/cli/issues/3968

venil7:
I'm afraid solution suggested here http://stackoverflow.com/questions/38670295/brew-refusing-to-link-openssl doesn't solve an issue

I encounter the same issue while installed `openssl101`

bartonjs:
Yeah, Homebrew fixed the part where the older version could bypass the block.

We'll be putting out new instructions soon, which don't involve `brew link`.

blackdwarf:
LGTM merging...

PR #239, opened 2016-08-01
Add DYLD_LIBRARY_PATH workaround for OS X install

joshka:
Remove deprecated suggestions to use `brew link --force openssl`

I recognize that a more permanent fix for this is forthcoming. Hopefully this fix will assist those coming to this until then.


blackdwarf:
Closing this as our workaround will be to manually link into /usr/lib. These will be in the updated instructions on the Getting Started website.

PR #256, opened 2016-08-26
Update manual-shared-update.md

Petermarcu:
These instructions are specific to runtime and JIT. Clarifying that in the title and summary. It needs a lot more instructions to cover all the other contents in the shared framework.


Petermarcu:
@leecow

PR #259, opened 2016-09-03
Added temp directory checking and clean up

leecow:
Add checks to determine if 1.0.1 is already installed and if tmp directory is possibly around after a previous install attempt. Add tmp directory clean up. Updated var names to be easier to read.


blackdwarf:
:shipit: I need to add a PS version of this as well

PR #277, opened 2016-09-26
Add Dockerfile

richlander:
- Added Dockerfile for app
- Made the app a bit shorter
- Improved app with string interpolation usage


blackdwarf:
:shipit: This was also a sample for corert back in the day so that's why it was...messy. :))

richlander:
Sure, sure. The focus should be on the Dockerfile. Good?

samples/dotnetbot/Program.cs
```diff
@@ -6,23 +6,23 @@ public static class Program
...
-                        if (args.Length < 1) 
-                        {
-                          message = "Welcome to .NET Core!";
-                        } else 
+                      string message = "Welcome to .NET Core!";
+                        
+                       if (args.Length > 0) 
                         {
+                          message = String.Empty;
                           foreach (string item in args)
```
blackdwarf: This here was done because of CoreRT. Since this is now a pure IL app, you could do something like `String.Join(args, ' ')` to get the final message to display. :)

blackdwarf:
If you're asking me if the Dockerfile is good, yes, it looks good.

PR #278, opened 2016-09-26
Update known-issues.md

blackdwarf:
Removing some stuff, adding TOC and adding the macOS Sierra workaround for $PATH.

/cc @richlander @leecow

### DIFF
--- a/cli/known-issues.md
+++ b/cli/known-issues.md
@@ -1,6 +1,43 @@
 Known issues & workarounds
 ==========================
 
+<!-- TOC -->
+
+- [Known issues](#known-issues)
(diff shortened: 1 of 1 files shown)


richlander:
I didn't realize that we had CLI-specific release notes. Some of these are CLI-specific while others are not. I think that most people will expect to find all of this information in the main release notes.

There are some minor wording issues that would be good to fix. Here is an example of something that could be worded better: "Users have reported this being the case." We should state what "the case" is. Something like the following would be better: "Users have reported that the dotnet installer does not update the path with the .NET Core install location."

Petermarcu:
@blackdwarf , Anything left here. Are you wanting to clear up the conflicts and update the PR or just close?

Petermarcu:
@blackdwarf , ready to commit or should we close this until you are?

PR #317, opened 2016-10-25
Update instructions for ANCM

shirhatti:
The server hosting bundle carries ANCM, LTS shared framework, and FTS shared framework. To install ANCM, you have to opt out of both the shared frameworks.


shirhatti:
@leecow

PR #326, opened 2016-11-01
Moving Rover (dotnet-bootstrapping tool) from private repository into dotnet/core

bryanar:
Tool itself is functional, but need to build the testing pieces.

Going to start a list here, we've got good discussions and feedback happening, so I thought I'd just keep the work centralized here so that everyone is in the loop as to what I am up to (I click the checkbox as I accomplish the work - pushes will be intermittent throughout the day; order is determined by mood):

- [ ] Explicitly copy just the binaries listed in the target repository directory (at the moment I am using the Kleene Star to select all the so files to copy) [ I'd like to add in a 'patching' feature that will just adopt the layout of the target directory and then move files from a source path; not this week though] 
- [x] Incorporate Run script for the docker containers (i.e. something that will just launch them and execute my test case)
- [x] Use [DotNet Bot](https://github.com/dotnet/core/blob/master/samples/dotnetbot/Program.cs) + [QOTD](https://github.com/dotnet/core/blob/master/samples/qotd/Program.cs) as test case
- [x] Incorporate intended Docker scripts
- [x] Provide examples in the README.md
- [x] Document the layout of the working directory
- [x] Do a 'satisfaction pass' over the documentation to make it... cleaner. [ used format of man pages ]
- [x] Place .gitattributes/.gitignore in root (if not there); crib (great word) from CoreFX
- [x] Remove dependency installer 
- [x] (Regarding colored text) Guard writing the escape codes with a check for sys.stdout.isatty()
- [x] (Minor bug fix) - On Unexpected Rover Exception I am still printing a success message (oops) [ added an os._exit(...) call to the UnexpectedRoverException() method so that after we catch the exception we just bail out. This should prevent any messages from popping up or risk of some future addition executing that maybe depends on this existing ]
- [x] (Minor bug fix) - If interrupted during the cloning of repositories, restarting causes us to skip over the clone step [ I removed the -clone switch, I think this was over-engineered so I just check if the directory exists, if it doesn't then I will clone ]
- [x] Add a -to that overrides the default directory name that we stage in. Currently we stage in a local directory with the format 'dotnet-<RID>'
- [x] Cache CLI download
- [x] sudo repro-script on creation (so long as uid == 0)


tools/cli-bootstrap/repo-dependencies-installer.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
```
ellismg: Can we just remove all this stuff?  I don't think it's super worthwhile.  I don't know of another project that tries to automagically install a tool-chain on your behalf.  I also worry that we'll just constantly be adding packages for different things.  We should just document the toolchain at a high level.

tools/cli-bootstrap/dockerfiles/debian8/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM debian:8.0
```
ellismg: It's by design that these docker files don't have toolchains installed right?

tools/cli-bootstrap/dockerfiles/debian8/Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+MAINTAINER Bryan P. Arant <bryanar@microsoft.com>
+RUN apt-get -qqy update
```
ellismg: Consider doing all this setup in one run command to reduce the number of layers that are constructed.  Also consider cleaning the apt-get cache as part of the same command to reduce layer size (the dockerfiles that we have for dotnet core should do this if you need an example).

tools/cli-bootstrap/dockerfiles/debian8/Dockerfile
bryanar: Yes, I will be 'borrowing' from the CLI dockers in a later PR.

tools/cli-bootstrap/.gitattributes
```diff
@@ -0,0 +1,3 @@
+*.py -crlf
```
ellismg: We could consider just doing this:

###############################################################################
# Set default behavior to automatically normalize line endings.
###############################################################################
* text=auto

in a top level .gitattributes (that's what most other dotnet projects do).

tools/cli-bootstrap/README.md
```diff
@@ -0,0 +1,36 @@
...
+However, things don't always 'just work.' When this is the case Rover hopes to reduce the amount of effort necessary to fix build or script breaks by placing 'repro' shell scripts when there is a failure so that a developer can go in and 'apply pressure' where appropriate. 
+
+## How
+
+There are default settings for each of the rover arguments. If all is going to plan, you should never need to specify any additional parameters. However
+many situations arise where it becomes a necessity or a nice-to-have. Namely, when things go wrong, you want to be able to get in there and fix
+up the build, make changes, etc. This is 'development mode' or DevMode for short. In DevMode, NO git commands are executed. This is to prevent 
+the script from stomping out any changes you have made in the working directory. 
+
+DevMode is triggered if the working directory is pre-existing. Consequently, when things fail in the script, we do not 'clean up' the working directory. We write a repro script
```
bryanar: Ah yes. I will revise the docs a bit overall. There is a bunch of stuff here that I do not like or is even a bit out-dated already.

ellismg:
It may make sense to move .gitignore and .gitattributes to the top level of the repository (and maybe just crib from the version we have in corefx)?

tools/cli-bootstrap/.gitattributes
bryanar: Cool! Lets do that (if we don't have it already, ill check)

tools/cli-bootstrap/cli.bootstrap.py
```diff
@@ -0,0 +1,562 @@
...
+    CoreCLRBinDirectory                 = ''
+    CoreFXBinDirectory                  = ''
+    CoreSetupBinDirectory               = ''
+    LibUVBinDirectory                   = ''
+
+    PatchTarget_Shared                  = ''
+    PatchTarget_SDK                     = ''
+    PatchTarget_Host                    = ''
+
+    DotNetCommitHash                    = 'rover-boot-strap'
```
ellismg: Gotcha.  Sorry.

tools/cli-bootstrap/dockerfiles/debian8/Dockerfile
mellinoe: Should these dockerfiles just be removed for now, then? It doesn't even look like most of them will actually build at all since they're all using the same Ubuntu commands.

tools/cli-bootstrap/cli.bootstrap.py
```diff
@@ -0,0 +1,562 @@
...
+from os import path
+from os import makedirs
+
+from string import find
+from urllib import urlretrieve
+
+
+# ROVER BASE #
+class RoverMods:
+    HEADER = '\033[95m'
```
mellinoe: Does this stuff work in docker containers? I'm pretty wary of us putting any colors into stuff now, since our attempts elsewhere rarely work in docker, and often break basic text output and formatting.

bryanar: Yep, works fine in my experience.

tools/cli-bootstrap/cli.bootstrap.py
```diff
@@ -0,0 +1,562 @@
...
+        if RoverSettings.Patch:
+            RoverPrint(RoverMods.Blue('is patching %s'%(RoverMods.Yellow(patchTarget_folder))))
+                                
+            # replace native dotnet in the base directory
+            # from core_setup
+            RoverShellCall('cp dotnet %s'%(path.join(patchTarget_folder)), cwd='%s/exe/'%(core_setup_cli_bin_directory))
+
+            # replace native files in 'shared' folder.
+            # from coreclr
+            RoverShellCall('cp *so %s'%(RoverSettings.PatchTarget_Shared),                               cwd=coreclr_bin_directory)
```
mellinoe: We don't need all of the *so files from coreclr's output, only the ones that should be in the shared framework directory.

tools/cli-bootstrap/dockerfiles/debian8/Dockerfile
bryanar: I don't see how having these here causes a problem other than personal preference.

tools/cli-bootstrap/cli.bootstrap.py
```diff
@@ -0,0 +1,562 @@
...
+    default_core_setup_cli_bin_directory= '%s/src/corehost/cli'%(core_setup_working_git_directory)
+    default_libuv_bin_directory         = '%s/.libs'%(libuv_working_git_directory)
+
+    RoverSettings.CoreCLRBinDirectory   = default_coreclr_bin_directory
+    RoverSettings.CoreFXBinDirectory    = default_corefx_native_bin_directory
+    RoverSettings.CoreSetupBinDirectory = default_core_setup_cli_bin_directory
+    RoverSettings.LibUVBinDirectory     = default_libuv_bin_directory
+    
+    platform_info                       = platform.uname()    
+    this_distro_name                    = str(platform.linux_distribution()[0]).lower()
```
mellinoe: What about non-Linux platforms?

ellismg: I think it's fine to be specific for Linux right now...

tools/cli-bootstrap/cli.bootstrap.py
```diff
@@ -0,0 +1,562 @@
...
+    ##
+    ## BEGIN DECLARATIONS
+    ##
+
+    coreclr_working_git_directory       = path.join(RoverSettings._srcDirectory, 'coreclr')
+    corefx_working_git_directory        = path.join(RoverSettings._srcDirectory, 'corefx')
+    core_setup_working_git_directory    = path.join(RoverSettings._srcDirectory, 'core-setup')
+    libuv_working_git_directory         = path.join(RoverSettings._srcDirectory, 'libuv')
+
+    default_coreclr_bin_directory       = '%s/bin/Product/Linux.x64.Release/'%(coreclr_working_git_directory)
```
bryanar: Yeah, I kind of want to see the build system overall stabilize there. Isn't there an API or something coming along in this regard? Otherwise this tool will have to chase the repositories behaviors around everywhere (as it seems we do anyways).

tools/cli-bootstrap/cli.bootstrap.py
bryanar: Not on mind right now.

tools/cli-bootstrap/dockerfiles/debian8/Dockerfile
mellinoe: I was just suggesting that we add them with the follow-up PR instead, since a lot of these don't actually work in their current state (they all use `apt-get` commands). It's not a big deal either way.

tools/cli-bootstrap/cli.bootstrap.py
mellinoe: I just tested and this works on an alpine container for me. core-setup and cli have their own variant of console color, and it doesn't work in most docker containers. There's also some shell scrpts around that don't work unless you install stuff that they use for console coloring.

But this works, so I'm cool with it.

tools/cli-bootstrap/dockerfiles/debian8/Dockerfile
bryanar: Actually, since there is more feedback than I anticipated, I will probably update this PR tomorrow with the intended Docker scripts; so, stay tuned.

tools/cli-bootstrap/cli.bootstrap.py
bryanar: @ellismg can you give some perspective on this too? I really do not have much of a clue after further reflection about the best way to handle the case @mellinoe is bringing up. I could just raise up a config for this, but I do not want that to become 'the way' we do things.

mellinoe: Yes, it makes sense. It's a difficult thing to solve unless the coreclr build tells you in some way where it is putting the binaries. The other problem with coding the logic here in this script is that, if the logic changes in coreclr, then the script here will work for some coreclr commits and not others.

In my script, I essentially pattern-matched the output location of CoreCLR from the build log. That is fragile in its own way, but it's also sort of similar to other pattern matching stuff in this script (to determine the commit hash). And it lets you be resilient to changes between commits, as long as the build output format doesn't change drastically. I don't know if that is a great, robust solution, either, but it's something to consider.

bryanar: I might steal your method Eric... I want to investigate a few things first before I do that.

One OTHER idea which I want to put on the table for us to think about is 'rewriting history'. Theoretically if a repository doesn't have a consistent interface (throughout time), I'm thinking we could go back in time to add the additional interfaces we want. For example, if build.sh didn't have a --bindir tag, we could go back and ADD it (no removing, no changing what is there already, just ADDING - to escape conflict, etc.). And it'd essentially just be a property setting we'd pipe through to MSBuild or whatever. Any thoughts on this approach?

bryanar:
test this please

bryanar:
test ci please

bryanar:
test this please

bryanar:
test ci please

bryanar:
test this please

bryanar:
test ci please

bryanar:
test ci please

bryanar:
test ci please

bryanar:
test ci please

bryanar:
test this please

bryanar:
test ci please

bryanar:
test ci please

bryanar:
test ci please

bryanar:
test ci please

bryanar:
test pr please

bryanar:
test ci please

bryanar:
CI test is passing, going to hold off on one more thing - I only want the CI to run when modifying dotnet.bootstrap.py

tmds:
I tried the dotnet.bootstrap.py tool on Fedora 25.

What I found out:
* The tool does not take care of dependencies, which is okay because each repo does a pretty good job of reporting it's own dependencies. It means "development mode" is not exceptional but it is a typical part of bootstrapping. Why isn't this the 'only' mode?
* The script files which are created to debug a failing build step miss the execute flag. They need a manual chmod +x.
* Each time I restarted the tool, it downloaded the cli again saying "Could not locate a payload at path ''". Perhaps the tool can store the downloaded file as dotnet.tar.gz file and automatically use it if no other payload was provided?
* I don't get what the -nopatch is for. The goal of the tool is to patch? Or perhaps I don't understand what the nopatch does.
* I had to make some changes to the repositories which were already fixed on master due to the compiler being a more recent one. This is a normal part of bootstrapping.
* In the end, I had a dotnet. It managed to 'dotnet new' and 'dotnet restore'. But the 'dotnet build' hang. It was using msbuild. This was a bit of a surprise since there hasn't been an SDK release with msbuild. How does the tool chose a tar.gz file to download? And how does it chose the tags/branches of the different repositories it is using?
* At that point I decided to manually pick the fedora LTS release and provide it using the -payload option. It took me a bit longer to find on the dot.net website that it would for a Windows/Mac user. When running it providing the payload option I got an uncaught exception: "SemanticVersion instance has no attribute 'find'" of type <type 'exception.AttributeError'>.
* Not sure what to delete, I deleted the fedora.25-x64-dotnet folder to get things in order again. I didn't expect the src folder to be in there, so this accidentally caused me to delete the changes I made to the repositories. Perhaps the src folder can move up one level?

All in all, I liked how the tool helps bootstrapping. Using the scripts, it has a nice workflow a la "bootstrap --continue".

bryanar:
@tmds Fantastic feedback! Thank you! 

- "The tool does not take care of dependencies" - True, this was snipped off (too many instances out there where dependencies can't just be 'guaranteed' which makes the feature useless except for cases that are 'well known', but in the well known instances we are less likely to need this. If really in a bind for dependencies, review the docker containers in base/lab/containers that has the apt-get, yum, zypper, dnf, etc. commands to pull in packages. @gkhanna79 (Relevant to you).
- "The script files which are created to debug a failing build step miss the execute flag" - Will fix. 
- "It downloaded the CLI again..." - Will fix.
- "What is no patch?" - This is an experimental feature that probably does not need to exist any longer (i.e. I'll remove it after a bit more review). 
 - "How does the tool choose?" - At the moment it does no choosing - we simply grab the latest debian dotnet bits, then we grab specific checkout commits and build, patch, run. In the future we will take the version number/commit hash from the binary and build only those pieces (in other words, whatever binary versions you provide (via -payload; or whatever we download) we will reproduce for the selected platform). 

- Given your story the DevMode makes me think this has caused a headache having this distinction. I too think DevMode should be 'default' now - or at least removed. I also like your idea of 'bootstrap --continue' - probably going to steal that ;)

Again, thank you for sharing your experience with using this tool!

tmds:
Second attempt of bootstrapping on Fedora 25.

# Install Fedora 25

Used iso: https://download.fedoraproject.org/pub/fedora/linux/releases/25/Workstation/x86_64/iso/Fedora-Workstation-Live-x86_64-25-1.3.iso

# Install dependencies

```
$ sudo dnf -y install python cmake clang llvm libicu-devel lldb-devel libunwind-devel lttng-ust-devel libuuid-devel zlib-devel libcurl-devel krb5-devel openssl-devel automake libtool
```

# Get Rover

```
$ git clone https://github.com/dotnet/core
$ cd core
$ git checkout bootstrap
$ git rev-parse HEAD
f895ac6b595ff03a22bd1278f803dbfe29c61e75
$ cd tools/dotnet-bootstrap
```

# Get payload SDK

Using Fedora 24 .NET Core 1.1 SDK:
```
$ curl -sSL -o dotnet.tar.gz https://go.microsoft.com/fwlink/?LinkID=835025
```

# Select your branches

```
$ mkdir -p fedora.25-x64-dotnet/src
$ cd fedora.25-x64-dotnet/src
$ git clone https://github.com/dotnet/corefx
$ cd corefx
$ git checkout release/1.1.0
$ git rev-parse HEAD
639eeadbf43c0b2be334ea0ee83ae9da2ca03578
$ cd ..
$ git clone https://github.com/dotnet/coreclr
$ cd coreclr
$ git checkout release/1.1.0
$ git rev-parse HEAD
673561935b3112e1128392b61b4e974f1b04545c
$ cd ../..
```

I let rover pick branches for libuv and core-setup:
```
core-setup at v1.1.0 (928f77c4bc3f49d892459992fb6e1d5542cb5e86)
libuv at v1.9.0 (229b3a4cc150aebd6561e6bd43076eafa7a03756)
```

# Bootstrap

```
$ ./dotnet.bootstrap.py -payload dotnet.tar.gz
...
** ROVER spawned a 'dotnet' in ./fedora.25-x64-dotnet/bin/(enjoy!)
```

# Does it work?

Doing the `dotnet new;dotnet restore;dotnet run` gets stuck compiling:

```
Project hwapp (.NETCoreApp,Version=v1.1) will be compiled because expected outputs are missing
Compiling hwapp for .NETCoreApp,Version=v1.1
```

ellismg:
This PR also adds a ubuntu directory with submodules.  Was that expected or was output from a local invocation of the tool accidentally included?

bryanar:
@ellismg Accident. I will fix this up and close it up now actually.

bryanar:
@tmds Oops, I do actually have the check for chmod +x on the repro scripts. If you run the script in sudo mode this is the only time I will set the X bit on the repro script.

bryanar:
test this please

bryanar:
test ci please

PR #332, opened 2016-11-06
Fix DotNetFoundation links.

agriffard:
http://www.dotnetfoundation.org/net-core
http://www.dotnetfoundation.org/net-compiler-platform-roslyn
http://www.dotnetfoundation.org/asp-net-core


Petermarcu:
Thanks!

PR #343, opened 2016-11-16
Preview 3 URLs mixed up for Windows 32-bit/64-bit

AndyMDoyle:
The 32-bit link downloads the x64 executable, and the 64-bit link downloads the x86 executable.

Fixed by swapping the URLs around.


slang25:
This had me confused

shaunluttin:
What will it take to merge this PR?

Petermarcu:
Thanks for pointing it out and making the fix.

PR #350, opened 2016-11-17
links to 32bit and 64bit installers where inverted

bitbonk:

PR #358, opened 2016-11-20
Fix 32/64 bit links.

shaunluttin:
The problem was that the 32 bit link pointed to the 64 bit download and vice-versa. This commit fixes that.


svick:
This is a duplicate of https://github.com/dotnet/core/pull/343 and https://github.com/dotnet/core/pull/350.

shaunluttin:
Lol. Why not bring in those fixes?

PR #385, opened 2016-12-08
Fix broken link to portable PDB reader

atifaziz:
The link below was broken:

> The portable format is [open source](https://github.com/dotnet/roslyn/tree/master/src/Debugging/Microsoft.DiaSymReader.PortablePdb)…

Looks like the new home is https://github.com/dotnet/symreader-portable so this PR fixes the link to point there instead.


Petermarcu:
Thanks

PR #394, opened 2016-12-14
add f# info in README

enricosada:


terrajobst:
Hey @enricosada! Thanks for adding that.

@cartermp and I just updated the roadmap to include more of the .NET components, which also includes F#. That's in PR #397. Does this work?

enricosada:
@terrajobst awesome! can i also mention https://github.com/dotnet/netcorecli-fsc repo?

Because it's where f# sdk issue are posted. There is also source code for integration, [wiki](https://github.com/dotnet/netcorecli-fsc/wiki) for info about .net core versions, [example project per version](https://github.com/dotnet/netcorecli-fsc/tree/master/examples), etc

And there are the TODO issues :D 

I can add it with this pr if you want

Petermarcu:
@terrajobst , thoughts on @enricosada 's question?

terrajobst:
The README doesn't include dotnet/sdk. We should probably add both.

@enricosada, do you mind doing this?

PR #399, opened 2016-12-16
Update roadmap.md

dpen2000:
Fix link to Roslyn

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -34,7 +34,7 @@ there are some big themes you will identify in the early work on GitHub:
   applications faster, especially in the inner loop. This is the cycle of
   changing the source code and then restarting the application and making that
   as fast as possible. You can follow part of this work in the
-  [dotnet/roslyn](https://github.com/dotnet/standard) repo.
+  [dotnet/sdk](https://github.com/dotnet/sdk) repo.
 
 * **.NET Core and Cloud**. Continue to improve how you run .NET Core
   applications in Azure. Better logging, tracing and diagnosing errors in your


Petermarcu:
Actually, I don't think thats where this work is happening. Its not about the compiler's perf, its more about the end to end pipeline. Its really not in a single repo but it is probably not really in the standard repo at all. @terrajobst, what should this link point to?

terrajobst:
I thought part of the performance work is done in the IDE experiences of C#/VB which are in dotnet/roslyn.

@Pilchie, is there a better location I should point to?

Petermarcu:
I think the work is spread across dotnet/roslyn-project-system, microsoft/msbuild, dotnet/sdk, dotnet/cli, and dotnet/roslyn. There are probably others :). Maybe we just shouldn't point to where people can follow the work? :P

roadmap.md
```diff
@@ -34,7 +34,7 @@ there are some big themes you will identify in the early work on GitHub:
   applications faster, especially in the inner loop. This is the cycle of
   changing the source code and then restarting the application and making that
   as fast as possible. You can follow part of this work in the
-  [dotnet/roslyn](https://github.com/dotnet/standard) repo.
+  [dotnet/roslyn](https://github.com/dotnet/roslyn) repo.
```
Petermarcu: I think the best place to point this will be the SDK repo. Thats where the common tasks and targets live that will be responsible for driving the various build components including the compiler, msbuild, cli, etc. This work is about making the whole pipeline of those tools lightning fast :) Do you want to update the PR?

dpen2000:
I've amended and pushed my commit, @Petermarcu with change to sdk

dpen2000:
Hi @Petermarcu I'm going to send .net foundation an email about a CLA issue so perhaps this may have to wait to be merged.

Petermarcu:
This one doesn't need a CLA. Thanks for the update.

PR #401, opened 2016-12-19
fix dotnet/roslyn url.

hikalkan:


muojp:
dupl: https://github.com/dotnet/core/pull/399

(I actually noticed the duplication when I tried to send yet another pull-request 😆 )

hikalkan:
For #399, "Update roadmap.md" title is not clear :stuck_out_tongue_closed_eyes:

Petermarcu:
@hikalkan, thanks for making the change. I'm going to close this as a duplicate of #399 but I'm not sure we really intended to point to Roslyn here. This work isn't about compiler throughput. Its about the end to end build pipeline of which Roslyn is only a small part.

hikalkan:
Thanks @Petermarcu for the explanation.

PR #407, opened 2016-12-27
Add .csproj based samples

jpierson:
The csproj based samples were added which were derived by running the `dotnet migrate` command. I've also updated the Markdown documentation to mention specific usage of the .NET Tooling versions as I understand them and how the relate to the two main existing project file formats for .NET Core.

This PR was created to resolve [issue #5154](https://github.com/dotnet/cli/issues/5154) which points out an inconsistency that I noticed in the documentation which points out that the newer builds of the CLI Tools only work with .csproj based projects but then goes on to suggest the reader follow the project.json samples.


Petermarcu:
@jpierson , Thanks for doing this. I just saw this after I submitted a similar PR today that makes the existing samples work both ways. Thoughts? I like the simplicity of keeping them where they are and that we can just delete the project.json's when the time comes. 

https://github.com/dotnet/core/pull/417

jpierson:
I'll take a look but o I doubt it matters much. I just didn't want others to stumble the same way I have been trying to jump in and learn about .net core. For myself these changes were a way to motivate me to figure it out so that I could contribute back.

Petermarcu:
Thanks. Looks like the other changes were already merged and I heard that @blackdwarf was going to go through them and simplify the msbuild even more based on improvements that were made in this version of MSBuild. I'll go ahead an close this PR now but again, thanks for making the PR. It was definitely a fix that was needed.

PR #411, opened 2016-12-29
Fix typos in SDK preview4 version

naamunds:
/cc @blackdwarf


svick:
There already is a PR to fix those versions: https://github.com/dotnet/core/pull/408.

naamunds:
@svick Thanks for pointing that out.  I'll close this one.

Issue #363, opened 2016-11-23
Sample projects no longer build


cydhaselton:
The sample projects under samples/ no longer build according to the posted instructions.

`dotnet build` in samples/helloworld (or other directories) produces an error

`MSBUILD : error MSB1003: Specify a project or solution file. The current working directory does not contain a project or  solution file.`

There are no additional instructions on creating a project file.  Running `dotnet new` throws an error:

`Creating new C# project failed, directory already contains Program.cs`

svick:
The sample projects use project.json. It looks like you're on SDK preview3 (or newer), which switched to msbuild. To build and run those projects with your version of the SDK, you first need to run `dotnet migrate`.

The instructions could probably mention this.

cydhaselton:
Thanks for the info; I'll add the instruction when I'm in the office next

Petermarcu:
I just submitted a PR that will enable both project.json and msbuild tooling to build the samples. https://github.com/dotnet/core/pull/417 . For the transition period that should help and then we can delete the project.json's when the time is right.

Petermarcu:
I think this was fixed.

PR #417, opened 2017-01-05
Adding msbuild support to the samples

Petermarcu:
project.json and msbuild support can live together for these samples for now. We should delete the project.json's when the time is right.


Petermarcu:
@richlander

Petermarcu:
This is a fix to address this: https://github.com/dotnet/core/issues/363

blackdwarf:
A question: do we want to actually start anew and show off the easier new csproj syntax for these samples? They are pretty basic, so it shouldn't be a problem. I can make that PR.

Petermarcu:
Sure, let's merge this and then you can clean them up as a good example for people who migrate.

blackdwarf:
LGTM then

PR #432, opened 2017-01-19
Create sdk-implicit-refs.md

blackdwarf:
/cc @dsplaisted


richlander:
Please avoid the word "glob". I think that this is a jargon word.

blackdwarf:
@richlander a "glob" as a term that refers to a pattern that qualifies 0 or more items to be worked on is a pretty known term. It could be that it is jargon, but still it is something that people use.

richlander:
We should put these files lower down than in the root. It's a bit odd, but I'd suggest the 1.0 folder. maybe 1.0/sdk  folder?

blackdwarf:
@richlander changed the title. As agreed, let's merge this first and then I can move them around.

PR #437, opened 2017-01-21
Rename release note files and titles

richlander:


serhioV:
"set the <EnableDefaultCompileItems> property to false"

where , in what file should we set that property to false?

serhioV:
the phrase is from the Readme file, "Rename release note files and titles (#437)". 
Thanks! Maybe is worth specifying in the Doc(readme file) as well.

PR #439, opened 2017-01-21
.NET Core on ARM boards

gitchomik:
Add info about supporting .NET Core on other ARM board (Orange Pi PC)


Petermarcu:
Looks good. Thanks!

PR #457, opened 2017-01-29
Fix formatting

j2ghz:


Petermarcu:
I believe this was fixed already by a different PR. Thanks for doing it though!

PR #459, opened 2017-01-31
add rc3 (1.0, 1.1) to download-archive

enricosada:
The rc3 page already exists, but is not linked from download archive.
This add rc3 to both Current/LTS because contains both

@blackdwarf @kendrahavens @leecow


leecow:
lgtm

Issue #386, opened 2016-12-10
Where can I find the minimum memory requirements specification for .NET core open source?


crowdwave:
I've searched but I can't find any documentation specifying the memory requirements.

Can anyone suggest to me where the minimum memory requirement on Linux is specified?

svick:
Minimum memory requirement of what? Building .Net Core (corefx, coreclr, cli, sdk, ...)? Building your own simple console or ASP.NET Core application? Running your own simple console or ASP.NET Core application?

CodeChaser:
I came here looking for the same information.  I guess it is not documented or at the very least buried inside of another point.

crowdwave:
@CodeChaser - I am getting the impression from the answers here and elsewhere that I have asked, that .NET core effectively takes almost no memory, thus it will work on the smallest possible RAM and CPU configurations, and the needed RAM is defined by the size of your application.

Petermarcu:
We haven't officially doc'd it but we should. I'll look into that. It does run in very constrained environments including low end phones and pretty small IoT devices that are able to run Linux or Windows. Do you have the constraints you are dealing with?

crowdwave:
@Petermarcu when using any new software, the first question is "how much RAM does this need?", that answers the basic question of "how small is the machine I can run this on?".

My base expectation is that .NET is some sort of big virtual machine that needs lots of memory to get going.  That's why I ask the question.

And yes I would also like to know, well just how small can the machine be that runs this?

It seems to be assumed knowledge in the community that .NET core runs on any sized RAM.

Put more succinctly: .NET core should specify the minimum hardware requirements to run it.

CodeChaser:
Yea an Arduino ;)..   Well looks like I will be using it on a Raz PI 3 in the near future.  My goal is to see if I can get EF Core running against a SQLite db pn a PI3.   I have been using EF Core for the pass few months on a desktop, DB first.  If I remember , I believe it requires .Net Standard, so I am not sure if it is even going to work.  I need the ability use a SQL based DB such as SQLite the results of my decision tree's, as in realtime ML results.

PR #465, opened 2017-02-03
Adding basic system requirements

Petermarcu:


Petermarcu:
This should address the initial ask in https://github.com/dotnet/core/issues/386

Petermarcu:
@richlander

Documentation/minimum_requirements.md
```diff
@@ -0,0 +1,14 @@
...
+##1.X 
+ 
+The minimum available resources required for a simple standalone application are:
+ 
+- 25MB Disk Space
+- 16MB available RAM
+
+Depending on your application and performance requirements, these requirements may be higher. 
+
+*SSE2 processor instruction support is required.
```
mairaw: This seems lost in the text. Perhaps move it to the first paragraph? Also, should we link somewhere or give more info on what this means? How would a customer know if their computer has that? It's not something we typically call out in a minimum system requirements, I think.

Petermarcu: It's not something that is important to many PC users because its pretty much been the standard for a decade. It is more interesting to people who are considering porting to other processor types and IoT. For them, I think this pops enought to them.

mellinoe: This is only relevant for x86/64 processors, though. Should there be a list of architectures supported, instead, and put this as a sub-bullet under x86-64?

mellinoe: There might also be similar "min-bar" prereqs for ARM, but I'm not sure what they are.

Petermarcu:
I just removed the SSE bullet. It's one of many other requirements that could be listed but are details beyond the high level requirements most are looking for.

leecow:
LGTM

SailorMax:
@Petermarcu
"25MB Disk Space" Is it for native compiled application? "Self-contained"?
Can you also add information for "Framework-dependent"? (including framework size)
https://docs.microsoft.com/en-us/dotnet/articles/core/deploying/

PR #490, opened 2017-02-12
Fix 16.04/14.04 link mix-up for rc4 sdk download

kontomondo:
Currently the link for 14.04 downloads the 16.04 version and vice-versa.


Petermarcu:
Thanks! @leecow

PR #508, opened 2017-02-16
reference additional repos in readm.md

leecow:
resolves https://github.com/dotnet/core/issues/451 

cc/ @Petermarcu , @mairaw

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,12 @@ The core repository is the starting point to engage in and learn about the
 > * [dotnet/cli](https://github.com/dotnet/cli) - for CLI tools and questions
 > * [dotnet/corefx](https://github.com/dotnet/corefx) - for API issues and questions
 > * [dotnet/coreclr](https://github.com/dotnet/coreclr) - for runtime issues
-> * [dotnet/docs](https://github.com/dotnet/docs) - for .NET Core documentation issues and questions
-> * [dotnet/sdk](https://github.com/dotnet/sdk) - for .NET Core SDK for Visual Studio issues and questions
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,6 +9,8 @@ The core repository is the starting point to engage in and learn about the
 > * [dotnet/cli](https://github.com/dotnet/cli) - for CLI tools and questions
 > * [dotnet/corefx](https://github.com/dotnet/corefx) - for API issues and questions
 > * [dotnet/coreclr](https://github.com/dotnet/coreclr) - for runtime issues
+> * [dotnet/docs](https://github.com/dotnet/docs) - for .NET Core documentation issues and questions
```
svick: Since dotnet/docs is not .Net Core-specific, I think this should say just:

> for .NET documentation issues and questions

README.md
```diff
@@ -9,6 +9,8 @@ The core repository is the starting point to engage in and learn about the
...
+> * [dotnet/sdk](https://github.com/dotnet/sdk) - for .NET Core SDK for Visual Studio issues and questions
```
svick: The SDK is used by more than just VS, specifically the CLI. So I think this should say:

> for .NET Core SDK issues and questions

Though that doesn't really say anything to those who don't already know what the SDK is, so maybe something like:

>  for .NET Core SDK (used by Visual Studio and CLI) issues and questions

mairaw:
I agree with @svick's point on the docs. I think we should also add https://github.com/aspnet/Docs which hosts ASP.NET and ASP.NET Core docs. Other than that, LGTM>

PR #531, opened 2017-03-07
1.0.4 / 1.1.1 release notes

leecow:


richlander:
Should the SDK parts be in a separate release note? Meaning, should we have the following, too:

- release-notes/1.0/sdk-1.0.0.md
- release-notes/1.0/sdk-1.0.1.md

leecow:
My initial thought was that the minimal diff didn't warrant separate release notes.

richlander:
We could use the same scheme we do for Docker images, which is "runtime first". Maybe that's what we're doing here.

PR #542, opened 2017-03-10
Update docker image reference

kendrahavens:


release-notes/docker-image-reference.md
```diff
@@ -2,23 +2,19 @@
 
 In order for Docker containers to run dotnet cli commands, the sdk that you use to build your project with locally must match the dotnet sdk version inside your container. If you want a lighter-weight container and don’t need to use the cli tools inside the container, then you can pull the runtime images.
 
-|    Docker Images   2/07/17             |    .NET Core   Runtime version in image   |    .NET Core SDK   version in image    (NOTE: All SDK versions can target all   runtimes)    |    Full SDK   version number    |
+|    Docker Images   03/08/17            |    .NET Core   Runtime version in image   |    .NET Core SDK version in image    (NOTE: All SDK versions can target all   runtimes)      |    Full SDK   version number    |
```
richlander: Proposal on alternate column headers. I think they can be simplified and I believe one isn't needed.

|    Docker Image    |    Included .NET Core Runtime Version(s)  |    .NET Core Tools Version  |
  |----------------------------------------|---------------------------------------------------------------------------------------------|---------------------------------|
 -|    1.0.3-runtime                       |    1.0.3                                  |     *None                        |

kendrahavens: I'd prefer to continue listing the full .NET Core SDK version number in this table since it is different from the .NET Core download version. 

There is confusion when users download the 1.0.4 .NET Core SDK but they see 1.0.1 installed in their C:\Program Files\dotnet\sdk folder. They also see 1.0.1 when they run `dotnet --version` and in their Programs and Features view.

richlander: Sure, but isn't that the 4th column. The 3rd column doesn't seem useful to me (in fact, I think it is confusing).

kendrahavens: Gotcha, consolidating!

Issue #541, opened 2017-03-08
Visual Studio Preview 4 (7.0 Build 1566) has broken link for downloading .NET core


vboctor:
Just installed Visual Studio Preview 4 on my Mac and attempted to create a .NET Core Web API project, it detected that I don't have .NET core installed and directed me to Github link below to download it which is broken:
https://github.com/dotnet/core/blob/master/release-notes/rc4-download.md

I know this is not strictly a .NET core issue, but I'm not sure about the repository to report this into.  If you let me know, I will report it accordingly.

Petermarcu:
Here is the correct link. It got moved to archives: https://github.com/dotnet/core/blob/master/release-notes/download-archives/rc4-download.md

@leecow @richlander we should probably not move the docs so people referring to them aren't broken.

richlander:
Interesting. I'll ask if they are using a forward link or what. Either way, we'll get a fix in place today.

blackdwarf:
We will fix this with returning the RC4 page but we will point users to install RTW bits to keep them current.

richlander:
Fixed by #544

PR #544, opened 2017-03-10
Create rc4-download.md

blackdwarf:
Added the page to unblock #541. The link points to the official pages for download to avoid people getting stuck on older versions of the CLI/SDK.

/cc @richlander

### DIFF
--- a/release-notes/rc4-download.md
+++ b/release-notes/rc4-download.md
@@ -1,0 +1,4 @@
+.NET Core Tools download
+------------------------
+
+Thank you for you interest in .NET Core Tools. We have released v1, so please go to https://dot.net/core to download the released version. 


richlander:
Should we point to the page that was intended? If they want RC4, then we should give RC4 (right?), which BTW isn't available from the site you linked to.

https://github.com/dotnet/core/blob/master/release-notes/download-archives/rc4-download.md

blackdwarf:
@richlander I would like to avoid people installing RC4 when we released RTW. The main rationale would be that I would like to decrease potential user frustration when they run into issues that we actually solved in RTW.

richlander:
You should validate that the VS 4 Mac folks are OK with that. 

In general, I agree. Good point. We should decide on our long-term plan here with the VS 4 Mac folks.

/cc @mhutch

blackdwarf:
I already asked the VS 4 Mac folks on whether this would be OK which is why I made this change. 

I believe that the long-term plan should be that VS 4 Mac is tested against release versions of the Core SDK with the preference for stable version where/if possible. 

Thoughts?

PR #555, opened 2017-03-14
Update RaspberryPiInstructions.md

hqueue:
Add more detailed step.

### DIFF
--- a/samples/RaspberryPiInstructions.md
+++ b/samples/RaspberryPiInstructions.md
@@ -9,9 +9,10 @@ These steps have been tested on a RPi 2 and RPi 3 with Linux and Windows.
 
 * [Install .NET Core 2.0 SDK](https://github.com/dotnet/cli/tree/master) into a supported developer configuration.
 
-* From the terminal/commandline create a folder named helloworld and go into it.
-* Run `dotnet new console -n helloworld`.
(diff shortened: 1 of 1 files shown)


samples/RaspberryPiInstructions.md
```diff
@@ -11,6 +11,7 @@ These steps have been tested on a RPi 2 and RPi 3 with Linux and Windows.
 
 * From the terminal/commandline create a folder named helloworld and go into it.
 * Run `dotnet new console -n helloworld`.
+* Change directory into `helloworld` and you can find helloworld.csproj file.
```
Petermarcu: If we're going to tell people here to change to the helloworld directory, we should delete the words telling people to do that from two lines higher.

Petermarcu:
One piece of feedback, then I think this is good to merge. Thanks!

samples/RaspberryPiInstructions.md
hqueue: You are right, I will update it. 
With recent SDK, if I ran `dotnet new console -n helloworld`, a new directory `helloworld` is created under current directory. I will update this paragraph.

hqueue:
@Petermarcu PTAL. I've removed `-n helloworld` option from dotnet command, since this option make a new  `helloworld` directory under current directory. And also refine the paragraph.

PR #573, opened 2017-03-24
Raspberry Pi Instructions addition

MaxMommersteeg:
The current instructions were hard to follow for me. I did not know what version I needed to insert at `RuntimeFrameworkVersion`. Running the `dotnet` command returned the correct value.

Adding the new package source is essential for resolving packages before restoring the project.

### DIFF
--- a/samples/RaspberryPiInstructions.md
+++ b/samples/RaspberryPiInstructions.md
@@ -11,7 +11,7 @@ These steps have been tested on a RPi 2 and RPi 3 with Linux and Windows.
 
 * From the terminal/commandline create a folder named helloworld and go into it.
 * Run `dotnet new console -n helloworld`.
-* Run `dotnet` to get the current RuntimeFrameworkVersion.
+* Run `dotnet` and note the current RuntimeFrameworkVersion.
 * Edit the helloworld.csproj file to look like this (Note the version may be newer for RuntimeFrameworkVersion but not lower than 2.0.0-beta-001620-00).
 
 ```
--- a/samples/RaspberryPiInstructions.md
+++ b/samples/RaspberryPiInstructions.md
@@ -27,7 +27,7 @@ These steps have been tested on a RPi 2 and RPi 3 with Linux and Windows.
 </Project>
 ```
 
-* Add a new package source to the project with source: `https://dotnet.myget.org/F/dotnet-core/api/v3/index.json`.
+* Make sure you have a nuget.config file next to your csproj that includes the dotnet-core myget feed: `<add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />`.
 
 * Run `dotnet restore`.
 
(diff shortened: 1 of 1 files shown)


Petermarcu:
Thanks. I'll merge the updates. I'm guessing people will still be a little confused about where to put the package source. If you use the 2.0 SDK, it is supposed to drop a nuget.config with that source already there but it seems many people are doing something different. Probably using the 1.0 SDK?

samples/RaspberryPiInstructions.md
```diff
@@ -26,6 +27,8 @@ These steps have been tested on a RPi 2 and RPi 3 with Linux and Windows.
 </Project>
 ```
 
+* Add a new package source to the project with source: `https://dotnet.myget.org/F/dotnet-core/api/v3/index.json`.
```
Petermarcu: I'm actually thinking it may be best to give an example of a complete but minumal NuGet.config here telling people that they should have a file by that name next to their csproj. Can you make that update?

MaxMommersteeg: Today I tried the steps over again.

So I reinstalled Microsoft .NET Core 2.0.0 - SDK Preview 1. I expected `2.0.0` when running the `dotnet --version` command,  however it returned `1.0.0`. I used the following link to install the .NET Core SDK: [Windows x64 - Installer](https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/dotnet-dev-win-x64.latest.exe)

> Probably using the 1.0 SDK?

So yes, you are correct. I was not presented with a NuGet.config file probably due to the fact that I was using .NET Core 1 SDK. Others could be running into the same issue. I had to manually add the package source with Visual Studio 2017. Perhaps the first line should be elaborated on, since it only points to the repository without giving a 'how-to':
>Install .NET Core 2.0 SDK into a supported developer configuration

Creating multiple paths for multiple environments could be an option, but could confuse the reader even more. I'm not sure what the correct approach is.

Petermarcu: I think you are hitting the problem that the 2.0 SDK isn't taking over for the 1.0 SDK because its a pre-release. @livarcocc I believe is working to make 2.0 take precedence over 1.0. So, I think you are right that a lot of people will be generating the console app based on 1.0 because of this.

You can add a line that says:

Make sure you have a nuget.config file next to your csproj that includes the dotnet-core myget feed.

`<add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />`

Complete file example:
```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
  </packageSources>
</configuration>
```

Petermarcu:
@MaxMommersteeg , did you wan to finish this up? Looks like there are merge conflicts now. If not, I'm happy to take what you had and get it merged in.

MaxMommersteeg:
@Petermarcu Sorry for my late response. I will update this pull request tomorrow.

Petermarcu:
Thanks!

PR #577, opened 2017-03-28
Updating versions and support.

Petermarcu:


Petermarcu:
@leecow, look good?

leecow:
LGTM

lemmaa:
@Petermarcu What does 'support' mean on this document? Does it mean customer service or partner service? If so, it would be better to remove the `[S]` tag from Ubuntu/arm32. We do our best to ensure that .NET Core will work properly on Ubuntu/arm32, but we have no plans to officially support Ubuntu/arm32 yet. For now, only Tizen is on that range. We will do our best to be part of the .NET community. :)

Petermarcu:
@lemmaa, I just updated it to community based on that. Thanks!

lemmaa:
@Petermarcu, Thanks!

PR #598, opened 2017-04-20
Add initial Source Link spec

chuckries:


Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Why Source Link?
+Most debugging is done against a locally built source code on a developer's machine. In this scenario, matching the binary to the source is not difficult. However, there are many debugging scenarios where the original source code is not immediately availble. Two good examples of this are debugging crash dumps or third party libraries. In these scenarios, it can be very difficult for a developer to acquire the exact source code that was built to produce the binary they are debugging. Source Link solves this problem by embedding unique information about the source code (such as a git commit hash) in the Portable PDB. Diagnostic tools, such as debuggers, can use this unique information to retrieve the original source code from hosted services (such as [Github](https://github.com)).
```
chuckries: >a [](start = 31, length = 1)

remove

chuckries:
@gregg-miskelly @tmat @caslan

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,103 @@
...
+```json
+{
+    "documents": {
+        "C:\\src\\CodeFormatter\\src\\CodeFormatter\\Program.cs":               "https://raw.githubusercontent.com/dotnet/codeformatter/bcc51178e1a82fb2edaf47285f6e577989a7333f/src/CodeFormatter/Program.cs",
+        "C:\\src\\CodeForamtter\\src\\CodeFormatter\\CommandLinerParser.cs":    "https://raw.githubusercontent.com/dotnet/codeformatter/bcc51178e1a82fb2edaf47285f6e577989a7333f/src/CodeFormatter/CommandLineParser.cs",
+        "C:\\src\\CodeFormatter\\src\\DeadRegions\\Program.cs":                 "https://raw.githubusercontent.com/dotnet/codeformatter/bcc51178e1a82fb2edaf47285f6e577989a7333f/src/DeadRegions/Program.cs",
+        "C:\\src\\CodeFormatter\\src\\DeadRegions\\Options.cs":                 "https://raw.githubusercontent.com/dotnet/codeformatter/bcc51178e1a82fb2edaf47285f6e577989a7333f/src/DeadRegions/Options.cs",
+        "C:\\src\\CodeFormatter\\src\\DeadRegions\\OptionParser.cs":            "https://raw.githubusercontent.com/dotnet/codeformatter/bcc51178e1a82fb2edaf47285f6e577989a7333f/src/DeadRegions/OptionParser.cs",
+        ...
```
tmat: Github highlights this as an error if using the text type "json"

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,103 @@
...
+        ```json
+        {
+            "documents": {
+                "/FOO/*": "http://.../X/*",
+                "/foo/*": "http://.../Y/*"
+            }
+        }
+        ```
+    - The current implentation does not support a Source Link entry that contains a URL that will cause a redirect. If the URL in the Source Link map causes a redirection, downloading the source file will fail.
+- [C# Extension for Visaul Studio Code](https://marketplace.visualstudio.com/items?itemName=ms-vscode.csharp)
```
tmat: Typo: "Visaul"

tmat: Add a link to https://github.com/ctaggart/SourceLink

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Why Source Link?
+Most debugging is done against locally built source code on a developer's machine. In this scenario, matching the binary to the source is not difficult. However, there are many debugging scenarios where the original source code is not immediately availble. Two good examples of this are debugging crash dumps or third party libraries. In these scenarios, it can be very difficult for a developer to acquire the exact source code that was built to produce the binary they are debugging. Source Link solves this problem by embedding unique information about the source code (such as a git commit hash) in the Portable PDB. Diagnostic tools, such as debuggers, can use this unique information to retrieve the original source code from hosted services (such as [Github](https://github.com)).
```
tmat: I wouldn't specifically called out "Portable PDB". We'll have support for Windows PDB as well. Perhaps as a not you can add that VS 2017 only supports this feature in Portable PDBs.

Documentation/diagnostics/source_link.md
chuckries: fixed.

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,103 @@
...
+
+## MSBuild Task
+TBD
+### Example
+`TBD`
+
+## Source Link File Specification
+The `/sourcelink:<file>` compiler flag will embed a JSON configuration file in the Portable PDB. This configuration file would be generated as part of the build process. A default MSBuild task is already availble for Source Link generation (see above), but the generation can be customized to suit the needs of the project. The JSON configuration file contains a simple mapping of local file path to URL where the source file can be retrieved via http or https. A debugger would retrieve the originfal file path of the current location from the Portable PDB, look this path up in the Source Link map, and use the resulting URL to download the source file.
```
tmat: Again, I'd remove "Portable"

tmat: Typo: "availble"

tmat: Typo: "originfal"

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,103 @@
...
+    "description": "A mapping of source file paths to URI's",
+    "type": "object",
+    "properties": {
+        "documents": {
+            "type": "object",
+            "minProperties": 1,
+            "additionalProperties": {
+                "type": "string"
+            },
+            "descripttion": "Each document is defined by a file path and a URI. Original source file paths are compared
```
tmat: Typo: descripttion

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,103 @@
...
+        "documents": {
+            "type": "object",
+            "minProperties": 1,
+            "additionalProperties": {
+                "type": "string"
+            },
+            "descripttion": "Each document is defined by a file path and a URI. Original source file paths are compared 
+                             case insensitive to documents and the resulting URI is used to download source. The document 
+                             may contain an asterisk to represent a wildcard in order to match anything in the asterisk's 
+                             location. If the document contains an asterisk, the URI must also contain an aterisk. The
```
tmat: Isn't the asterisk only allowed at the end of the document path?

tmat: Typo: aterisk

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,103 @@
...
+```
+In this example:
+- All files under directory `bar` will map to a relative URL beginning with `http://MyBarDomain.com/src/`.
+- All files under directory `foo` will map to a relative URL beginning with `http://MyFooDomain.com/src/` **EXCEPT** `foo/specific.txt` which will map to `http://MySpecificFoodDomain.com/src/specific.txt`.
+- All other files anywhere under the `src` directory will map to a relative url beginning with `http://MyDefaultDomain.com/src/`.
+
+## Tooling
+- [Visual Studio 2017](https://www.visualstudio.com/downloads/)
+    - First debugger to support Source Link
+    - The current implementation uses a case insensitive string comparison between the file path and the Source Link entry. Using multiple entries that differ only in casing is not supported and the behavior is undefined. Here is an example of an unsupported Source Link JSON:
```
tmat: I wouldn't say the behavior is undefined. It is defined - the first matching one wins.

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,103 @@
...
+        ```json
+        {
+            "documents": {
+                "/FOO/*": "http://.../X/*",
+                "/foo/*": "http://.../Y/*"
+            }
+        }
+        ```
+    - The current implentation does not support a Source Link entry that contains a URL that will cause a redirect. If the URL in the Source Link map causes a redirection, downloading the source file will fail.
```
tmat: Typo: "implentation"

Documentation/diagnostics/source_link.md
tmat: We should include exact rules the debugger implementation uses when validating the data -- e.g. can't have more than one ```*```, if present ```*``` must be the last character of the document name, etc.

chuckries: I believe you are correct, I will double check. There was a time when I thought VSTS urls would require the relative path to be placed in the middle of a URL.

tmat:
@ctaggart FYI

Documentation/diagnostics/source_link.md
chuckries: We do a sort of the map entries though, they are not matched in the exact order they appear in the json file. I wasn't 100% sure how our sort would treat case. I will double check.

gregg-miskelly:
I don't have any additional feedback.

Documentation/diagnostics/source_link.md
tmat: Shouldn't the sort be stable?

Documentation/diagnostics/source_link.md
ctaggart: *Please* support redirects if not already supported. They work in source server support.

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,103 @@
+# Source Link
+[Portable PDB](https://github.com/dotnet/core/blob/master/Documentation/diagnostics/portable_pdb.md) introduces a new developer productivity feature called Source Link. Source Link allows a Portable PDB to embed unique information about the assembly's original source code during compilation. The [Roslyn](https://github.com/dotnet/roslyn/wiki/Roslyn%20Overview) compiler flag for Source Link is `/sourcelink:<file>`.
```
ctaggart: It is no longer just Portable PDB files, right?

chuckries: @tmat suggested removing all 'Portable' references. It is being resolved.

Documentation/diagnostics/source_link.md
chuckries: The current implementation uses [qsort](http://www.cplusplus.com/reference/cstdlib/qsort/) and compares the strings based on length. qsort does not guarantee the ordering of equivalent elements.
> The order of equivalent elements is undefined.

Documentation/diagnostics/source_link.md
ctaggart: Yeah, hoping it doesn't have to be the last character. That doesn't sound right.

Documentation/diagnostics/source_link.md
tmat: It's coming, but we are not sure yet when we will be able to ship the Windows PDB support.

ctaggart:
Can you please add specs on authentication support? https://github.com/dotnet/roslyn/issues/12759#issuecomment-282793617

Any idea when that may be supported. Those of us at Tachyus **really** want to enable source link for our private GitHub repositories.

Documentation/diagnostics/source_link.md
tmat: Sounds like something we should fix.

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Why Source Link?
+Most debugging is done against locally built source code on a developer's machine. In this scenario, matching the binary to the source is not difficult. However, there are many debugging scenarios where the original source code is not immediately availble. Two good examples of this are debugging crash dumps or third party libraries. In these scenarios, it can be very difficult for a developer to acquire the exact source code that was built to produce the binary they are debugging. Source Link solves this problem by embedding unique information about the source code (such as a git commit hash) in the Portable PDB. Diagnostic tools, such as debuggers, can use this unique information to retrieve the original source code from hosted services (such as [Github](https://github.com)).
+
+## MSBuild Task
+TBD
```
ctaggart: A simple example that will work with GitHub or BitBucket and is highly configurable would be adding a `Directory.Build.props` with:
``` xml
<Project>
  <PropertyGroup>
    <DebugType>portable</DebugType>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="SourceLink.Create.CommandLine" Version="2.1.0" PrivateAssets="all" />
  </ItemGroup>
</Project>
```
That is all it takes. Or you can [do it manually](https://github.com/Microsoft/visualfsharp/blob/818c6277c8360cd095f404ee158a519d2cc72649/src/FSharpSource.targets#L517-L537). The result of those two is the same.

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,103 @@
...
+    },
+    "required": ["documents"]
+}
+```
+
+### Examples
+The simplest Source Link JSON would map every file in the project directory to a URL with a matching relative path. The following example shows what the Source Link JSON would look like for the .NET [Code Formatter](https://github.com/dotnet/codeformatter). In this example, Code Formatter has been cloned to `C:\src\CodeFormatter\` and is being built at commit `bcc51178e1a82fb2edaf47285f6e577989a7333f`.
+```json
+{
+    "docuemnts": {
```
svick: Typo: docuemnts → documents.

The same typo also appears in the next example.

Documentation/diagnostics/source_link.md
ctaggart: You should be able to adjust the properties of that MSBuild task to support Hg or Subversion, etc. Let me know if you have an example in mind.

Documentation/diagnostics/source_link.md
chuckries: resolved

chuckries:
@ctaggart Authentication is planned but not yet supported. cc @caslan

Documentation/diagnostics/source_link.md
chuckries: I will clarify this, the exact rules are these:
1. The only acceptable wildcard is one and only one '*', which if present will be replaced by a relative path
2. If the filepath does not contain a *, the uri cannot contain a * and if the filepath contains a * the uri must contain a *
3. If the filepath contains a *, it must be the final character
4. If the uri contains a *, it may be anwhere in the uri

Documentation/diagnostics/source_link.md
chuckries: I will log a bug on the debugger for this, but I am leaving it here for now.

ctaggart:
@chuckries @caslan It is something that will really help all the developers that share private binaries with each other within a company. It seems like a *relatively* easy feature to make for such a big win.

Documentation/diagnostics/source_link.md
chuckries: @ctaggart are these only supported for dotnet cli projects?

ctaggart: No, both `SourceLink.Create.CommandLine` & `SourceLink.Create.GitHub` work on both dotnet and framework. See the 2.1 test plan for several tests https://github.com/ctaggart/SourceLink/issues/188. The latter makes use of the `dotnet sourcelink-git` tool, and has more features. The former just uses command line tools and is easy to get started with.

Documentation/diagnostics/source_link.md
chuckries: @tmat Where should this link be added? Is this what we want to use as an example of the msbuild task?

tmat: Yeah, I'd add it to the build task section as an example of an OSS tool that generates source link.

chuckries:
@ctaggart @tmat I updated the msbuild example with a pointer to https://github.com/ctaggart/SourceLink. Please take one last look.

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,136 @@
...
+
+    <Exec Command="git merge-base --fork-point refs/remotes/CodeFormatter/master HEAD" ConsoleToMsBuild="true" Condition = " '$(TF_BUILD)' == 'True' ">
+      <Output TaskParameter="ConsoleOutput" PropertyName="LatestCommit" />
+    </Exec>
+
+    <WriteLinesToFile File="$(IntermediateOutputPath)source_link.json" Overwrite="true" Lines='{"documents": { "$(SourceLinkRoot)\\*" : "$(RemoteUri.Replace(".git", "").Replace("github.com", "raw.githubusercontent.com"))/$(LatestCommit)/*" }}' />
+  </Target>
+...
+```
+There is also an open source tool available for generating Source Link configuration at build time. See https://github.com/ctaggart/SourceLink for the most up to date information.
```
gregg-miskelly: I wonder if this should be at the top of the section rather than at the end since we probably would like folks to use SourceLink rather than copy/pasting the wheel.

chuckries: fixed.

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,137 @@
...
+
+    <Exec Command="git config --get remote.origin.url" ConsoleToMsBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="RemoteUri" />
+    </Exec>
+
+    <Exec Command="git rev-parse HEAD" ConsoleToMsBuild="true" Condition = " '$(TF_BUILD)' != 'True' ">
+      <Output TaskParameter="ConsoleOutput" PropertyName="LatestCommit" />
+    </Exec>
+
+    <Exec Command="git merge-base --fork-point refs/remotes/CodeFormatter/master HEAD" ConsoleToMsBuild="true" Condition = " '$(TF_BUILD)' == 'True' ">
```
gregg-miskelly: Should we add:
```<!-- NOTE: Replace 'CodeFormatter' with the appropriate git remote -->```

Or replace it with 'origin' since it seems likely that that is the right remote, and what is used on line 26?

File: Documentation/diagnostics/source_link.md [hunk 15]
```diff
@@ -0,0 +1,137 @@
...
+
+## Source Link File Specification
+The `/sourcelink:<file>` compiler flag will embed a JSON configuration file in the PDB. This configuration file would be generated as part of the build process. The JSON configuration file contains a simple mapping of local file path to URL where the source file can be retrieved via http or https. A debugger would retrieve the original file path of the current location from the PDB, look this path up in the Source Link map, and use the resulting URL to download the source file.
+
+### MSBuild Example
+There is an open source tool available for generating Source Link configuration at build time. See https://github.com/ctaggart/SourceLink for the most up to date information.
+
+Generating and embedding a Source Link file can also be done manually with an MSBuild Target. The following is an example of enabling source link in a .csproj for a project using git.
```
gregg-miskelly: Should we say 'github' instead of 'git' here. Since this example is using GitHub URLs.

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,137 @@
...
+
+    <Exec Command="git rev-parse HEAD" ConsoleToMsBuild="true" Condition = " '$(TF_BUILD)' != 'True' ">
+      <Output TaskParameter="ConsoleOutput" PropertyName="LatestCommit" />
+    </Exec>
+
+    <Exec Command="git merge-base --fork-point refs/remotes/CodeFormatter/master HEAD" ConsoleToMsBuild="true" Condition = " '$(TF_BUILD)' == 'True' ">
+      <Output TaskParameter="ConsoleOutput" PropertyName="LatestCommit" />
+    </Exec>
+
+    <WriteLinesToFile File="$(IntermediateOutputPath)source_link.json" Overwrite="true" Lines='{"documents": { "$(SourceLinkRoot)\\*" : "$(RemoteUri.Replace(".git", "").Replace("github.com", "raw.githubusercontent.com"))/$(LatestCommit)/*" }}' />
```
gregg-miskelly: Suggest adding a comment:
```<!-- Write out the source file for this project to point at raw.githubusercontent.com -->```

File: Documentation/diagnostics/source_link.md [hunk 17]
```diff
@@ -0,0 +1,137 @@
+# Source Link
+Source Link is a developer productivity feature that allows unique information about an assembly's original source code to be embedded in its PDB during compilation. The [Roslyn](https://github.com/dotnet/roslyn/wiki/Roslyn%20Overview) compiler flag for Source Link is `/sourcelink:<file>`.
+
+## Why Source Link?
+Most debugging is done against locally built source code on a developer's machine. In this scenario, matching the binary to the source is not difficult. However, there are many debugging scenarios where the original source code is not immediately available. Two good examples of this are debugging crash dumps or third party libraries. In these scenarios, it can be very difficult for a developer to acquire the exact source code that was built to produce the binary they are debugging. Source Link solves this problem by embedding unique information about the source code (such as a git commit hash) in the PDB. Diagnostic tools, such as debuggers, can use this unique information to retrieve the original source code from hosted services (such as [Github](https://github.com)).
```
bleroy: Nit: the proper capitalization is "GitHub".

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Source Link File Specification
+The `/sourcelink:<file>` compiler flag will embed a JSON configuration file in the PDB. This configuration file would be generated as part of the build process. The JSON configuration file contains a simple mapping of local file path to URL where the source file can be retrieved via http or https. A debugger would retrieve the original file path of the current location from the PDB, look this path up in the Source Link map, and use the resulting URL to download the source file.
+
+### MSBuild Example
+There is an open source tool available for generating Source Link configuration at build time. See https://github.com/ctaggart/SourceLink for the most up to date information.
```
bleroy: Can we have this also available on this repo, instead of on your GH?

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,137 @@
...
+### Examples
+The simplest Source Link JSON would map every file in the project directory to a URL with a matching relative path. The following example shows what the Source Link JSON would look like for the .NET [Code Formatter](https://github.com/dotnet/codeformatter). In this example, Code Formatter has been cloned to `C:\src\CodeFormatter\` and is being built at commit `bcc51178e1a82fb2edaf47285f6e577989a7333f`.
+```json
+{
+    "documents": {
+        "C:\\src\\CodeFormatter\\*": "https://raw.githubusercontent.com/dotnet/codeformatter/bcc51178e1a82fb2edaf47285f6e577989a7333f/*"
+    }
+}
+```
+In this example, the file `C:\src\CodeFormatter\src\Microsoft.DotNet.CodeFormatting\FormattingEngine.cs` would be retrieved from https://raw.githubusercontent.com/dotnet/codeformatter/bcc51178e1a82fb2edaf47285f6e577989a7333f/src/Microsoft.DotNet.CodeFormatting/FormattingEngine.cs
```
bleroy: Nit: backticks missing on the second URL.

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,137 @@
...
+```json
+{
+    "documents": {
+        "/usr/local/src/CodeFormatter/*": "https://raw.githubusercontent.com/dotnet/codeformatter/bcc51178e1a82fb2edaf47285f6e577989a7333f/*"
+    }
+}
+```
+
+Source Link configuration can also contain absolute mappings from file paths to URLs without an asterisk. Using the Code Formatter example from above, every source file could be mapped explicitly:
+```
```
bleroy: Missing `json` after the three backticks.

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,137 @@
...
+In this example:
+- All files under directory `bar` will map to a relative URL beginning with `http://MyBarDomain.com/src/`.
+- All files under directory `foo` will map to a relative URL beginning with `http://MyFooDomain.com/src/` **EXCEPT** `foo/specific.txt` which will map to `http://MySpecificFoodDomain.com/src/specific.txt`.
+- All other files anywhere under the `src` directory will map to a relative url beginning with `http://MyDefaultDomain.com/src/`.
+
+## Tooling
+- [Visual Studio 2017](https://www.visualstudio.com/downloads/)
+    - First debugger to support Source Link.
+    - Only supports the [Portable PDB](https://github.com/dotnet/core/blob/master/Documentation/diagnostics/portable_pdb.md) format.
```
bleroy: Should be "Only supports SourceLink with the [Portable PDB](...) format."

Documentation/diagnostics/source_link.md
```diff
@@ -0,0 +1,137 @@
...
+    - The current implementation uses a case insensitive string comparison between the file path and the Source Link entry. Using multiple entries that differ only in casing is not supported; the first entry will be used and the second entry that differs only by case will be ignored. Here is an example of an unsupported Source Link JSON:
+        ```json
+        {
+            "documents": {
+                "/FOO/*": "http://.../X/*",
+                "/foo/*": "http://.../Y/*"
+            }
+        }
+        ```
+    - The current implementation does not support a Source Link entry that contains a URL that will cause a redirect. If the URL in the Source Link map causes a redirection, downloading the source file will fail.
```
bleroy: Is there a plan to fix that eventually? It looks like something that could potentially lead to big problems down the road since it's common practice to refactor web sites with permanent redirects to maintain old URL schemes. If GitHub were to do something like that, it would break debugging on lots of libraries.

Documentation/diagnostics/source_link.md
chuckries: intentional, the `...` will show in red error color if json is used.

Documentation/diagnostics/source_link.md
chuckries: We have a debugger bug to fix this.

Documentation/diagnostics/source_link.md
chuckries: fixed

Documentation/diagnostics/source_link.md
chuckries: fixed

Documentation/diagnostics/source_link.md
chuckries: fixed

Documentation/diagnostics/source_link.md
chuckries: I do not think it makes sense to try to replicate the documentation here.

Documentation/diagnostics/source_link.md
bleroy: Ah, well, in that case, you can use `js` instead and put it in a comment: `// ...`.

Documentation/diagnostics/source_link.md
chuckries: replaced with origin

Documentation/diagnostics/source_link.md
chuckries: fixed.

Documentation/diagnostics/source_link.md
chuckries: fixed.

Documentation/diagnostics/source_link.md
chuckries: Turns out redirection does work.

chuckries:
I was mistaken about redirect. .NET's HttpListener supports auto redirection. I did a simple test with a local server that would redirect to github and VS 2017 handles the redirection correctly. I will remove this from the issues.

Documentation/diagnostics/source_link.md
bleroy: Oh, sorry, you're right. For some reason I misread this.

tmat:
@ctaggart I see SourceLink build task still depends on ```dotnet cli```: https://github.com/ctaggart/SourceLink/blob/v2/SourceLink.Create.GitHub/GitCreateTask.cs#L42
That means it can't be used in repos that don't run their build thru ```dotnet cli```. 

Another issue: in repos that build thru ```dotnet cli``` the dotnet tool doesn't necessarily need to be on search paths. Can we avoid the dependency on dotnet completely?

tmat:
@ctaggart I think ideally https://github.com/ctaggart/SourceLink/blob/v2/dotnet-sourcelink-git/Program.cs would be an msbuild task that can be called directly from msbuild as well as from a dotnet cli tool.

ctaggart:
@tmat Yes, I made SourceLink.Create.CommandLine based on your feedback and feedback from GitLink authors and others. It simply uses the `git` command line. Rx.NET and Paket have both adopted it. That will be the one I promote going forward. I'll update the documentation soon.

tmat:
@ctaggart Ah, I see. I haven't noticed that project. Will check it out.

chuckries:
@tmat I don't have permission to merge this so please hit the button when you are happy with the doc.

tmat:
@ctaggart Cool. Seems to be working :)

tmat:
👍

PR #605, opened 2017-04-28
Wrong link

karolklp:


imiuka:
heya, pull request #603 and this pull request (#605) are fixes for the same issue (with #603's scope being a little bigger). we'll only need to merge one of them.

karolklp:
Ahh, ok. 
Closing then.

PR #616, opened 2017-05-07
Supported OS Versions clarifications

karelz:
* Make logical link/heading to 'Supported OS Versions'
* Integrate unclear section 'Architecture' into existing topics above
* Add Unsupported OS list

cc @Petermarcu @danmosemsft


roadmap.md
```diff
@@ -111,7 +110,15 @@ Mac OS X                      | 10.12+                        | x64           |
 * [S] This is supported by Samsung.
 * [C] This is supported by the Community
 
-Version 1.0 OS Support:
+### Unsuported OS Versions
```
danmosemsft: "unsupported"

karelz: Thanks! I actually changed it to `Community-support`, which IMO better captures the real status.
I also did minor text tweaks, can you please check them Dan? Thanks!

Petermarcu:
Any reason we don't just add the community distro's just to the list with the "C" label as community supported?

Perhaps we should create a status page or issue for each distro to list current known issues and things rather than having to update the main roadmap as bugs get fixed, etc?

karelz:
@Petermarcu I think it would be great to distinguish distros which already have something existing + community support vs. those which are still "in progress" (e.g. Arch Linux).
I'll take a stab at that, then then ping you for additional feedback ...

karelz:
@Petermarcu tables merged - what do you think now? Let me know if you have suggestions for further tweaks.

Petermarcu:
I think we can go with this.

PR #650, opened 2017-05-11
2.0.0 Preview 1 contributors list

leecow:


swgillespie:
So if anyone reading this is like me and you're surprised to see your personal email show up for commits that should be attached to a work email, I learned today that GitHub by default will attach your account's primary email address to any squash commits it introduces on your behalf (through the squash PR button), regardless of the email addresses attached to the commits being squashed. I had to go to my GitHub settings and set my primary email to be my work email to fix this for future commits.

PR #662, opened 2017-05-24
Update license

richlander:
- Change license holder to ".NET Foundation and Contributors"
- Add file extension to match other license files for .NET Core.

### DIFF
--- a/LICENSE.TXT
+++ b/LICENSE.TXT
@@ -1,6 +1,8 @@
 The MIT License (MIT)
 
-Copyright (c) Microsoft Corporation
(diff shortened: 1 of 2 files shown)


GitMensch:
@richlander It looks like "All rights reserved." and the following "Permission is hereby granted" do conflict.
Maybe that somehow slipped in?

Please also see the "official MIT license" at https://opensource.org/licenses/MIT which is also referenced in https://en.wikipedia.org/wiki/MIT_License - and no "rights reserved" in there (but commonly a copyright year).

PR #678, opened 2017-06-08
updating samples as coreapp1.0 so they can run anywhere. 

shanselman:
updating samples as coreapp1.0 so they can run anywhere. Updating csprojs, removing project.jsons


Petermarcu:
LGTM

shanselman:
:shipit:

PR #705, opened 2017-06-28
2.0 Preview 2 contributor and commit docs

leecow:


g0t4:
I like the idea of having a single place to go to see everything that happened but this list is daunting. Commits are too granular and quite often irrelevant or indecipherable.

I really like what Docker does (https://github.com/moby/moby/releases/tag/v17.03.2-ce). They link to pull requests (maybe issues too) and only list substantial changes. They also group related PRs and I believe they sanitize a one-liner per change for quick consumption. This makes it incredibly easy to keep up with all the change in Docker and its related components. I would love to have the same for .NET Core and ASP.NET Core, etc.

PR #727, opened 2017-07-03
Add .NET Framework dependency sample

richlander:
- Hello world sample with a .NET Framework 4.6.1 dependency


Petermarcu:
And you can't make the project look like an SDK style project? The one in there is pretty noisy.

PR #764, opened 2017-07-20
Add build pages

richlander:
The feedback on #761 made me re-think my approach. I like this idea better.

/cc @Petermarcu @livarcocc


Petermarcu:
LGTM

PR #778, opened 2017-07-27
Fix duplicate link in telemetry data tsv

edgan8:
2016 Q3 previously had a duplicate copy of the 2016 Q4 link. Changed it to link to the correct file.


edgan8:
@richlander

PR #852, opened 2017-08-16
Typo

ajryan:
http://www.thefreedictionary.com/comprise


karelz:
Sounds reasonable (thanks for the link though!) ... cc @richlander who is language expert.
Overall I find the word weird and harder to understand for ESLs, but I think it is ok.

PR #878, opened 2017-08-22
Add linker instructions

richlander:
/cc @russellhadley

### DIFF
--- a/samples/linker-instructions-advanced.md
+++ b/samples/linker-instructions-advanced.md
@@ -1,6 +1,6 @@
 # Using IL Linker Advanced Features
 
-This document described the more advanced features for the IL Linker and provides more insight into how it functions.
+This document describes the more advanced features for the IL Linker and provides more insight into how it functions.
 
 The basic features and instructions for the linker are described in the [Using the .NET IL Linker](linker-instructions.md) document.
 
@@ -12,7 +12,7 @@ The dynamic features of .NET make it hard for this analysis to catch all cases i
 
 ## Using the tasks package
 
-The linker tasks package ([ILLink.Tasks](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Illink.Tasks)) contains MSBuild targets that become a part of the referencing project's build, using the mechanism described in the nuget docs.
+The linker tasks package ([ILLink.Tasks](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Illink.Tasks)) contains MSBuild targets that become a part of the referencing project's build, using the [mechanism described in the nuget docs](https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#including-msbuild-props-and-targets-in-a-package).
 
 When this package is referenced in a project, the publish target is augmented with additional targets that run the linker on the app's assemblies before they are placed in the publish directory.
 
--- a/samples/linker-instructions-advanced.md
+++ b/samples/linker-instructions-advanced.md
@@ -60,4 +60,4 @@ The ILLink task can be invoked just like any other msbuild task. For example, it
 </Target>
 ```
 
-Here, the ItemGroups AssemblyFilesToLink, LinkerRootAssemblies,and LinkerRootDescriptors would be defined elsewhere in the project, and the ExtraArgs input consists of flags described in the mono/linker documentation. By default, the RootAssemblyNames are rooted as if illink had been called with -a RootAssemblyName1 -a RootAssemblyName2 ... as arguments, but this behavior is subject to change and should not be relied upon.
+Here, the ItemGroups `AssemblyFilesToLink`, `LinkerRootAssemblies`, and `LinkerRootDescriptors` would be defined elsewhere in the project, and the `ExtraArgs` input consists of flags described in the [mono/linker documentation](https://github.com/mono/linker/blob/master/linker/README). By default, the `RootAssemblyNames` are rooted as if illink had been called with `-a RootAssemblyName1 -a RootAssemblyName2` ... as arguments, but this behavior is subject to change and should not be relied upon.
--- a/samples/linker-instructions-advanced.md
+++ b/samples/linker-instructions-advanced.md
@@ -1,0 +1,63 @@
+# Using IL Linker Advanced Features
+
+This document described the more advanced features for the IL Linker and provides more insight into how it functions.
+
+The basic features and instructions for the linker are described in the [Using the .NET IL Linker](linker-instructions.md) document.
+
+## How the linker works
+
+At a high level, the linker uses a mark-and-sweep algorithm to remove unused code: starting from some set of roots in the code (roots can be classes, methods, properties), the linker scans the IL using mono/cecil to look for code that gets called in other functions, classes, and dlls, marking the code that is reachable from the set of roots as it goes. In the sweep phase, the linker will scan through all of the code, removing any parts that weren't marked in the mark phase.
+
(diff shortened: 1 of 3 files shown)


samples/linker-instructions.md
```diff
@@ -0,0 +1,64 @@
+# Using the .NET IL Linker
+
+The .NET team has built a linker to reduce the size of applications. In trivial cases, it can reduce the size of applications by 50%. The size wins may be more favorable or more moderate for larger applications.
```
swaroop-sridhar: Should we say here that ILLiker is built on mono-linker?

samples/linker-instructions.md
```diff
@@ -0,0 +1,64 @@
...
+
+There are two straightforward approaches to determining the benefit of the linker.
+
+* Publish an application with the linker enabled and pass the `/p:ShowLinkerSizeComparison=true` flag. The output will describe the benefit. For example:
+  * `dotnet publish -c release -r osx-x64 -o out /p:ShowLinkerSizeComparison=true`
+* Publish an application two different ways:
+  * `dotnet publish -c release -r linux-x64 -o linkedapp`
+  * `dotnet publish -c release -r linux-x64 -o notlinkedapp /p:LinkDuringPublish=false`
+
+Note: The `osx-x64` and `linux-x64` runtime IDs are used in the examples above. Feel free to use any runtime ID that you would like, such as `win-x64`.
```
swaroop-sridhar: This line is redundant, because we've already specified the set of values for `RID`

samples/linker-instructions.md
```diff
@@ -0,0 +1,64 @@
+# Using the .NET IL Linker
+
+The .NET team has built a linker to reduce the size of applications. In trivial cases, it can reduce the size of applications by 50%. The size wins may be more favorable or more moderate for larger applications.
+
+The linker currently only supports publishing self-contained applications. It will fail unless you specify a runtime ID. It is currently an experimental feature. We intend to productize it in a subsequent .NET Core release.
```
swaroop-sridhar: I think it is helpful to describe very briefly what the linker does -- remove unreachable definitions -- at this stage. This will help users understand the later usage such as "linker is too aggressive" later

samples/linker-instructions.md
```diff
@@ -0,0 +1,64 @@
...
+   * Linux/macOS: `./out/testapp`
+
+## Linker Switches
+
+The linker can be controlled with the following commandline switches.
+
+* `/p:LinkDuringPublish=false` -- Disable the linker.
+* `/p:ShowLinkerSizeComparison=true` -- Displays a table of size reductions for the application.
+
+You must disable the linker if you want to publish a [framework dependent application](https://docs.microsoft.com/en-us/dotnet/core/deploying/) while you have ILLink.Tasks as a dependency. This behavior will b changed in a later release.
```
swaroop-sridhar: Type: be

swaroop-sridhar: Maybe move this paragraph to the advanced document, since we haven't talked about the `ILLink.Tasks` package here.

samples/linker-instructions.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Determining the Benefit of the linker
+
+There are two straightforward approaches to determining the benefit of the linker.
+
+* Publish an application with the linker enabled and pass the `/p:ShowLinkerSizeComparison=true` flag. The output will describe the benefit. For example:
```
swaroop-sridhar: Suggestion: clarify "describe the benifit" to something like "show the code-size win for each file"

samples/linker-instructions-advanced.md
```diff
@@ -0,0 +1,63 @@
+# Using IL Linker Advanced Features
+
+This document described the more advanced features for the IL Linker and provides more insight into how it functions.
```
swaroop-sridhar: "More" should be dropped, unless we describe some functionality in the basic document.

samples/linker-instructions-advanced.md
```diff
@@ -0,0 +1,63 @@
...
+
+The dynamic features of .NET make it hard for this analysis to catch all cases in which code is called from the set of roots. In particular, reflection enables developers to do things like scan for assemblies at runtime and call their code. In general, it is impossible for the linker to determine exactly which code will get called at runtime, so there will be cases in which the linker removes code that should not be removed. We may add more sophisticated heuristics in the future to catch some common patterns of reflection usage, but the general problem remains, which necessitates a mechanism by which developers can explicitly tell the linker which parts of the code to consider as roots. Thus the linker operates on a set of assemblies and a set of specified code roots, producing an output set of assemblies with unused dependencies removed.
```
swaroop-sridhar: I think it would be good to call out the limitations of the linker in its own section.

samples/linker-instructions-advanced.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Using the tasks package
+
+The linker tasks package ([ILLink.Tasks](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Illink.Tasks)) contains MSBuild targets that become a part of the referencing project's build, using the mechanism described in the nuget docs.
```
swaroop-sridhar: Can we add a hyperlink to the nuget docs that describe the mechanism

samples/linker-instructions-advanced.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+This ItemGroup should contain the logical names of assemblies, not the filenames (so the assembly names should not have extensions).
+
+The linker roots can also be specified at a more granular level using xml root descriptor files, whose format is documented in the mono/linker repo. These files should be specified in the LinkerRootDescriptors ItemGroup:
+
+```xml
+<ItemGroup>
+  <LinkerRootDescriptors Include="path/to/rootDescriptor.xml" />
+</ItemGroup>
```
swaroop-sridhar: Should we give a sample XML exclusion file? or a link to mono-linker repo that contains one?

swaroop-sridhar:
Overall. the documentation looks good; a few minor comments are inline.

benaadams:
Very nice 
```
                            Before linking (B)   After linking (B)    Size decrease
-----------                 -----------          -----------          -----------
Total size of assemblies    48,494,336           15,093,496           68.88%
-----------                 -----------          -----------          -----------
```

Though it doesn't touch
```
System.Private.CoreLib.dll  11,860,184           11,860,184           0.00%
```
Or any `Microsoft.AspNetCore` libs, is this because the versions I'm referencing are precompiled native libs in the runtime cache; rather than pure il? If so maybe call this out?

samples/linker-instructions-advanced.md
MichaelSimons: "This document described" => "This document describes"

samples/linker-instructions-advanced.md
sbomer: Good idea. [This](https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#including-msbuild-props-and-targets-in-a-package) would be the link.

samples/linker-instructions-advanced.md
```diff
@@ -0,0 +1,63 @@
...
+```
```
russellhadley: For the above cases, would an example about that illustrated the "root whole assembly" vs the "add particular field/method/type".   This would help answer the "what do I do when something goes wrong" question.

samples/linker-instructions.md
richlander: You'd think. I worry that people will still get confused, hence why I partially repeated them. I think this is goodness w/ little harm.

samples/linker-instructions.md
richlander: That point doesn't seem very advanced. The package is covered in the instructions, per necessity.

samples/linker-instructions-advanced.md
richlander: I think it does.

richlander:
Applied feedback to the intro doc. I'll let others update the advanced doc. When folks are happy with both docs, we can merge.

richlander:
As an aside, I didn't write these docs, particularly the advanced one. I'm happy to change the authors if the original authors want that.

PR #883, opened 2017-08-22
Order releases as descending

karelz:
Order releases as descending to lightly nudge users to use the latest version.

### DIFF
--- a/release-notes/download-archive.md
+++ b/release-notes/download-archive.md
@@ -2,7 +2,35 @@
 
 This page provides an archive of previously released versions of the .NET Core runtime, libraries and the .NET Core SDK.
 
+## ["Current"](https://www.microsoft.com/net/core/support) Releases
+
+### .NET Core 2.0
+
+| Release Date | Description | Release Notes | |
+| :-- | :-- | :--: | :--: |
+| 2017/08/14 | 2.0.0 with SDK 2.0.0 | [release notes](2.0/2.0.0.md) | [download](download-archives/2.0.0-download.md) |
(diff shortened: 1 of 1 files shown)


richlander:
This also occurred to me. I was going back-and-forth on how to present LTS and Current. I don't have a plan yet that I'm fully happy with. That said, this looks fine to me as a first step.

@leecow?

leecow:
LGTM

bartonjs:
I find it a little odd that we have non-LTS releases (and non-supported releases) listed under LTS.

Yeah, "1.0" is LTS, but the preview builds weren't.  And "1.0.0" is out of support (if I understand our policy correctly) since we'd immediately say "jump to 1.0.(6?)" and see if we've already fixed your problem.  Likewise, 1.1.current was the first 1.1 build to be inherited into the LTS support cycle.

So I think I'd expect

* "Current" releases
  * 2.0.1 when it releases
  * 2.0.0 (also) until 30 days (or whatever the number is) after 2.0.1
* LTS releases
  * 1.1.next when it releases
  * 1.1.current (also) until 30 days (etc) after 1.1.next
  * 1.0.next when it releases
  * 1.0.current (also) until (etc)
* History
  * 2.0
    * release date | 2.0.1 | release train | (empty) | links
    * release date | 2.0.0 | Current | (empty) | links
    * release date | 2.0.0-Preview2 | Preview | whatever date support ended (if applicable for preview) | links
    * (etc)
  * 1.1
    * release date | 1.1.2 | Promoted to LTS | (empty) | links
    * release date | 1.1.1 | Current | whatever date support ended | links
    * (etc)
  * 1.0
    * 2017/05/09 | 1.0.5 | LTS | (empty) | links
    * (etc)

So it's clearer of "these are supported", and "this is the history" (and right next to the download link for a "history" build it says that it's out of support).  But that's easily a followup change.

richlander:
Good point @bartonjs. So, let's take this PR and then fix the issue.

PR #892, opened 2017-08-24
Use win10-arm on Raspberry Pi

bricelam:
Raspberry Pi runs Windows 10 IoT Core. When publishing, you should specify `win10-arm` so that the correct runtime assets get published. For example, a package could contain assets for both `win8-arm` and `win10-arm`, but the `win10-arm` ones should be used.


bricelam:
Closing: pull request timeout.

PR #906, opened 2017-08-28
Document using .NET Core on RHEL 6 / CentOS 6

janvorli:
Adds document that describes how to build .NET Core dependencies on RHEL
6 / CentOS 6 and how to install them in order to run .NET Core apps.


GavinScottQss:
@janvorli,
I'm confused because as far as I can tell, current .Net Core apps built for a linux-64 target have dependencies on C/C++ runtime versions which can't be built on a 2.6 kernel and thus aren't available on RHEL 6, and this document does not mention that issue. Is this something new since you documented this, or am I missing something. Thanks!

PR #911, opened 2017-08-30
Update a instruction to copy whole published directory

hqueue:
Update a instruction for copying app to Raspberry Pi, because sometimes newcomers were confused and only copy `helloworld`.


Petermarcu:
Thanks!

PR #1101, opened 2017-11-18
Add Platforms doc

richlander:
This file is intended to provide a location to link to all the key install locations, for both official and unofficial builds.

/cc @leecow @janvorli @Petermarcu


Petermarcu:
LGTM

PR #1135, opened 2017-12-07
Include release links in top-level README

richlander:
Goal: Make it easier to discover the latest .NET Core releases.

### DIFF
--- a/release-notes/README.md
+++ b/release-notes/README.md
@@ -1,9 +1,21 @@
-# .NET Core Releases
+# .NET Core Release Notes
 
-This area contains release information and download links for all .NET Core releases.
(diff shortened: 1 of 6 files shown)


release-notes/README.md
```diff
@@ -1,9 +1,21 @@
-# .NET Core Releases
+# .NET Core Release Notes
 
-This area contains release information and download links for all .NET Core releases.
+The latest major release is [.NET Core 2.0](2.0/2.0.0.md).
```
leecow: Should this point to the latest patch release?

richlander: This is a good point. I thought about that. My rationale was the following:

- Take the reader to the best explanation of the release.
- Maintain a latest servicing version at the top of the release (check and you'll see that).

That felt like the sweet spot experience. Thougths?

PR #1155, opened 2017-12-12
Added info for Ubuntu 17.10

yooakim:


danmosemsft:
any reason for wget instead of curl as the other ones?

richlander:
PTAL @leecow 

We should keep the command patterns the same unless there is a reason to do something different.

yooakim:
@danmosemsft the reason is that later Ubuntu does not have CURL installed. So if we stick with CURL a note about installing CURL would be appropriate.

leecow:
I've updated all downloads to use wget instead of curl to clear up some other inconsistencies.

PR #1250, opened 2018-02-02
Update portable pdb guidance in 4.7.1

noahfalk:
Unfortunately due to https://github.com/Microsoft/dotnet/issues/529 we had to withdraw Portable PDB from 4.7.1 but we anticipate having the feature restored in 4.7.2 with much improved performance.


noahfalk:
@tmat - look good?

tmat:
We should also include info how to enable Portable PDBs in stack traces on 4.7.2 if the app wasn't built against 4.7.2.

Something like: 
https://github.com/dotnet/designs-microsoft/blob/symbols2/proposed/publishing-symbols.md#stack-traces

PR #1253, opened 2018-02-02
add runtime-only install steps

leecow:


Petermarcu:
Good start. Do you want to open an issue on the dot.net website to make sure we get runtime instructions there as well?

PR #1332, opened 2018-03-12
fix version name typo

thstarshine:


leecow:
Looks like this was already fixed.

PR #1369, opened 2018-03-21
Update RaspberryPiInstructions.md

CIPop:
The prerequisites are not clear. The link takes you to prerequisites for SDK install which is not possible on arm. If you're not familiar with Linux It's also unclear that the current Raspbian is a Debian 8 Jessie.


CIPop:
Thanks @wfurt !

PR #1380, opened 2018-03-23
os-lifecycle-policy.md: Fixed typo

perlun:
Also added a year annotation for the Fedora note. (In general, when speaking about "later this year" etc, it's always best to be explicit with which year you are talking about, since documentation otherwise grows old faster than necessary.)


Petermarcu:
@leecow

PR #1394, opened 2018-03-30
Added 'async Main' helloworld example.

slawomirbrys:
Added new ```helloworldasync``` example project.

### DIFF
--- a/samples/helloworld/helloworld.csproj
+++ b/samples/helloworld/helloworld.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
-    <OutputType>Exe</OutputType>
-  </PropertyGroup>
+	<PropertyGroup>
+		<TargetFramework>netcoreapp2.0</TargetFramework>
+		<OutputType>Exe</OutputType>
+	</PropertyGroup>
 </Project>
--- a/samples/helloworld/helloworld.csproj
+++ b/samples/helloworld/helloworld.csproj
@@ -3,10 +3,4 @@
     <TargetFramework>netcoreapp2.0</TargetFramework>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <LangVersion>7.2</LangVersion>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <LangVersion>7.2</LangVersion>
-  </PropertyGroup>
 </Project>
--- a/samples/helloworldasync/helloworld.csproj
+++ b/samples/helloworldasync/helloworld.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
-    <OutputType>Exe</OutputType>
-	<LangVersion>7.1</LangVersion>
-  </PropertyGroup>
+	<PropertyGroup>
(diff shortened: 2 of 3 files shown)


samples/helloworld/helloworld.csproj
```diff
@@ -1,6 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp1.0</TargetFramework>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
Petermarcu: Why the need for these conditionals and the 7.2 specification?

samples/helloworldasync/helloworld.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
Petermarcu: Same comment, do you need these conditionals?

Petermarcu: Also async main was added in C# 7.1 I believe. Do you even need this Property?

slawomirbrys: Conditions taken from other github issue when faced the problem with compiler, but just checked without them and works well. Will change to ```latest``` instead of 7.2.
And yes, ```async``` was available since 7.1.

samples/helloworld/helloworld.csproj
slawomirbrys: Will remove.

samples/helloworldasync/helloworld.csproj
```diff
@@ -0,0 +1,7 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+	<LangVersion>latest</LangVersion>
```
svick: I think I would prefer explicitly specifying that this requires version 7.1, to make it clear which version introduced this option. (And in theory, if someone tried to compile this code with a 7.0 compiler, using `latest` would fail here.)

Also nit: tabs vs. spaces.

slawomirbrys: Done.

samples/helloworldasync/helloworld.csproj
```diff
@@ -0,0 +1,7 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+	<LangVersion>7.1</LangVersion>
```
Petermarcu: The 2.0 SDK support LangVersion 7.1 by default. I don't think you need this line at all.

svick: It is necessary, because the default `LangVersion` is `default`, which is the latest major version.

Petermarcu: @jaredpar

Petermarcu: My goal is for samples to be as simple as possible. (correct as well of course).

Petermarcu: Ok, I see now what the default ends up being and that this is needed. Can you fix the indentation (LangVersion should not be nested lower than OutputType). Then I can merge this. Thanks!

slawomirbrys: Done.

Petermarcu:
Thanks

PR #1421, opened 2018-04-11
Fixed broken links

weitzhandler:


leecow:
Looks like these were fixed up already.

PR #1454, opened 2018-04-17
Add C# language version

jaredpar:
Include the C# language version in the release.json file.

### DIFF
--- a/release-notes/releases.json
+++ b/release-notes/releases.json
@@ -37,7 +37,8 @@
         "checksums-sdk": "https://dotnetcli.blob.core.windows.net/dotnet/checksums/2.1.300-preview2-008533-sdk-sha.txt",
         "security": "FALSE",
         "lts-runtime": "FALSE",
-        "lts-sdk": "FALSE"
+        "lts-sdk": "FALSE",
+        "csharp-language": "7.3"
     },
     {
         "version-sdk": "2.1.104",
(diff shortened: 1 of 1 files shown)


jaredpar:
CC @jcouv another place where we track the C# version.

release-notes/releases.json
```diff
@@ -37,7 +37,8 @@
         "checksums-sdk": "https://dotnetcli.blob.core.windows.net/dotnet/checksums/2.1.300-preview2-008533-sdk-sha.txt",
         "security": "FALSE",
         "lts-runtime": "FALSE",
-        "lts-sdk": "FALSE"
+        "lts-sdk": "FALSE",
+        "csharp-language": "7.3"
```
jcouv: Thanks for the heads-up.
Is `csharp-language` a well-known string? Should it be `csharplang`?
Should we also record VB language version?

I assume the .NET Core folks would be responsible for updating this when they pick up new versions of the compiler packages. Is that right?

jaredpar: This is still hand authored and i just found out about it today. Im' trying to back fill all of the versions that I know of. 

The pattern seemed to be using `-` to separate words in this file hence I followed that for `csharp-language`.

jcouv: Who owns updating this?

jaredpar: Working that out.

jcouv: @jaredpar Did we resolve the question of ownership? Thanks

jcouv:
@jaredpar Did we sort out ownership for this moving forward?
When it comes to 8.0, are we going to update this file in 8.0 beta timeframe, or wait until 8.0?

PR #1609, opened 2018-05-29
Api diff between 2.0 and 2.1

weshaggard:
@terrajobst @leecow @preetikr @richlander 

Here is the API diff between 2.0 and 2.1, let me know if you would like it in a different location.


leecow:
Should probably live in release-notes/2.1 rather than Preview.

weshaggard:
So you suggest I put it here `release-notes/2.1/api-diff/2.0-vs-2.1.md`

leecow:
Hadn't thought of the api-diff directory but I like it.

weshaggard:
Moved.

terrajobst:
Looks good!

PR #1611, opened 2018-05-30
Add installers for 1.0 & 1.1 releases

rowanmiller:
Adding installers needed to drive the .NET Website. Only filling in the minimum set, not an exhaustive add of all previous versions.


leecow:
lgtm

PR #1616, opened 2018-05-30
Fix incorrect command name in 2.1 release notes

martincostello:
Change `buildserver` to `build-server`, as it was renamed in `preview2` I think.


leecow:
LGTM, thanks @martincostello

PR #1625, opened 2018-05-31
Acknowledge .NET Core 2.1 and Fedora 28 were released.

adelton:
Signed-off-by: Jan Pazdziora <jpazdziora@redhat.com>

### DIFF
--- a/os-lifecycle-policy.md
+++ b/os-lifecycle-policy.md
@@ -2,12 +2,13 @@
 
 .NET Core is [supported by Microsoft](microsoft-support.md) on a range of operatings systems and versions. Each platform has distinct Lifecycles defined by the parent organization. .NET Core support will take these Lifecycle schedules into account when adding or removing versions from the supported list.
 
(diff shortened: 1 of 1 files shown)


smudge202:
Does the [releases.csv](https://github.com/dotnet/core/blob/master/release-notes/releases.csv) need updating also? (I'm seeing this referenced to by Azure VSTS .Net Core Tool Installer Step)

adelton:
Possibly yes, but I'm not really familiar with everything that might need to be changed with the new release, or indeed, where to find those download URLs. I was only looking from the point of view of Fedora releases, EOL'ed and current.

smudge202:
Yeh, fair enough. Given that not everything seems to be working for 2.1.0 yet (see #1627 for example) it might be jumping the gun to merge this in just yet. 😞

leecow:
The example in the document is just that, an example. Adding the link to the 2.1 supported OS versions makes sense. Maintaining a generic example doesn't feel like something that's needed.

PR #1641, opened 2018-06-01
Update releases.csv

rolfsjod:
Added 2.1.0 release according to the JSON-release file


leecow:
Thanks for the submission. Looks like @preetikr updated the file already.

Issue #1644, opened 2018-06-01
API diffs not showing readonly ref as readonly


jnm2:
It looks like the API diff generator not aware of readonly ref, so it reads as though you're adding APIs that return a writable ref to ImmutableArray elements:

https://github.com/dotnet/core/blob/5185cd6ad31a8e6dd9cb9d4b5a2986cc28e3f925/release-notes/2.1/api-diff/2.0-vs-2.1_System.Collections.Immutable.md#systemcollectionsimmutable

Differentiating `readonly ref` from `ref` seems like it could be helpful to folks like me.
@karelz suggested filing the issue until @weshaggard is available.

ViktorHofer:
cc @terrajobst

leecow:
cc @weshaggard

weshaggard:
I used a UI version of our tools which don't understand this concept, we did add support with https://github.com/dotnet/buildtools/pull/1849. I will use a commandline version to update the diff. Thanks for pointing out @jnm2

PR #1664, opened 2018-06-06
Refresh 2.0 vs 2.1 diff with latest tools

weshaggard:
Address https://github.com/dotnet/core/issues/1644.


release-notes/2.1/api-diff/2.0-vs-2.1_System.Collections.Specialized.md
```diff
@@ -0,0 +1,11 @@
+# System.Collections.Specialized
+
+``` diff
+ namespace System.Collections.Specialized {
+     public struct BitVector32 {
+-        public struct Section
++        public readonly struct Section
```
weshaggard: @VSadov @stephentoub @jkotas can you guys refresh my memory is adding readonly to an existing struct a breaking change?

stephentoub: No, as long as it's truthful and it actually is read-only.

weshaggard: What defines it to be readonly? Do all the fields need to marked readonly?

VSadov: Yes, it is a nonbreaking if that is the only change.

VSadov: Adding `readonly` modifier will force that all instance fields/autoproperties/events are readonly.

If you have to change any of that, be careful!

VSadov: `readonly struct S1`    basically tells to the outer world "hey, I do not have any mutable state".

In a context of reference assemblies, - if a struct says it is readonly in a ref assembly, but an implementation adds a mutable instance field (even private one), it would be bad.

weshaggard: Thanks. I just wanted to be sure that if someone was using it in some mutable way in the past and we now added readonly to that the compiler wouldn't start complaining.

PR #1737, opened 2018-06-28
Fix download links for ASP.NET Core 2.1.1

natemcmaster:
Made some fixes to the download links.

* Update name in releases.json for the linux-arm (i.e. arm32v7) aspnetcore-runtime download
* Add link for ASP.NET Core runtime binaries for linux-x64
* Clarify the meaning of glibc/musl/ARM in the columns of the download table.


natemcmaster:
Also, how do I go about getting this to say arm32, not arm64? https://www.microsoft.com/net/download/linux
screenshot

cref https://github.com/dotnet/core-setup/issues/3122#issuecomment-400957801

natemcmaster:
Ping @leecow

release-notes/releases.json
```diff
@@ -93,7 +93,7 @@
         "rps-win-x86": "aspnetcore-runtime-2.1.0-win-x86.zip",
         "rps-mac-x64": "aspnetcore-runtime-2.1.0-osx-x64.tar.gz",
         "rps-linux-x64": "aspnetcore-runtime-2.1.0-linux-x64.tar.gz",
-        "asp-runtime-linux-arm-x64":"aspnetcore-runtime-2.1.0-linux-arm.tar.gz",
+        "asp-runtime-linux-arm":"aspnetcore-runtime-2.1.0-linux-arm.tar.gz",
```
leecow: This is the 64bit download, right? I see in the storage account that asp is producing

* aspnetcore-runtime-internal-2.1.1-linux-arm.tar.gz
* aspnetcore-runtime-internal-2.1.1-linux-musl-x64.tar.gz
* aspnetcore-runtime-internal-2.1.1-linux-x64.tar.gz

I presume the *linux-arm-tar.gz is 32bit. If this is true, wouldn't it be better to append the release asset with the correct architecture (x64) which will make the eventual addition of the 32 bit asset natural? Otherwise, when that time comes, the thing people have been accessing as "asp-runtime-linux-arm" will change from 64 to 32 bit.

leecow:
arm32 links are only on the download archive pages at the moment (https://github.com/dotnet/core/blob/master/release-notes/download-archives/2.1.1-download.md). If it's time to promote them to the marketing pages, we can make that addition.

release-notes/releases.json
natemcmaster: `aspnetcore-runtime-2.1.0-linux-arm.tar.gz` is a 32bit release. It's technically arm32v7. IIUC running arm32 from arm64 is optional in the Linux kernel, and in practice some arm64 processors do not support arm32.

nzain:
The key problem is the missing download link for arm32 aspnet+dotnet runtime. In
https://github.com/dotnet/core-setup/issues/3122#issuecomment-401087300
@natemcmaster clarified that the 64bit download link is actually 32bit and I verified this on real arm32 hardware. It would be nice, if there is an official dotnet+aspnet arm32 download link, so we can integrate this link in our buildroot workflow.

natemcmaster:
:bell:

nzain:
@natemcmaster Sorry for asking stupid questions! The PR is merged and the website looks different now... nevertheless I can't find the download link for
* dotnet runtime
* arm32
* including ASP.Net runtime store

Side note: it seems the ASP.Net runtime is hidden even deeper now.
https://www.microsoft.com/net/download/dotnet-core/2.1

natemcmaster:
@nzain it looks like the website refactoring still has an error in it. The "arm64" link on the downloads page is incorrectly labeled. The link it downloads is actually arm32. I'll send an email to our website team and hopefully they will fix it. In the meantime, this is the link you're looking for. It contains the 2.1.2 ASP.NET Core runtime nad .NET Core runtime for arm32.

screenshot

Note: in 2.1, ASP.NET is "shared runtime", not a runtime store anymore (subtle, but important distinction). So, you won't find links in 2.1 that are labeled "package runtime store."

PR #1742, opened 2018-06-29
Update microsoft-support.md

jingeno:
.NET Core 2.1 is a LTS release and not a Current release.


jingeno:
LTS period for .NET Core 2.1 has not begun yet.

PR #1814, opened 2018-07-25
update linux prereq documenation

leecow:


danmosemsft:
I do not know what the correct values are. If they are these, feel free to merge..

leecow:
@omajid I thought I ran into libunwind while testing my way through the matrix but checked again and looks like it can be pulled from the table for 2.1.

PR #1844, opened 2018-08-09
Removed reference to Red Hat errata policy for RHEL6 from release notes

RheaAyase:
I replaced it with Microsoft support policy instead, not sure if that's the right thing to refer to...


leecow:
Seems fine to me.

PR #1897, opened 2018-08-27
Sort download archive by versions

karelz:
Sort releases by order from newest to oldest major.minor version.


karelz:
@leecow are you ok with my change? Should I move ".NET Core Tools for Visual Studio 2015" to the bottom? (I don't think it needs Current vs. LTS attribution)

PR #1960, opened 2018-09-24
Updates to move to v2 of the releases.json format.

leecow:
- Each version directory will have it's own releases.json
- Top level (/release-notes) has releases-index.json


release-notes/2.0/releases.json
```diff
@@ -0,0 +1,639 @@
+{
+        "netcore-release": [
```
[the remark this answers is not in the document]
leecow: fixed

release-notes/2.0/releases.json
```diff
@@ -0,0 +1,639 @@
...
+                "sdk": {
+                    "version": "2.1.202",
+                    "version-display":"2.1.202",
+                    "vs-version":"15.7",
+                    "lts": false,
+                    "eol": false,
+                    "csharp-language": "7.2",
+                    "checksums": "https://dotnetcli.blob.core.windows.net/dotnet/checksums/2.1.202-sdk-sha.txt",                
+                    "files":[
+                        {   "name":"sdk-linux-x64",
```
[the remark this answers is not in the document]
leecow: Working on it and final formatting resolution will happen when generator tooling is in place.

release-notes/1.0/releases.json
```diff
@@ -0,0 +1,446 @@
...
+    {
+        "release-date": "2018-03-13",
+        "channel-version": "1.0",
+        "release-version": "1.0.10",
+        "security": true,
+        "release-notes": "https://github.com/dotnet/core/blob/master/release-notes/1.0/1.0.10.md",
+        "runtime":{
+            "version": "1.0.10",
+            "lts": true,
+            "checksums-": "https://dotnetcli.blob.core.windows.net/dotnet/checksums/1.0.10-runtime-sha.txt",
```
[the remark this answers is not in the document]
leecow: Hand crafted at the moment. The checksums file ref will be dropped from this json.

release-notes/2.0/releases.json
```diff
@@ -0,0 +1,639 @@
...
+                    "vs-version":"15.7",
+                    "lts": false,
+                    "eol": false,
+                    "csharp-language": "7.2",
+                    "checksums": "https://dotnetcli.blob.core.windows.net/dotnet/checksums/2.1.202-sdk-sha.txt",                
+                    "files":[
+                        {   "name":"sdk-linux-x64", 
+                            "url":"https://download.microsoft.com/download/f/c/1/fc16c864-b374-4668-83a2-f9f880928b2d/dotnet-sdk-2.1.202-linux-x64.tar.gz", 
+                            "hash": "c1b07ce8849619ca505aafd2983bcdd7141536ccae243d4249b0c9665daf107e03a696ad5f1d95560142cd841a0888bbf5f1a8ff77d3bdc3696b5873481f0998"},
+                        {   "name":"sdk-mac-x64",
```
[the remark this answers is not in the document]
leecow: I have been treating it as an unversioned identifier rather than a filename. Will need to think through changing it to mean 'filename'.

release-notes/2.0/releases.json
```diff
@@ -0,0 +1,639 @@
...
+                    ]
+                },
+                "sdk": {
+                    "version": "2.1.202",
+                    "version-display":"2.1.202",
+                    "vs-version":"15.7",
+                    "lts": false,
+                    "eol": false,
+                    "csharp-language": "7.2",
+                    "checksums": "https://dotnetcli.blob.core.windows.net/dotnet/checksums/2.1.202-sdk-sha.txt",
```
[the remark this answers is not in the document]
leecow: agreed, see other comment regarding the checksum entry.

release-notes/2.0/releases.json
```diff
@@ -0,0 +1,639 @@
...
+                "release-version": "2.0.9",
+                "security": true,
+                "release-notes": "https://github.com/dotnet/core/blob/master/release-notes/2.0/2.0.9.md",
+                "runtime": {
+                    "version": "2.0.9",
+                    "version-display":"2.0.9",
+                    "vs-version":"15.7",
+                    "lts": false,
+                    "eol": false,
+                    "eol-date": "",
```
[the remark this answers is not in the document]
leecow: Massaged your suggestion just a bit. How about this?

{
    "channel-version": "2.1",
    "latest": "2.1.5",
    "release-date":"2018-10-12",
    "lts": true,
    "eol": false,
    "eol-date": null,
    "lifecycle-policy": "https://www.microsoft.com/net/support/policy",    
    "releases": 
    [
        {
            "release-date":  "2018-10-02",
            "release-version":  "2.1.5",
            "security":  false,
            "release-notes":  "https://github.com/dotnet/core/blob/master/release-notes/2.1/2.1.5/2.1.5.md",
            "runtime":  
            {
                "version":  "2.1.5",
                "version-display":  "2.1.5",
                "vs-version":  "15.8.6",
                "files":  
                [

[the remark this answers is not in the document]
MichaelSimons: Looks good.

release-notes/2.0/releases.json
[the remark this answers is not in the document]
MichaelSimons: The part that is confusing to me is that it is being treated as an intelligent key/identifier yet the value is not unique enough to distinguish the difference between the various files.  By that I mean `"name": "sdk-win-x64"` doesn't contain enough information to distinguish between the zip and exe.  This could lead to issues with consuming code.  IMO filename would be a better choice here.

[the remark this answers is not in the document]
leecow: Updated `name` values to be what we chatted about in IM. eg filename=dotnet-runtime-2.1.5-linux-arm.tar.gz would yield name=dotnet-runtime-linux-arm.tar.gz

release-notes/1.0/releases.json
```diff
@@ -310,47 +310,47 @@
...
                 },
                 {
-                    "name": "sdk-mac-x64.pkg", 
+                    "name": "dotnet-dev-osx-x64.pkg", 
                     "url": "https://download.microsoft.com/download/6/5/F/65F1653E-F835-4DE3-BB36-F324D3925F32/dotnet-dev-osx-x64.1.1.8.pkg", 
                     "hash": "f1d438376839cb9343b2a893608a482d91134e4fa4aa1bc015b50ca58e9aea419043da134102496f13c9a90302abeaf9cb7b07df4e2dfaf528b3614f84847ea2"
                 },
                 {
-                    "name": "sdk-win-x86", 
+                    "name": "BB36-F324D3925F32/dotnet-dev-wizip",
```
MichaelSimons: This seems incorrect.  This entry is repeated in more than one place.

release-notes/1.0/releases.json
```diff
@@ -0,0 +1,1261 @@
...
+                    "url": "https://download.microsoft.com/download/4/0/2/4022CFC7-5061-4762-B9BA-48B35632572D/dotnet-dev-osx-x64.1.1.9.tar.gz", 
+                    "hash": "c6e74231bdfd9d02dee203e37eafa4bd3a99645fdd982aa319800136b3ee9f4d90769095d05337f80ef50722e16d6c66eac87d83154055bd085d84bed31d50aa"
+                },
+                {
+                    "name": "dotnet-dev-osx-x64.pkg", 
+                    "url": "https://download.microsoft.com/download/4/0/2/4022CFC7-5061-4762-B9BA-48B35632572D/dotnet-dev-osx-x64.1.1.9.pkg", 
+                    "hash": "002eaebff614b5ce20472e1247cf980c6659ff03913cf1d7fadf675ada651e1fc35caa55fe8d2b0ede6b250adac85bc41378c331088d853f4decd009bf7ee2b0"
+                },
+                {
+                    "name": "B9BA-48B35632572D/dotnet-dev-wizip",
```
MichaelSimons: This doesn't seem correct.  This appears in more than one spot.

release-notes/releases-index.json
```diff
@@ -0,0 +1,44 @@
...
+            "eol-date": null,
+            "releases.json": "https://github.com/dotnet/core/blob/master/release-notes/2.1/releases.json",
+            "lifecycle-policy": "https://www.microsoft.com/net/support/policy"
+        },
+        {
+            "channel-version": "2.0",
+            "latest": "2.0.9",
+            "release-date":"2018-05-21",
+            "lts": false,
+            "eol": false,
```
MichaelSimons: This should be set to true now and get propagated to the 2.0 file.

leecow: Fixed. Also updated releases-index.json.

release-notes/1.0/releases.json
```diff
@@ -0,0 +1,1261 @@
...
+                    "url": "https://download.microsoft.com/download/9/e/6/9e6e1700-f682-4e4c-9e02-583f102cb048/dotnet-dev-osx-x64.1.1.10.tar.gz", 
+                    "hash": "bf3bf5f22c70c944a9459565626ef935cd15465a189ef9dd65d18560f7e71e44e2ac93552999774eb9a6672530a99baa30347bf3ea1cc7f4ac4d583e1f0e60b9"
+                },
+                {
+                    "name": "dotnet-dev-osx-x64.pkg", 
+                    "url": "https://download.microsoft.com/download/9/e/6/9e6e1700-f682-4e4c-9e02-583f102cb048/dotnet-dev-osx-x64.1.1.10.pkg", 
+                    "hash": "990d54a53702f5314ad06aab3fa7b6ad0944f2d8863d5ba99f9e86a660fc542378e0fa2cf9aebbd44bdee71bdd37d287409ded40f98d4a43f77e613a3d84fee3"
+                },
+                {
+                    "name": "9e02-9e02-583f102cb048/dotnet-dev-w.zip",
```
MichaelSimons: There are move of these incorrect names.

MichaelSimons: search for `dotnet-dev-wizip` and you will find a number of them.  I'm not sure if there are more as well.

leecow: Sigh, the hazards of hand editing / pattern matching. Fixed up these instances then hit on a good way to view all "name" fields in a contiguous list which made reviewing everything much easier. Found a handful of other patterns that were incorrect and a few other errors in the process. Updated the 2.0 file as penance for being sloppy.

release-notes/releases-index.json
```diff
@@ -0,0 +1,44 @@
+{
+    "core-releases-index": [
+        {
+            "channel-version": "2.1",
+            "latest": "2.1.4",
+            "release-date":"2018-09-11",
+            "lts": true,
+            "eol": false,
+            "eol-date": null,
```
richlander: I would suggest and empty string rather than null for eol-date. This would make processing easier.

release-notes/releases-index.json
```diff
@@ -0,0 +1,44 @@
+{
+    "core-releases-index": [
```
richlander: You might name this property just "releases-index" or even just "releases". I'd add another property for product called "product" with the value ".NET Core". This would enable us or others to use this same schema for other products.

release-notes/releases-index.json
```diff
@@ -0,0 +1,44 @@
...
+        {
+            "channel-version": "2.1",
+            "latest": "2.1.4",
+            "release-date":"2018-09-11",
+            "lts": true,
+            "eol": false,
+            "eol-date": null,
+            "releases.json": "https://github.com/dotnet/core/blob/master/release-notes/2.1/releases.json",
+            "lifecycle-policy": "https://www.microsoft.com/net/support/policy"
```
richlander: I'd make "lifecycle-policy" a top-level property and include it per releases within the target releases.json file.

release-notes/releases-index.json
```diff
@@ -0,0 +1,44 @@
+{
+    "core-releases-index": [
+        {
+            "channel-version": "2.1",
+            "latest": "2.1.4",
+            "release-date":"2018-09-11",
+            "lts": true,
+            "eol": false,
```
richlander: Instead of "eol" as a property, you might go with: "support-phase": "full-support" | "maintenance" | "eol"

release-notes/releases-index.json
leecow: It's already in the target releases.json files. The question is, is duplicating this top level information in the index useful?

release-notes/releases-index.json
leecow: Change will be in the next commit.

release-notes/releases-index.json
leecow: Was thinking the same thing after our 'maintenance' exchange. Change will be in the next commit.

release-notes/releases-index.json
leecow: going with 'releases-index'

release-notes/releases-index.json
richlander: Primary point is not duplicating it in every array entry in this file. We can likely get away w/o it being in this file at all. There are many links that we could include here. Only one of them is required.

release-notes/releases-index.json
```diff
@@ -0,0 +1,44 @@
...
+            "eol-date": null,
+            "releases.json": "https://github.com/dotnet/core/blob/master/release-notes/2.1/releases.json",
+            "lifecycle-policy": "https://www.microsoft.com/net/support/policy"
+        },
+        {
+            "channel-version": "2.0",
+            "latest": "2.0.9",
+            "release-date":"2018-05-21",
```
richlander: I get these three property names. I'm wondering if we can do better. How about (in order):

- version
- latest-release
- latest-release-date

leecow: latest-release and latest-date make sense. channel-version has a specific meaning in both files that I would like to retain.

PR #1987, opened 2018-10-08
Make wget slightly more verbose

guntbert:
`wget -q` is completeley quiet and doesn't even report any errors.
This can lead to a situation where the subsequent command fails but
the user is not made aware of that issue (like with a 404 error,...).

Therefore I propose `wget -nv` which prevents `wget` from being verbose
but reports errors.

### DIFF
--- a/Documentation/linux-setup.md
+++ b/Documentation/linux-setup.md
@@ -21,34 +21,34 @@ Open a command prompt and run the commands below that match your distro:
 
 ```bash
 # Debian 8 - .NET Core 2.0 and newer
-wget -qO- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.asc.gpg
+wget -nv -O- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.asc.gpg
 sudo mv microsoft.asc.gpg /etc/apt/trusted.gpg.d/
-wget -q https://packages.microsoft.com/config/debian/8/prod.list
+wget -nv https://packages.microsoft.com/config/debian/8/prod.list
 sudo mv prod.list /etc/apt/sources.list.d/microsoft-prod.list
 sudo chown root:root /etc/apt/trusted.gpg.d/microsoft.asc.gpg
 sudo chown root:root /etc/apt/sources.list.d/microsoft-prod.list
 
 # Debian 9 - .NET Core 2.0 and newer
-wget -qO- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.asc.gpg
+wget -nv -O- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.asc.gpg
 sudo mv microsoft.asc.gpg /etc/apt/trusted.gpg.d/
-wget -q https://packages.microsoft.com/config/debian/9/prod.list
+wget -nv https://packages.microsoft.com/config/debian/9/prod.list
 sudo mv prod.list /etc/apt/sources.list.d/microsoft-prod.list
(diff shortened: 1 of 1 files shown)


Documentation/linux-setup.md
```diff
@@ -21,34 +21,34 @@ Open a command prompt and run the commands below that match your distro:
 
 ```bash
 # Debian 8 - .NET Core 2.0 and newer
-wget -qO- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.asc.gpg
+wget -nvO- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.asc.gpg
```
wfurt: This gives me error:
```
furt@Ubuntu:~$ wget -nvO- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor
wget: illegal option -- `-nO'
```
It should probably be 
`wget -nv -O -  https://packages.microsoft.com/keys/microsoft.asc`

guntbert: Ouch, I was too enthusiastic and did not check every command :blush: . Thx for catching it. 
Separating the switches (like you suggested) works now.

karelz:
cc @tmds @omajid

wfurt:
note that the output looks like (old and new):

```
furt@Ubuntu:~$ wget -q https://packages.microsoft.com/config/debian/8/prod.list
furt@Ubuntu:~$
furt@Ubuntu:~$ wget -nv https://packages.microsoft.com/config/debian/8/prod.list
2018-10-08 13:58:16 URL:https://packages.microsoft.com/config/debian/8/prod.list [75/75] -> "prod.list" [1]
 ```

the `-q` provides no feedback.

karelz:
What about the `-O-` arg?
@guntbert do you have example of what is the difference in `-q` and `-nv` when there is error?

guntbert:
@karelz 
1. The `-O-` redirects the received content to `stdout`
2. 
- `$ wget -q https://packages.microsoft.com/config/ubunut/18.04/packages-microsoft-prod.deb` (mind the typo *ubunut*) produces no output at all, thus giving the impresssion that all is well.
- `$ wget -nv https://packages.microsoft.com/config/ubunut/18.04/packages-microsoft-prod.deb` produces
`https://packages.microsoft.com/config/ubunut/18.04/packages-microsoft-prod.deb:
2018-10-09 12:35:24 ERROR 404: Not Found.` and thus gives a clear indication of an error.

karelz:
Thanks @guntbert!
I think the `-O-` spits out rarely used information and adds unnecessary clutter to the output. I would suggest to remove it, unless there is proof this is common way install steps behave on Linux.

guntbert:
@karelz That would break the whole thing. What happens here is the following:

- `wget ... -O-` sends the content of the requested file to `stdout` so it can be processed further in the pipeline (This is the public key in an "amored" format, able to be sent through text-only protocols)
- `| gpg --dearmor` converts this key to the standard format needed on the system
- `> microsoft.asc.gpg` sends it to a file in the current directory
- the next command moves this file into the directory of trusted gpg-keys

It is true that with `-nv` (or `--no-verbose`) on the "happy path" we get some unneeded clutter but that does not hurt the process. With `-q` there is no clutter but also no information when things go wrong.

If that clutter is too much we should leave the whole thing alone but removing `-O-` (or `-O -` which does the same) would break it.

guntbert:
Thank you - I was surprised how fast paced your review process is (given the size of the project) :+1:

karelz:
core repo is easier to manage if you look at number of issues and PRs. CoreFX, CoreCLR and others are more challenging ;)

Issue #1986, opened 2018-10-05
.NET Core Release Data (releases.json) Redesign and Deprecation 


leecow:
# .NET Core Release Data (releases.json) Redesign and Deprecation

The releases.json and releases.csv files began life as helper tools for some of our internal release work, some experimentation and to possibly be useful to the community. It has been great to learn the files are indeed in use and many folks depend on them as a trigger for a number of activities.

* [release-notes/releases.json](https://github.com/dotnet/core/blob/master/release-notes/releases.json)
* [release-notes/releases.csv](https://github.com/dotnet/core/blob/master/release-notes/releases.csv)

Between the usages which we have become aware and in considering further activities that could be enabled to reason about things such as checking .NET Core version availability and support status, it became clear the current format is inadequate.

An updated format has been designed and tested which will replace the current files. The timeline below describes the path to obsoleting the current files.

## The Plan

| Date | Action |
| :--- | :----- |
| Oct 2018 | Notify the community of the coming change |
| Oct 2018 | Release the new json format into dotnet/core |
| Oct 2018 - Feb 2019 | Maintain the old json and csv files |
| Feb 2019 | Reminder announcement that old format file is coming soon |
| Mar 2019 | Old json and csv files are obsoleted and removed from dotnet/core |

## Breaking change

If you are currently utilizing the old version json or csv, you will need to update the relevant code to work with the new format. If you are just getting started with consuming this data, use the new version.

## The Format

Below are fragments of the files from our most recent .NET Core release.

### release-index.json

This file consists of a top level array with data on the latest release versions and a property for locating the more detailed channel-specific releases.json. This file is found in [dotnet/core/release-notes](https://github.com/dotnet/core/tree/new-json/release-notes/)

[releases-index.json](https://github.com/dotnet/core/blob/new-json/release-notes/releases-index.json)

```
{
    "releases-index": [
        {
            "channel-version": "2.1",
            "latest-release": "2.1.5",
            "latest-release-date":"2018-10-02",
            "product": ".NET Core",
            "support-phase": "lts",
            "eol-date": "",
            "releases.json": "https://github.com/dotnet/core/blob/master/release-notes/2.1/releases.json",
        },
        ...
}
```

### releases.json

[...]

PR #2006, opened 2018-10-18
WIP: Fixes in the new releases.json format

arthurrump:
Fixes for the issues I raised in #1986.

Note that for 2.2.0-preview1 and 2.2.0-preview2 I could not find some of the symbol files that are listed for preview3, so I removed those entries.

TODO:
- [ ] Add SDK release 2.1.400

### DIFF
--- a/release-notes/1.0/releases.json
+++ b/release-notes/1.0/releases.json
@@ -1227,6 +1227,7 @@
     },
     {
         "release-date": "2016-09-13",
+        "release-version": "1.0.1",
         "security": false,
         "release-notes": "https://github.com/dotnet/core/blob/master/release-notes/1.0/1.0.1-release-notes.md",
         "runtime":
--- a/release-notes/1.1/releases.json
+++ b/release-notes/1.1/releases.json
@@ -603,7 +603,7 @@
             },
             "sdk":
             {
-                "version-sdk": "1.1.8",
+                "version": "1.1.8",
                 "vs-version": "",
                 "files":
                 [
@@ -962,7 +962,7 @@
             },
             "sdk":
             {
-                "version-sdk": "1.1.4",
+                "version": "1.1.4",
                 "vs-version": "",
                 "files":
                 [
--- a/release-notes/1.1/releases.json
+++ b/release-notes/1.1/releases.json
@@ -673,7 +673,7 @@
             }
         },
         {
-            "release-date": "2017-11-14",
+            "release-date": "2018-01-09",
             "release-version": "1.1.6",
             "security": true,
             "release-notes": "https://github.com/dotnet/core/blob/master/release-notes/1.1/1.1.6.md",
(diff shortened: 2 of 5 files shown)


arthurrump:
Question: are the `version-display` fields supposed to strip off the junk after the second dash in preview versions? That seems to be how it works in the old releases.json, but the new one isn't consistent with that. The latest 2.2.0-preview3 release has for the runtime `"version-display": "2.2.0-preview3-27014-02"` in 2.2/releases.json, but in the old releases.json it has `"version-runtime-display": "2.2.0-preview3"`. The latter looks more sensible to me.

arthurrump:
I only modified the `version-display` in 2.2/releases.json for now, which makes it consistent with the old releases.json, which used `version-*-display` from version 2.1 onwards.

I would be in favor of changing this for 2.0 too (both in the new and old releases.json), as it looks to be consistent with the messaging around those releases, both talking about just [Preview 1](https://blogs.msdn.microsoft.com/dotnet/2017/05/10/announcing-net-core-2-0-preview-1/) and [Preview 2](https://blogs.msdn.microsoft.com/dotnet/2017/06/28/announcing-net-core-2-0-preview-2/), but I'm not sure if it is something that should be changed.

As for 1.1 and 1.0, the releases that are listed for those channels only have previews for the SDK. For 1.1, that version number uses a different scheme than all others, so I don't think it makes sense to apply the "rule" there. For 1.0, applying the "rule" would result in four releases with the same SDK version number displayed, even though they are different builds, so I think those should be left alone too.

arthurrump:
The download links should be good now. (At least if I didn't look over some again...)

arthurrump:
@leecow Please let me know if there are other issues blocking this

arthurrump:
ping @leecow

leecow:
Sorry for the delay. Will have this looked over by tomorrow.

release-notes/3.0/releases.json
```diff
@@ -199,7 +199,9 @@
             "aspnetcore-runtime": {
                 "version": "3.0.0-preview-18579-0056",
                 "version-display": "3.0.0-preview-18579-0056",
-                "version-aspnetcoremodule": "13.0.18333.0",
+                "version-aspnetcoremodule": [
+                    "13.0.18333.0"
```
leecow: I don't know that we would ever have a list of ancm versions for a given release. @shirhatti ?

arthurrump: In all other new releases.json files the version-aspnetcoremodule field is specified as a list, and in some occasions multiple versions are indeed specified, e.g. [for release 2.1.6](https://github.com/dotnet/core/blob/master/release-notes/2.1/releases.json#L268-L272).

But while looking this up I also noticed that in the old releases.json this is not the case, so it might just be a mistake.

arthurrump: I think this has been resolved, judging from f84493d626f21b42cebdaf3ea18e841f32ac550a

arthurrump:
While we're at it again, would it make sense to add the new `latest-runtime` and `latest-sdk` fields to releases-index.json too? That would allow you to find the latest runtime and sdk versions with just one round trip.

PR #2014, opened 2018-10-21
Fix space ammount

meriturva:
Current value proposed for extra space is 10GB. Please correct settings to 1GB as commented above.


wfurt:
I don't think there is "correct" value. That really depends on you board. For example if you have 2-6G flash 10G will not fit. However 1G may not be enough in many cases. This was mostly example how to set it. 
This change makes description and value consistent.

karelz:
Thanks @meriturva for submitting your change!

PR #2028, opened 2018-10-29
added year to the license file

azkarmoulana:
added year as 2017-present in license file


richlander:
We were guided not to add a year. That's our plan for the file unless we find out it prevents someone from using the product.

PR #2076, opened 2018-11-27
Update ANCM versions

shirhatti:
Propagate changes from https://github.com/dotnet/dotnet-core-website/pull/830/


vivmishra:
We are carrying these edits for the big push on 12/4 so we can probably skip this PR

PR #2091, opened 2018-11-29
Reference latest package version from NuGet linker

nightblade9:
This way, you get the latest and greatest version, not a 2017 version.


nightblade9:
@sbomer sorry for the delay. Is that better? I never noticed, but the line right above that mentions the same link that you posted, from which you can get the latest version. So, I added the latest version back in, with a hint ("as of writing") that might trigger future readers to look up the latest version.

Is this sufficient, or would you like something else?

sbomer:
Perfect, thank you!

karelz:
@sbomer is it ready for merge?

sbomer:
Yup!

PR #2154, opened 2018-12-15
Add Code of Conduct

karelz:


richlander:
Can we use the simpler language that we used on the [dotnet/wpf](https://github.com/dotnet/wpf#code-of-conduct) repo? When we first added the CoC, it was important to explain it heritage. It isn't anymore. We just need to say that it is the .NET Foundation CoC.

Is the `CODE_OF_CONDUCT.md` file necessary? Does GitHub respect it?

karelz:
Yes, GitHub respects CODE_OF_CONDUCT.md file: https://help.github.com/articles/adding-a-code-of-conduct-to-your-project/. We can move it to `.github` subdir to "hide it" from root dir.
However, GitHub does not (yet) recognize custom text in the file, so technically it does not help with community checklist: https://github.com/dotnet/wpf/community.

@richlander which wording do you want to use? We have 2 in WPF:
* https://github.com/dotnet/wpf/blob/master/CODE_OF_CONDUCT.md
* https://github.com/dotnet/wpf#code-of-conduct

richlander:
> which wording do you want to use? 

The second one. That's the one I linked to initially.

Is there a link that shows what that file does? I cannot see what that is.

I'm wondering if we should go with the text I prefer in the README and then copy the whole contributor covenant text  into the C_O_C.md file? @jongalloway ... do you have thoughts on this?

jongalloway:
From looking at other projects, the standard seems to be linking to code of conduct rather than copying the text in ([k8s](https://github.com/kubernetes/kubernetes/blob/master/code-of-conduct.md), [rails](https://github.com/rails/rails/blob/master/CODE_OF_CONDUCT.md)). So LGTM as-is.

karelz:
> That's the one I linked to initially.

Ah, I didn't notice it is specific link - the title hides the subsection link :(

> Is there a link that shows what that file does? I cannot see what that is.

The linked docs is all I have: https://help.github.com/articles/adding-a-code-of-conduct-to-your-project/
It is integrated in https://github.com/dotnet/wpf/community checklist.

jongalloway:
Okay, after looking at this more, the [Contributor Covenant](https://www.contributor-covenant.org/#using-the-contributor-covenant) guidance recommends adding the text or markdown version at the root of the project. Also, at a minimum we also need to add a contact method.

terrajobst:
I'd agree with @jongalloway. It seems more logical to copy & paste the text into that file. We do the same for licenses and CoC seems to be like that -- the point is, it rarely -- if ever changes.

karelz:
Just to clarify my understanding: The prefered way is to use "default" GH way and include full text of the CoC in the CODE_OF_CONDUCT.md file in root, is that correct?
If that's the case, I will merge this PR and create a new one via GH "Add CoC".
Please upvote if you agree. Thanks!

richlander:
@karelz -- yes. However, we need to validate that the default text and the .NET Foundation CoC text match. Otherwise, it would be odd. That's a question for @jongalloway.

karelz:
This issue has stalled. I will merge the current form of CoC.
If there are changes needed, please submit a separate PR.

PR #2315, opened 2019-02-14
Fix maintenance releases header in microsoft-support.md

EarthlingRich:
Fix markdown error for maintenance releases header in microsoft-support.md.


leecow:
Thanks @EarthlingRich - lgtm.

PR #2326, opened 2019-02-18
Fixes for releases.json files (new style)

arthurrump:
Most of these would have been prevented if the [NetCore.Versions.Checks](https://github.com/apps/netcore-versions-checks) app [(code on GitHub)](https://github.com/arthurrump/NetCore.Versions) had been installed; I'm looking into making it catch the other issues I noticed too.


leecow:
Fixing up my generator code as part of this review to resolve the root cause. Will approve and merge this PR when that's done. 

cc @vivmishra

PR #2341, opened 2019-02-25
Target .NET Core 2

NextTurn:
Why not?

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -7,10 +7,8 @@ This directory contains samples that you can use to test out [.NET Core](http://
 In order to run these samples, you first need to [install .NET Core](http://dotnet.github.io/getting-started/). After that, you can clone this repo, go into each of the samples folders and either:
 
 * Run from source using the following commands:
-	* `dotnet restore`
 	* `dotnet run`
 * Compile and run using the following commands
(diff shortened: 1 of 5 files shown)


KexyBiscuit:
Because .NET Core 2.0 is [out of support](https://github.com/dotnet/core/blob/master/microsoft-support.md#out-of-support-net-core-releases).

NextTurn:
Thanks @KexyBiscuit. Target .NET Core 2.2 then.

samples/dotnetsay/dotnetsay.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
```
karelz: The sample README says 2.1 or higher (twice) - maybe it is ok to leave it at 2.1? Shouldn't it build just fine even if only 2.2 is installed?

NextTurn: I guess 2.0 is just OK.

PR #2367, opened 2019-03-03
Enforce coding styles

NextTurn:
It's odd that the CLI produces files encoded in UTF-8 with BOM, while files in most repos are encoded in UTF-8 (without BOM).


NextTurn:
How can BOM not be at the start?

karelz:
I don't follow ... your PR is merged. Are you pushing against it and suggesting it should be rolled back?
Or are you asking about CLI behavior? (that would be a CLI repo bug report, not question here)

To be honest, personally I don't have strong opinions about BOMs, as soon as major editors work fine :) (I am practical).

NextTurn:
It's nothing... I don't like BOMs because they are causing problems everywhere inadvertently. [dotnet/docs#10963](https://github.com/dotnet/docs/pull/10963)

PR #2401, opened 2019-03-07
Enforce coding styles

NextTurn:


NextTurn:
Just default behaviors of CLI...

NextTurn:
And to be consistent with the full sample [samples/dotnetsay/dotnetsay.csproj](https://github.com/dotnet/core/blob/master/samples/dotnetsay/dotnetsay.csproj).

PR #2408, opened 2019-03-08
Update links in dotnetsay readme

NextTurn:


karelz:
Thanks @NextTurn!

NextTurn:
I don't know why you cared about where it was when I updated a small link.

karelz:
@NextTurn are you refering to @leecow's rename above? https://github.com/dotnet/core/pull/2408#event-2190843879
In general, the more descriptive PR descriptions and titles, the better - why do you think name of the tutorial hurts to be in the title?
BTW: It would be nice to have brief descriptions in PR Description, but for small changes where title describes it, it should be fine.

NextTurn:
Don't you think my last modification is more descriptive?

karelz:
Last one is better. But it was done after merge, so it is not in commit message where it matters most for future :(

PR #2498, opened 2019-03-24
Fixed install guide that breaks shell

Seikilos:
```export PATH=$HOME/dotnet```
Replaces path contents, essentially breaking the shell.
Fixed for latest releases.


leecow:
Thanks!

PR #2848, opened 2019-06-09
Add diagnosis doc for common Linux package repository install failures

dagood:
I wrote this up to describe some common issues with installing from the Microsoft package repositories. Without running some commands first, the error messages don't have enough info to do anything at all with, so I'm hoping we can point people to this doc maybe even before they file an issue. I haven't done much polishing so it might not be ready for that yet.

I'm a little unsure about pointing people to binary installs so directly. I'm concerned we might not hear about package repository issues that way, and end up with people using outdated SDKs for longer because they have to manually update that way. However, it is necessary for non-x64 at least, because the package repo doesn't support them.

https://github.com/dotnet/core-setup/issues/6116

### DIFF
--- a/Documentation/linux-install-diagnosis.md
+++ b/Documentation/linux-install-diagnosis.md
@@ -1,0 +1,142 @@
+# Linux package manager install failure diagnosis
+
+Before reporting a package manager install failure, check if your error message is on the list below. There are workarounds for some common failures, and for others, there are diagnosis steps you should run to reveal critical information.
+
+The Linux package manager install instructions are located at https://dotnet.microsoft.com/download.
+
+
+## `E: Unable to locate package dotnet-sdk-2.2`
+
+After running `apt-get install dotnet-sdk-2.2`, you may see:
+
+```
+$ apt-get install dotnet-sdk-2.2
+Reading package lists... Done
+Building dependency tree
+Reading state information... Done
+E: Unable to locate package dotnet-sdk-2.2
+E: Couldn't find any package by glob 'dotnet-sdk-2.2'
+E: Couldn't find any package by regex 'dotnet-sdk-2.2'
+```
+
+This is probably because the Microsoft package repository isn't set up correctly on your machine.
+
+The following commands give more info about the error and the current state of your machine. Please run them and include the output in your issue report:
+
+```
+# This shows the package repositories you have configured. (Source: https://askubuntu.com/a/741948)
+grep -r --include '*.list' '^deb ' /etc/apt/sources.list /etc/apt/sources.list.d/
+
+# These commands show processor and Linux distribution information.
+uname -a
+cat /etc/os-release
+
+# You can copy this entire block of text and paste it into your terminal for convenience.
+```
+
+Next are common issues and resolutions:
+
+### Attempting to use the package repository on a non-x64 device
+The Microsoft package repository only works on x64 (`x86_64`). For example, a Raspberry Pi uses an ARM architecture, so you can't install .NET Core on that device using the package repository.
+
+### Package feed not installed due to bug in `packages-microsoft-prod.deb`
+The output from the commands above should include a line like this:
+
+```
+/etc/apt/sources.list.d/microsoft-prod.list:deb [arch=amd64] https://packages.microsoft.com/ubuntu/18.10/prod cosmic main
+```
+
+First, try reinstalling the package to work around a bug in an old version that improperly handles upgrades:
+
+```
+sudo dpkg --purge packages-microsoft-prod && sudo dpkg -i packages-microsoft-prod.deb
+sudo apt-get update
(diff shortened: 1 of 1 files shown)


Documentation/linux-install-diagnosis.md
```diff
@@ -0,0 +1,144 @@
+# Linux package manager install failure diagnosis
```
dleeapho: I think it would be good to clarify the scope of the diagnosis to the packages we put up in the Microsoft feed as opposed to distro maintained feeds (eg, Red Hat).

Documentation/linux-install-diagnosis.md
```diff
@@ -0,0 +1,144 @@
...
+The output from the commands above should include a line like this:
+
+```
+/etc/apt/sources.list.d/microsoft-prod.list:deb [arch=amd64] https://packages.microsoft.com/ubuntu/18.10/prod cosmic main
+```
+
+First, try reinstalling the package to work around a bug in an old version that improperly handles upgrades:
+
+```
+sudo dpkg --purge packages-microsoft-prod && sudo dpkg -i packages-microsoft-prod.deb
```
leecow: The deb will probably need to be downloaded (or found) before attempting to install.

Documentation/linux-install-diagnosis.md
```diff
@@ -0,0 +1,144 @@
...
+# You can copy this entire block of text and paste it into your terminal for convenience.
+```
+
+Next are common issues and resolutions:
+
+### Mismatched package repository vs. Linux distro/version
+The list of package repositories should include a `microsoft-prod.list` line, like this one:
+
+```
+/etc/apt/sources.list.d/microsoft-prod.list:deb [arch=amd64] https://packages.microsoft.com/ubuntu/18.10/prod cosmic main
```
leecow: Should make it clear that this is output from the previous grep command and is not the actual contents of microsoft-prod.list.

Documentation/linux-install-diagnosis.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+Notice `ubuntu/18.10` in that example. That indicates that repository is meant for Ubuntu 18.04. If you try to use it on Ubuntu 19.04, for example, installation will fail.
```
leecow: > That indicates that repository is meant for Ubuntu 18.04

Should be 18.10 to match the beginning of the example.

Documentation/linux-install-diagnosis.md
```diff
@@ -0,0 +1,144 @@
...
+## `W: Failed to fetch https://packages.microsoft.com/ubuntu/14.04/prod/dists/trusty/main/binary-amd64/Packages Hash Sum mismatch`
+
+This indicates a service outage in the Microsoft repository, not limited to .NET Core. Please file an issue so we can have it fixed as soon as possible!
+
+---
+
+
+## Alternative install methods
+
+The Microsoft package repository is the preferred way to install .NET Core, but there are alternatives you can use as a workaround.
```
leecow: I'm conflicted about the 'preferred' statement here. I can see us moving to a position where Snaps are the preferred method in a relatively short time frame. Perhaps say "is the most common way to install"?

Documentation/linux-install-diagnosis.md
```diff
@@ -0,0 +1,144 @@
...
+
+First, try reinstalling the package to work around a bug in an old version that improperly handles upgrades:
+
+```
+sudo dpkg --purge packages-microsoft-prod && sudo dpkg -i packages-microsoft-prod.deb
+sudo apt-get update
+sudo apt-get install dotnet-sdk-2.2
+```
+
+If that doesn't work, there are commands listed on each download page that show how to directly install the feed rather than using `packages-microsoft-prod.deb`. These commands are specific to your Linux distro and version.
```
bartonjs: Perhaps there should be a link to https://github.com/dotnet/core/blob/master/Documentation/linux-setup.md here?  Something like

If that doesn't work, find the manual installation instructions for your Linux distro and version at https://github.com/dotnet/core/blob/master/Documentation/linux-setup.md.

dagood:
I think we should add an early diagnostic command to make sure package is the Microsoft build rather than a community build from source. Related: https://github.com/dotnet/core/issues/3577

dagood:
Closing. I don't think we need this anymore. There have been a number of improvements to the docs and a few to the packages themselves.

(One particularly troublesome issue, the broken `packages-microsoft-prod.deb`, hasn't shown up recently. Since the rate will naturally go down over time, I don't think there's value in making it more visible at this point.)

`W: Failed to fetch ... Hash Sum mismatch` is still a huge issue, but I have https://github.com/dotnet/docs/issues/16670 tracking addressing it directly in the install docs. The docs are the ideal place to address these things--now that we can modify it without as much overhead as there was on the site (the docs are mostly just markdown), we should do that.

PR #2855, opened 2019-06-12
Updated rhel8 and fedora support in the release notes.

RheaAyase:


RheaAyase:
Marked as WIP - I'll find out what does CentOS look like and maybe add that next to rhel8.

RheaAyase:
Although dotnet package did not fail the rhel-centos rebuild, since it isn't released yet I guess that we can't include it ;)

PR #3111, opened 2019-07-29
add mac path issue to known issues

leecow:


release-notes/2.1/2.1-known-issues.md
```diff
@@ -1,8 +1,18 @@
 # .NET Core 2.1 Known Issues and Resolutions
 
+## .NET Core 2.1 SDK, all releases
+
+### `PATH` intermittently not updated when installing on MacOS ([core-sdk/3331](https://github.com/dotnet/core-sdk/issues/3331))
+
+There are some situations, which we do not yet understand, where PATH is not properly updated after installing the .NET Core SDK on MacOS. The result is a `dotnet: command not found` error reported in the terminal when attempting to run a `dotnet` command.
```
KathleenDollard: We have a global tools scenario where the path isn't set until you restart the shell. Will some scenarios be resolved by restarting active shells? If so, we might add that here. 

@wli3

wli3: > There are some situations, which we do not yet understand

We do. Only bash shell is fully supported. If the user is on zsh or fish shell they will see the error. And manually add ~/.dotnet/tools/ to PATH according to the shell is needed. (there is instruction for zsh in 3.0).

When restart session is needed, we will print "please restart session" after the tool is installed.

wli3: First time I heard this issue. restarting active shells might help. if the shell is opening while installing the pkg

PR #3122, opened 2019-07-30
Releases.json fixes

arthurrump:
Here's some more @leecow


leecow:
I'm in the middle of some changes to pull the 2.2.401 sdk into the 2.2.6 sdks collection. Will include your updates in that PR since I've already pushed part of the changes.

PR #3123, opened 2019-07-30
Move 2.2.401 sdk into the the 2.2.6 release SDKS collection and update missing vs-versions

leecow:
The intent of the SDKs collection is to enable us to group SDKs within the supported runtime. This means that 2.2.401 should have been added inside of the 2.2.6 release. I realize there’s a release-date missing to make this completely right but will need to determine if that’s necessary.

Also including missing vs-version fixups discovered by @arthurrump.

### DIFF
--- a/release-notes/2.2/releases.json
+++ b/release-notes/2.2/releases.json
@@ -205,7 +205,7 @@
           "version": "2.2.401",
           "version-display": "2.2.401",
           "runtime-version": "2.2.6",
-          "vs-version": "",
+          "vs-version": "16.2.0",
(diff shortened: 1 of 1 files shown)


arthurrump:
For the sake of completeness, I'd be in favor of adding the release date to SDKs as well. With the current setup there is no way of knowing that an SDK was released at a later date than other SDKs or the runtime.

Also: why only do this for 2.2.401 and not for 2.1.801? That's exactly the same situation, right? 
(If you're interested in backporting this change to older releases, please let me know. I'm planning to write that logic for [versionsof.net](https://versionsof.net) anyway, so I could just as well change it at the source then.)

PR #3237, opened 2019-08-22
Fix RHEL6 doc

janvorli:
While looking into problems a customer had with running .NET core on
RHEL6, I've found that there are two typos and one misleading
information.

This change fixes it.

### DIFF
--- a/Documentation/build-and-install-rhel6-prerequisites.md
+++ b/Documentation/build-and-install-rhel6-prerequisites.md
@@ -7,7 +7,7 @@ This document describes the prerequisites required to run .NET Core applications
 * openssl
 * libnghttp2
 * libidn
-* krb5
+* krb5-libs
 * libuuid
 * lttng-ust
 * zlib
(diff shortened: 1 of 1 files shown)


Documentation/build-and-install-rhel6-prerequisites.md
```diff
@@ -94,8 +94,9 @@ tar -xf icu4c-57_1-RHEL6-x64.tgz -C /
 ```
 Then set `LD_LIBRARY_PATH` to `/usr/local/lib` before running the application:
 ```sh
-export LD_LIBRARY_PATH=/usr/local/lib
+LD_LIBRARY_PATH=/usr/local/lib your_dotnet_app
 ```
+**Important** - please make sure that you do not set the LD_LIBRARY_PATH using `export` in your current shell or even globally for your session. Linux applications that depend on the default CURL (like the `yum`) would stop working.
```
jkotas: The CURL handler is optional and it is going to be deleted post .NET 3.0. Do we need to complicate the instruction for it?

The CURL handler does not work on number of Linux distros out of the box and nobody really cares. It may be better to just stay that the CURL handler does not work on RH6 and not complicate the instructions for it.

janvorli: I guess people will still use .NET Core 2.2 / 2.1 on RHEL 6. So I'd prefer keeping the details in, but we can say that it is optional for 3.0+

PR #3314, opened 2019-09-05
Consistent naming and casing across preview 8 and preview 9

Anipik:

PR #3320, opened 2019-09-06
Fixed preview version number in API Diff.

jroessel:


ahsonkhan:
This was already fix in https://github.com/dotnet/core/pull/3314, but that had a bunch of other changes.

PR #3329, opened 2019-09-07
Releases.json fixes

arthurrump:
Ping @leecow


leecow:
Hey Arthur - I'll be publishing a schema for the releases.json before too long (likely October). We're using it now for internal validation of the files before publishing. One thing I've done for a number of the fields is to allow an empty string or null. 

```
        "vs-version": {
          "type": ["string", "null"],
          "description": "Version of Visual Studio which first includes the release",
          "examples": ["15.9.13, 16.1.3, 16.2.0"]
```

arthurrump:
Ah ok, I'll fix that on my end then.

PR #3423, opened 2019-09-22
Correct F# versions

cartermp:
F# 4.7 has been in .NET Core 3.0 since preview 7


arthurrump:
The 3.0.0 final release also has version 4.6 listed. 
@leecow Is this something that should also be fixed on your end?

leecow:
Yep, thanks for the cc.

PR #3471, opened 2019-09-25
Remove resx-in-subfolder known issue

rainersigwald:
This was resolved before RTW by microsoft/msbuild#4702.


vivmishra:
Can you also add the RESOLVED tag in preview known issues file?

PR #3519, opened 2019-09-30
Adding Asp.net diff between netcoreapp2.2 and netcoreapp3.0

Anipik:
cc @davidfowl @danroth27 @mkArtakMSFT @anurse @wtgodbe @ericstj @terrajobst @ahsonkhan @danmosemsft @stephentoub


davidfowl:
Somethings aren't right in this PR:

https://github.com/dotnet/core/blob/998a01174da44a28d0e95f63972dec6c3a170b4f/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net/3.0.0_Microsoft.AspNetCore.md

Microsoft.AspNetCore is showing up as completely new which isn't right.

anurse:
> Microsoft.AspNetCore is showing up as completely new which isn't right.

I think that's my bad. I removed the `Microsoft.AspNetCore` package from the set I gave @Anipik because I thought it was a metapackage.

@Anipik I've updated the 2.2.0-final-api share to include `Microsoft.AspNetCore` now. @davidfowl is right, there shouldn't really be any change to that API in 3.0.

Anipik:
@davidfowl @anurse I added the Micrsoft.AspNetCore to 2.2 as well

davidfowl:
More issues:

https://github.com/dotnet/core/blob/7dc0bd09a9243241241ee372c09045d5f4864253/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net/3.0.0_Microsoft.AspNetCore.Authentication.md

This shows up like we changed a ton of existing API because of things like:

```
public Nullable<bool> AllowRefresh { get; set; }
+        public bool? AllowRefresh { get; set; }
```

Seems like there's a change with the tool?

davidfowl:
https://github.com/dotnet/core/blob/7dc0bd09a9243241241ee372c09045d5f4864253/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net/3.0.0_Microsoft.AspNetCore.Hosting.md

@anurse why is UseAzureAppServices showing up as new?

davidfowl:
@Anipik are you still looking at this?

Anipik:
@davidfowl i will try to debug the reason behind the nullable stuff tomorrow

davidfowl:
@Anipik Thanks! 

More feedback:
- Remove these assemblies
    - System.Diagnostics
    - System.Net.Http
    - System.Web.Http
- Microsoft.Extensons.* is showing up as completely new - e.g. https://github.com/dotnet/core/blob/7dc0bd09a9243241241ee372c09045d5f4864253/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net/3.0.0_Microsoft.Extensions.Primitives.md

Are we missing the 2.2 versions completely?

Anipik:
I removed the system* libraries. Microsoft.Extensions were not present in the 2.2 share so I just added that.
Similarly Microsoft.EntityFramework was absent from 3.0 share so I corrected that.

The Nullable error was due to missing dependcies due to which it was not able to map int? -> nullable<int> so it should be fixed now.

davidfowl:
@Tratcher can you take a look at these:

- https://github.com/dotnet/core/blob/db0b2970e40b79484828f45cf549bfe25ecbe762/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net/3.0.0.md
- Why does cookies and google say we removed all the properties:
 - [Microsoft.AspNetCore.Authentication.Cookies](https://github.com/dotnet/core/blob/db0b2970e40b79484828f45cf549bfe25ecbe762/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net/3.0.0_Microsoft.AspNetCore.Authentication.Cookies.md)
 - [Microsoft.AspNetCore.Authentication.Google](https://github.com/dotnet/core/blob/db0b2970e40b79484828f45cf549bfe25ecbe762/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net/3.0.0_Microsoft.AspNetCore.Authentication.Google.md)

Tratcher:
Cookies: We cleaned up obsolete properties after moving to a CookieBuilder pattern in an earlier release.
Google: Newtonsoft.Json breaking change, JObject APIs were removed and/or replaced. In this case that helper wasn't very useful so we didn't bother replacing it. Use ClaimsActions instead.

davidfowl:
@Tratcher gotcha! Thanks

davidfowl:
https://github.com/dotnet/core/blob/db0b2970e40b79484828f45cf549bfe25ecbe762/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net/3.0.0_Microsoft.AspNetCore.Http.md

What are these methods:

```
public static class HttpResponseExtensions {
+        public static void AddCspHeaders(IHeaderDictionary headers, CspOptions options, string cspHeader);
+        public static void AddScriptCspHeaders(this HttpResponse response, CspOptions options, string hash);
+        public static void AddStyleCspHeaders(this HttpResponse response, CspOptions options, string hash, string frameSources);
+        public static void RedirectToAbsoluteUrl(this HttpResponse response, string url);
+        public static void SetCache(this HttpResponse response, int maxAge, params string[] varyBy);
+        public static void SetNoCache(this HttpResponse response);
+        public static Task WriteHtmlAsync(this HttpResponse response, string html);
+        public static Task WriteJsonAsync(this HttpResponse response, object o, string contentType = null);
+        public static Task WriteJsonAsync(this HttpResponse response, string json, string contentType = null);
+    }
```

Tratcher:
@davidfowl I don't know, I can't even find those in the repo. They look like things MVC would add. @rynowak ?

davidfowl:
> @davidfowl I don't know, I can't even find those in the repo. They look like things MVC would add. @rynowak ?

I can't find them in the repo.

ahsonkhan:
> What are these methods:

Could this be it? The public signatures look like it.

https://github.com/IdentityServer/IdentityServer4/blob/5a1c1f619fff365067bea16e56af729a2db2e507/src/IdentityServer4/src/Extensions/HttpResponseExtensions.cs#L17

davidfowl:
Great! Omg those really should be internal 😬. Can we remove that binary from the API diff?

davidfowl:
@Anipik Can you remove all non System.* or Microsoft.* packages as well. Seems like we have 3rd party binaries in here.

davidfowl:
@Anipik Some more tweaks:
- Remove Microsoft.EntityFrameworkCore.*
-  Nullable is still broken [here](https://github.com/dotnet/core/blob/779c6ab2bb36aa9c7fb22caa2233e0a8c41c81b1/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net/3.0.0_Microsoft.Extensions.Caching.SqlServer.md)
- IdentityServerBuilder extension are still showing up [here](https://github.com/dotnet/core/blob/779c6ab2bb36aa9c7fb22caa2233e0a8c41c81b1/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net/3.0.0_Microsoft.Extensions.DependencyInjection.md)

I think we need to re-generate the whole thing, the nullable issue is in a bunch of different diff files and things are looking out of date.

Anipik:
> IdentityServerBuilder extension are still showing up here

IdentityServerBuilder is not coming from a third party binaries its coming from Microsoft.AspNetCore.ApiAuthorization.IdentityServer . I don't think we want to remove this binary from api diff ?

> Nullable is still broken

Its a closure problem, when we remove dependencies before api diff, then some of the types are not resolved properly and we end with this. The solution here would be to do api-diff using assemblies and after getting the api-diff remove the third party .md files. would that be fine for Asp.net ?

davidfowl:
> IdentityServerBuilder is not coming from a third party binaries its coming from Microsoft.AspNetCore.ApiAuthorization.IdentityServer . I don't think we want to remove this binary from api diff ?

Ah, good catch. That stays.

> Its a closure problem, when we remove dependencies before api diff, then some of the types are not resolved properly and we end with this. The solution here would be to do api-diff using assemblies and after getting the api-diff remove the third party .md files. would that be fine for Asp.net ?

How are dependencies missing here https://github.com/dotnet/core/blob/779c6ab2bb36aa9c7fb22caa2233e0a8c41c81b1/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net/3.0.0_Microsoft.Extensions.Caching.SqlServer.md? Is it because we don't have the BCL?

Anipik:
@davidfowl  I generated an assembly based diff and there is no Nullable here in thsi subfolder. https://github.com/Anipik/core/tree/temp3/release-notes/3.0/preview/api-diff/3.0%20vs%202.2/Asp.Net 

Let me know if that's fine and I can push the latest commit here.

davidfowl:
@Anipik that list looks good. The summary view still has links to `Microsoft.IdentityModel.*` which 404, besides that I think we're good.

Anipik:
>@Anipik that list looks good. The summary view still has links to Microsoft.IdentityModel.* which 404, besides that I think we're good.

I checked the summary in this PR and The links are not present. could it be you are looking the link I mentioned in the comment ?

Merging this one as https://github.com/dotnet/core/pull/3526  requires this to get merged.

davidfowl:
Why is this under preview. Shouldn;t it be here https://github.com/dotnet/core/tree/master/release-notes/3.0/api-diff

PR #3547, opened 2019-10-04
Updating Known Issues, adding WPF known issue for gRPC

grubioe:
Added a known issue for WPF.


grubioe:
@vatsan-madhavan FYI, documented the gRPC issue as a starting point.

PR #3595, opened 2019-10-10
Moving apidiff out of preview

Anipik:
Related to https://github.com/dotnet/core/pull/3519#issuecomment-540684215


Anipik:
> I guess this is OK? I presume you looked for any references across documents that might need to be updated?

Yes no links are broken after the move.

PR #3650, opened 2019-10-17
Fix minimum Windows Server version

bartonjs:
2012 R2 doesn't have an SP1 (leftover from when the minimum was 2008 R2 SP1).


bartonjs:
This change has already be made (https://github.com/dotnet/core/commit/f24e7e7dba5900c9effc9e50979da7875f0df714)

PR #3657, opened 2019-10-18
Update main releases.json file

### LINKED ISSUE #3656: releases.json not current for Azure DevOps Server installer
We are using the .net core build tasks from within the Azure DevOps Server. The ".NET Core SDK Installer" Task uses the "releases.json" file from this Repo to get Version information but it seems to be not current.

wenz:
Fixes #3656

Adds the latest released stable .NET Core versions


leecow:
Thanks @wenz . This file is deprecated and will soon be removed. See https://github.com/dotnet/announcements/issues/87. We're behind on pulling this file down because we became aware of a few high volume users that had not yet migrated to the new format and didn't want to break them. Expect this version to be removed by the end of the year.

wenz:
I (probably) work with one of those users and they needed the recent releases, so at least the effort was not useless ;)

PR #3838, opened 2019-11-19
Artifacts for Nov 2019 release

vivmishra:


release-notes/3.1/releases.json
```diff
@@ -413,6 +413,7 @@
       ],
       "aspnetcore-runtime": {
         "version": "3.1.0-preview3.19555.2",
+        "version-display": "3.1.0-preview3.19555.2",
```
rowanmiller: This change will need to be reverted. There is already a display-version on this and the new one is incorrect.

rowanmiller: BTW it doesn't break the pipeline because JSON.NET just picks up the value from the second listing of the item. But it's invalid JSON, so we should revert it.

PR #4315, opened 2020-02-25
Solve erorr

SagiKaganovsky:
error: This revision of snap "dotnet-sdk" was published using classic
       confinement and thus may perform arbitrary system changes outside of the
       security sandbox that snaps are usually confined to, which may put your
       system at risk.

       If you understand and want to proceed repeat the command including
       --classic.


leecow:
cc @rbhanda @vivmishra

PR #4418, opened 2020-03-19
Fix preview1 link

marek-safar:


arhen:
I think, it should be preview1, not preview.1
so, rename file is the case.

PR #4420, opened 2020-03-19
Fix Net 5 Preview filename 

arhen:
Based on pattern of Net 1.0 - Net Core 3.0, the filename should be preview1, not preview.1


leecow:
Thanks, the preview version formatting has changed for 5.0 so `preview.N` is correct. The versioned md files will use this new convention.

PR #4920, opened 2020-07-07
Update microsoft-support.md

marcpopMSFT:
Per guidance from Ux design, suggested alternative title text


marcpopMSFT:
@leecow Jamshed said you could review and approve this change. I got this guidance from Brick.

PR #5045, opened 2020-08-10
Remove duplicate dotnet-hosting

leecow:
2.1.13 release had a duplicate dotnet-hosting entries in the runtime and aspnetcore sections. Convention is list this file under aspnetcore so removing the duplicate entry from the runtime section.


joeloff:
Thanks Lee

PR #5108, opened 2020-08-24
Simpler calls to ReadAllLines by using using static.

cincuranet:
Not sure it's worth it, but given the `Console.WriteLine` is using it, I'll give it a shot.


cincuranet:
@richlander

PR #5377, opened 2020-10-16
Update releases.json

leecow:
Correct 2.1.402 Win x86 entries which had mismatched names and download targets.

### DIFF
--- a/release-notes/2.1/releases.json
+++ b/release-notes/2.1/releases.json
@@ -9605,7 +9605,6 @@
           },
           {
             "name": "dotnet-sdk-win-gs-x86.exe",
-            "rid": "win-x86",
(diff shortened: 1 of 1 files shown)


rbhanda (approved):

release-notes/2.1/releases.json
```diff
@@ -9622,7 +9622,7 @@
             "hash": "405cbd7c65d63b36e3bd6bcdfc897ac6474c4eaf93db9db478a80ab511bfa7a1c4a84024cc6e4af0df0af86bcc0a1a96a8ba0864c77bf579f32bce437c28d5a8"
```
joeloff, line 9622: Can we add the missing Rid property for x86 too?

leecow: good catch - done

joeloff (approved):

PR #5489, opened 2020-10-29
Minor grammar correction.

lucas-andre:
Removing the unnecessary comma in the compound subject of line 41 of the global README.md.
A compound subject consists of two or more subjects. Commas separate the subjects when there are three or more. A compound subject with two items requires no punctuation to separate them.


dagood:
> A compound subject with two items

I read it as three items:

> [Instances of] abusive [behavior], [instances of] harassing [behavior], or [instances of] otherwise unacceptable behavior

Then it's just an [oxford comma](https://docs.microsoft.com/en-us/style-guide/punctuation/commas#use-a-comma), if I'm not mistaken.

lucas-andre:
You're right. Thank you for your attention to detail that I missed.

PR #5500, opened 2020-10-31
Update os-lifecycle-policy.md

jamshedd:
This is an update to the OS lifecycle support page to help address conflicts between our stated policy (we support OS platforms when they're in support and drop them when they're not) and specific choices we made to include Win7 SP1 as a supported platform for 5.0.

I have added a note to the OS lifecycle support page to clarify when there is a conflict between the general policy and the list of specific platforms documented for a given version the latter will supersede the former.


GrabYourPitchforks:
Nit: This could be construed as "we support .NET 5 on Win7 SP1." Where I think we actually meant "we support .NET 5 on Win7 SP1 _only if_ your Win7 SP1 deployment is running under a custom ESU support license." This would clarify that you can't run Win7 SP1 at home and expect support for .NET 5.

jamshedd:
> Nit: This could be construed as "we support .NET 5 on Win7 SP1." Where I think we actually meant "we support .NET 5 on Win7 SP1 _only if_ your Win7 SP1 deployment is running under a custom ESU support license." This would clarify that you can't run Win7 SP1 at home and expect support for .NET 5.

Ack, I'm adding a note about the need to install Win7 ESUs on each of the version specific (2.1, 3.1, 5.0) pages.

rbhanda (approved):

leecow (approved):

PR #5530, opened 2020-11-07
Update ICU breaking change in known issues with more details

safern:

PR #5537, opened 2020-11-09
Revert "Update ICU breaking change in known issues with more details"

safern:
Reverts dotnet/core#5530 as it is part of: https://github.com/dotnet/core-private/pull/75


rbhanda:
Reverting because the change is tracked in https://github.com/dotnet/core-private/pull/75

PR #5576, opened 2020-11-11
Update releases.json for 5.0 with EOL date

rbhanda:


mairaw:
That expects a full date:
image

We might need to do some site changes to support other date formats.

davidest623 (approved):

PR #5578, opened 2020-11-12
Add dependadotnet tool

richlander:
- From https://github.com/richlander/dependadotnet.
- Updated.


samples/dependadotnet/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+    if (filename.Length < 8)
+    {
+        return false;
+    }
+
+    return filename.Substring(filename.Length - 7) switch
+    {
+        ".csproj" or ".fsproj" or ".vbproj" => true,
+        _ => false
+    };
```
BillWagner, line 83: I think this can all be replaced with `Path.GetExtension`:

```suggestion
    return Path.GetExtension(filename) switch
    {
        ".csproj" or ".fsproj" or ".vbproj" => true,
        _ => false
    };
```

samples/dependadotnet/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    if (!IsProject(file))
+    {
+        continue;
+    }
+
+
+    string filename = Path.GetFileName(file);
+    string? parentDir = Path.GetDirectoryName(file);
+    string relativeDir = parentDir?.Substring(path.Length).Replace('\\','/') ?? "/";
```
BillWagner, line 46: Consider using the `SeparatorChar` and `AltSeparatorChar`. This should work on both Windows and unix:

```suggestion
    string relativeDir = parentDir?.Substring(path.Length).Replace(Path.DirectorySeparatorChar,Path.AltDirectorySeparatorChar) ?? Path.AltDirectorySeparatorChar.ToString();
```

BillWagner (approved):
This looks really good @richlander 

I had a couple suggestions to leverage some of the File IO routines that are built in. Decide what you want to do, and then this is ready to merge.

Issue #5646, opened 2020-11-23
Describe platform dependencies


mthalman:
### Summary

Provides a uniform description of all platform dependencies for .NET.  This is the basis for all of the other user stories in this [epic](https://github.com/dotnet/core/issues/5651).  It collects the dependency data into a machine-readable format that serves as the complete database for reasoning about .NET's dependencies.

### Goal

.NET contributors can describe the product's platform dependencies -- both runtime and source build toolset -- in a common format that can be used to satisfy other dependent workflows that require the data.

mthalman:
A big question here is the granularity of the dependency. In the case of Linux, you could describe dependencies in terms of which packages are needed. However, there are some scenarios where that can be overly broad.

For example, in .NET 5.0, `dotnet watch` has a dependency on `pgrep` (see https://github.com/dotnet/dotnet-docker/issues/2396).  That utility is part of the `procps` package.  In Alpine, `pgrep` comes pre-installed but the `procps` package is not. So it's not accurate to say that `procps` is a requirement for `dotnet watch` in Alpine.  A more granular way would be to say that it requires just the `pgrep` utility and not even describe it in terms of packages.

Maybe cases where a required component comes pre-installed in the operating system do not need to be described here.  It's not really actionable information when configuring what to install into an environment.  Although, perhaps it is necessary from the standpoint of knowing what can be _removed_ from an environment.

mthalman:
Runtime platform dependency model for .NET 6 has been defined: https://github.com/dotnet/core/pull/6170

PR #6170, opened 2021-04-16
Add runtime platform deps for .NET 6

mthalman:
Related to https://github.com/dotnet/core/issues/5646

Defines all of .NET's runtime platform dependencies for Linux, Windows, and macOS.  This uses the [platform dependency schema](https://github.com/dotnet/designs/blob/main/accepted/2021/platform-dependencies/platform-dependencies.md#schema).

This relies on the RID inheritance model.  For example, RHEL and CentOS have the same package dependencies and the `centos` RID inherits from `rhel`.  So in the dependency model, I defined dependencies just for the `rhel` RID which would be implicitly inherited by `centos`.

I decided not to include any Windows desktop UI packages such as WPF and WinForms because I didn't see much benefit to having them. One of the main reasons for tracking Windows dependencies is to differentiate what libraries exist between the different Windows variants (Windows Server, Windows desktop, Windows Server Core, and Nano Server).  Since Windows Server Core and Nano Server are headless, using the desktop UI packages isn't really an option.  And the remaining variants, Windows Server and Windows desktop, provide all the necessary dependencies.  Based on that, producing an itemized list of what those dependencies are seems like a lot of work without much benefit.  We can re-evaluate later if we determine there's a need.

I won't actually merge this until Preview 4 ships.

### DIFF
--- a/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
+++ b/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
@@ -530,7 +530,7 @@
               "usage": "printing"
             },
             {
-              "name": "libgdiplus:6.0.1",
+              "name": "libgdiplus0:6.0.1",
               "dependencyType": "LinuxPackage",
               "usage": "default"
             },
--- a/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
+++ b/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
@@ -409,11 +409,6 @@
           "type": "Framework",
           "platformDependencies": [
             {
-              "name": "Core Foundation",
-              "dependencyType": "MacOSFramework",
-              "usage": "default"
-            },
-            {
               "name": "libobjc.dylib",
               "dependencyType": "Library",
               "usage": "default"
@@ -421,11 +416,6 @@
             {
               "name": "libproc.dylib",
               "dependencyType": "Library",
-              "usage": "default"
-            },
-            {
-              "name": "System Configuration",
-              "dependencyType": "MacOSFramework",
               "usage": "default"
             }
           ]
@@ -456,11 +446,6 @@
           "name": "System.Drawing.Common",
           "type": "NuGetPackage",
           "platformDependencies": [
-            {
-              "name": "Carbon Framework",
-              "dependencyType": "MacOSFramework",
-              "usage": "default"
-            },
             {
               "name": "libgdiplus.dylib:6.0.1",
               "dependencyType": "Library",
--- a/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
+++ b/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
@@ -305,7 +305,7 @@
               "usage": "default"
             },
             {
-              "name": "libx11-dev",
+              "name": "libX11-devel",
               "dependencyType": "LinuxPackage",
               "usage": "xwindows"
             }
@@ -393,7 +393,7 @@
               "usage": "default"
             },
             {
-              "name": "libx11-dev",
+              "name": "libX11-devel",
               "dependencyType": "LinuxPackage",
               "usage": "xwindows"
             }
@@ -535,7 +535,7 @@
               "usage": "default"
             },
             {
-              "name": "libx11-dev",
+              "name": "libX11-devel",
               "dependencyType": "LinuxPackage",
               "usage": "xwindows"
             }
@@ -623,7 +623,7 @@
               "usage": "default"
             },
             {
-              "name": "libx11-dev",
+              "name": "libX11-devel",
               "dependencyType": "LinuxPackage",
               "usage": "xwindows"
             }
--- a/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
+++ b/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
@@ -980,6 +980,11 @@
               "usage": "default"
             },
             {
+              "name": "gdiplus.dll",
+              "dependencyType": "Library",
+              "usage": "default"
+            },
+            {
               "name": "winspool.drv",
               "dependencyType": "DeviceDriver",
               "usage": "printing"
--- a/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
+++ b/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
@@ -3,7 +3,6 @@
   "dependencyUsages": {
     "default": "Used by default or for canonical scenarios",
     "diagnostics": "Used for diagnostic scenarios such as tracing or debugging",
-    "graphics": "Used for scenarios involving graphics/drawing/fonts",
     "http.sys": "Used to access the HTTP.sys listener in Windows",
     "localization": "Used for locale and culture scenarios",
     "media": "Used for multimedia scenarios",
--- a/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
+++ b/release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
@@ -1,0 +1,1035 @@
+{
+  "dotnetReleaseVersion": "6.0.0-preview.4",
+  "dependencyUsages": {
+    "default": "Used by default or for canonical scenarios",
+    "diagnostics": "Used for diagnostic scenarios such as tracing or debugging",
+    "graphics": "Used for scenarios involving graphics/drawing/fonts",
+    "http.sys": "Used to access the HTTP.sys listener in Windows",
+    "localization": "Used for locale and culture scenarios",
+    "media": "Used for multimedia scenarios",
+    "printing": "Used for printing to printers",
+    "security": "Used for security scenarios",
+    "xwindows": "Used to access the X Window System of a Linux or Unix-based system"
+  },
+  "platforms": [
+    {
+      "rid": "alpine",
+      "components": [
+        {
+          "name": "Microsoft.NETCore.App",
+          "type": "Framework",
+          "platformDependencies": [
+            {
+              "name": "icu-libs",
+              "dependencyType": "LinuxPackage",
+              "usage": "default"
+            },
+            {
+              "name": "krb5-libs",
+              "dependencyType": "LinuxPackage",
+              "usage": "default"
+            },
+            {
+              "name": "libgcc",
+              "dependencyType": "LinuxPackage",
+              "usage": "default"
+            },
+            {
+              "name": "libintl",
+              "dependencyType": "LinuxPackage",
+              "usage": "default"
+            },
+            {
+              "name": "libssl1.1",
+              "dependencyType": "LinuxPackage",
+              "usage": "default"
+            },
+            {
+              "name": "libstdc++",
+              "dependencyType": "LinuxPackage",
+              "usage": "default"
+            },
+            {
+              "name": "lttng-ust",
+              "dependencyType": "LinuxPackage",
+              "usage": "diagnostics"
+            },
+            {
+              "name": "tzdata",
+              "dependencyType": "LinuxPackage",
+              "usage": "localization"
+            },
+            {
+              "name": "zlib1g",
+              "dependencyType": "LinuxPackage",
+              "usage": "default"
+            }
+          ]
+        },
+        {
+          "name": "System.DirectoryServices.Protocols",
+          "type": "NuGetPackage",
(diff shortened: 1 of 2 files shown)


release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
```diff
@@ -0,0 +1,1035 @@
...
+              "dependencyType": "LinuxPackage",
+              "usage": "printing"
+            },
+            {
+              "name": "libgdiplus:6.0.1",
+              "dependencyType": "LinuxPackage",
+              "usage": "default"
+            },
+            {
+              "name": "libx11-dev",
```
jkotas, line 730: Do we really have dependency on the `-dev` package? The `-dev` package typically contain static libraries, required for C/C++ compilation and linking. They should not be needed to by .NET runtime.

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
```diff
@@ -0,0 +1,1035 @@
...
+  "dotnetReleaseVersion": "6.0.0-preview.4",
+  "dependencyUsages": {
+    "default": "Used by default or for canonical scenarios",
+    "diagnostics": "Used for diagnostic scenarios such as tracing or debugging",
+    "graphics": "Used for scenarios involving graphics/drawing/fonts",
+    "http.sys": "Used to access the HTTP.sys listener in Windows",
+    "localization": "Used for locale and culture scenarios",
+    "media": "Used for multimedia scenarios",
+    "printing": "Used for printing to printers",
+    "security": "Used for security scenarios",
```
jkotas, line 11: How does one figure out whether they app has "security scenarios"? Should these descriptions rather list of the affected types, etc.?

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
```diff
@@ -0,0 +1,1035 @@
+{
+  "dotnetReleaseVersion": "6.0.0-preview.4",
+  "dependencyUsages": {
+    "default": "Used by default or for canonical scenarios",
+    "diagnostics": "Used for diagnostic scenarios such as tracing or debugging",
+    "graphics": "Used for scenarios involving graphics/drawing/fonts",
```
jkotas, line 6: "graphics" does not apper to be used anywhere

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
```diff
@@ -0,0 +1,1035 @@
...
+    },
+    {
+      "rid": "osx",
+      "components": [
+        {
+          "name": "Microsoft.NETCore.App",
+          "type": "Framework",
+          "platformDependencies": [
+            {
+              "name": "Core Foundation",
```
jkotas, line 413:
```suggestion
              "name": "CoreFoundation",
```
The framework name is without the space.

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
```diff
@@ -0,0 +1,1035 @@
...
+              "dependencyType": "MacOSFramework",
```
jkotas, line 414: We depend on number of other macOS frameworks. Is it really useful to list them all since they all come with macOS by default?

jkotas: dtto for Windows OS .dlls

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
```diff
@@ -0,0 +1,1035 @@
...
+              "dependencyType": "LinuxPackage",
+              "usage": "printing"
+            },
+            {
+              "name": "libgdiplus:6.0.1",
+              "dependencyType": "LinuxPackage",
+              "usage": "default"
+            },
+            {
+              "name": "libx11-dev",
```
omajid, line 554: This can't be right for RHEL. RHEL packages use `-devel` suffix. This package is named `libX11-devel` on RHEL.

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
```diff
@@ -0,0 +1,1035 @@
...
+          "name": "System.Drawing.Common",
+          "type": "NuGetPackage",
+          "platformDependencies": [
+            {
+              "name": "cups-libs",
+              "dependencyType": "LinuxPackage",
+              "usage": "printing"
+            },
+            {
+              "name": "libgdiplus:6.0.1",
```
omajid, line 549: `libgdiplus` is not a part of RHEL. I suppose users can install it from some third-party repository?

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
mthalman: That's what I thought but it is required right now.  After looking at what gets installed I see this:

Installing libx11-6:
* /usr/lib/x86_64-linux-gnu/libX11.so.6
* /usr/lib/x86_64-linux-gnu/libX11.so.6.3.0

If attempting to execute a code path that requires the X11 library with just `libx11-6` installed, you get this error:

```
System.DllNotFoundException: 'Unable to load shared library 'libX11' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: liblibX11: cannot open shared object file: No such file or directory'
```

It works after installing `libx11-dev`.  The additional file that `libx11-dev` provides is the following:
* /usr/lib/x86_64-linux-gnu/libX11.so

Should I log an issue to be able to search for the additional versioned .so file in order to remove the dependency on `-dev`?

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
```diff
@@ -0,0 +1,1035 @@
...
```
omajid, line 304: I think we can drop the version here. [All currently supported versions of Fedora (>= 32) have 6.0.4 or newer](https://koji.fedoraproject.org/koji/packageinfo?packageID=357).

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
```diff
@@ -0,0 +1,1035 @@
...
+              "dependencyType": "LinuxPackage",
+              "usage": "default"
+            },
+            {
+              "name": "libx11-dev",
+              "dependencyType": "LinuxPackage",
+              "usage": "xwindows"
```
omajid, line 311: Is there some way to test this behaviour/implementation?  The package name might be `libX11` or `libX11-devel` depending on what it's being used for. Also, with Fedora moving mostly to Wayland, I am not sure how well this would work.

mthalman: See https://github.com/dotnet/core/pull/6170#discussion_r615006176.  The test code I've used to exercise the X11 path is the following:

```csharp
var g = Graphics.FromImage(new Bitmap(10, 10));
g.CopyFromScreen(new Point(0, 0), new Point(1, 1), new Size(1, 1));
```

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
omajid: > Should I log an issue to be able to search for the additional versioned .so file in order to remove the dependency on -dev?

IMO, yes. The `-dev`/`-devel` packages are not generally installed on end-user systems. They are only used by developers.

mthalman: I've logged https://github.com/dotnet/runtime/issues/51404 to remove the dependency on the developer package.

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
mthalman: This is admittedly a broad way to describe the usage.  There's a balance that needs to be struck here between descriptivity and maintainability.  While I agree that it would be great if we could have a list of affected types, I'm concerned about our ability to keep such a list up-to-date.  I want the maintenance of this file to have low overhead.  Potentially we'll be able to supplement this file with results from static analysis that provides this level of granularity.  But that'd be well down the road.

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
mthalman: A valid question for Mac.  For Windows, not all Windows SKUs are equal.  For example, Nano Server is missing many libraries compared to Windows Server Core or Windows Server/Desktop.  It's helpful know what our dependencies are in order to know whether things will run in one of those systems.

jkotas: > Nano Server is missing many libraries compared to Windows Server Core

It is also missing APIs in number of libraries. For example, the full `kernel32` on Window Server has more APIs than `kernel32` on Nano Server or some of the `kernel32` APIs are stubbed out.

I am not sure how much the list of Windows .dll libraries helps with determining whether given package is going work on Nano Server.

mthalman: > It is also missing APIs in number of libraries. For example, the full kernel32 on Window Server has more APIs than kernel32 on Nano Server or some of the kernel32 APIs are stubbed out.

That's fair.  However, I'm not aware of any usage of such an API where the library exists in Nano Server but not that API.  Are there any that you know of?

OTOH, there are several known instances of libraries that are not available in Nano Server which we've directly received feedback from customers about.  Examples are the following:

* System.Data.Odbc
  * odbc32.dll
* System.DirectoryServices:
  * activeds.dll
  * netapi32.dll
* System.DirectoryServices.AccountManagement:
  * activeds.dll
  * credui.dll
* System.Drawing.Common
  * gdiplus.dll

Based on that, I do see value in tracking these libraries.

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
mthalman: Removed

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
mthalman: Yep, I've now updated to use `libX11-devel` for RHEL, Fedora, openSUSE, and SLES.

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
mthalman: The motivation for defining this is primarily to have a record of what minimum version the product depends on, not necessarily what version is available in the distro's repository.  This is particularly useful when evaluating whether some other unsupported distro is capable of running .NET.

release-notes/6.0/preview/6.0.0-preview.4-runtime-deps.json
mthalman: Yes, specifically Mono's repository.  I've updated the package name to correspond to that.

mthalman:
Any further feedback on this from anyone?

reflectronic (commented):
These suggestions don't really move the needle, but it's better to be more accurate

release-notes/6.0/runtime-deps.json
```diff
@@ -0,0 +1,1024 @@
...
+        },
+        {
+          "name": "System.Speech",
+          "type": "NuGetPackage",
+          "platformDependencies": [
+            {
+              "name": "winmm.dll",
+              "dependencyType": "Library",
+              "usage": "default"
+            }
```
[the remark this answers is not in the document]
mthalman: Good call.  I totally missed accounting for COM dependencies.

COOOOOM!
khan

release-notes/6.0/runtime-deps.json
```diff
@@ -0,0 +1,1024 @@
...
+              "usage": "default"
+            },
+            {
+              "name": "winspool.drv",
+              "dependencyType": "DeviceDriver",
+              "usage": "printing"
+            }
+          ]
+        },
+        {
```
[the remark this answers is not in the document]
mthalman: I also found that it depends on `wbemsvc.dll`.

Saiprasad16 (commented):
Remove unused dependency usage

mthalman:
> Remove unused dependency usage

@Saiprasad16 - Can you clarify this?  Your comment isn't associated with a specific line in the file.

PR #6670, opened 2021-09-09
Updated the lifecycle statement

jamshedd:
Changed from 15 months to 18 months of support through May 2022.


Varorbc:
https://github.com/dotnet/core/blob/main/release-notes/README.md

PR #7129, opened 2022-01-20
Artifacts for .NET 6.0.200 Preview SDK - VS 17.1 Preview 3

rbhanda:
Artifacts for .NET 6.0.200 Preview SDK - VS 17.1 Preview 3


leecow (changes requested):
Include the build number in the title as well. Should be similar to what we did for .NET 5 and 6 preview releases. e.g "6.0.200-preview.22055.15". This should be done for all instances of 6.0.200 in the release notes.

leecow (approved):

PR #7214, opened 2022-02-18
add schema declaration to the manifest file

baronfel:
This change allows editors to highlight data errors. While this won't entirely prevent bad data, it opens the door to using the schema in more automated fashions (for example a unit test that validates the document according to the schema).

Here's what the experience looks like for a subset of invalid field data in VS Code:

image

and in VS:

image

The tooltips are powered by the schema and offer tips for fixing the problem:

image

### DIFF
--- a/release-notes/releases-index.json
+++ b/release-notes/releases-index.json
@@ -1,4 +1,5 @@
 {
+    "$schema": "https://json.schemastore.org/dotnet-releases-index.json",
     "releases-index": [
         {
             "channel-version": "7.0",


rbhanda:
@leecow will this be a breaking change somehow to the release-index.json?

baronfel:
I checked the .NET library that parses this file and it would not impact that library because it [specifically pulls the `releases-index` property](https://github.com/dotnet/deployment-tools/blob/main/src/Microsoft.Deployment.DotNet.Releases/src/ProductCollection.cs#L122-L123) of the json document and then deserializes that array.

This library is used in the SDK, omnisharp, arcade, and probably others based on the NuGet listing, so I would hope that since it parses this correctly there would not be _too_ much churn.

leecow:
My primary question is how is the schema maintained updated in the event of an update in either the json structure (e.g. adding a property) or updating valid values (which we will likely be doing for support phase)?

axljoker03 (approved):

leecow:
This change needs to happen in the tooling which generates and maintains the file. I've submitted a [PR](https://github.com/SchemaStore/schemastore/pull/3005) to schemastore to update the releases-index schema. Once that's complete, I'll update the tooling.

PR #7301, opened 2022-03-21
Update Alpine arm32 minimum version to 3.13

am11:
This came up in https://github.com/dotnet/sdk/issues/24448#issuecomment-1073888341. The minimum supported version of musl libc in `linux-musl-arm` package is 1.2.0 (Alpine 3.13) due to the time64 change.

cc @jkotas

### DIFF
--- a/release-notes/5.0/5.0-supported-os.md
+++ b/release-notes/5.0/5.0-supported-os.md
@@ -29,6 +29,7 @@ OS                                    | Version                 | Architectures 
 OS                                    | Version               | Architectures     | Lifecycle
 --------------------------------------|-----------------------|-------------------|----------
 [Alpine Linux][Alpine]                | 3.12+                 | x64, Arm64        | [Alpine][Alpine-lifecycle]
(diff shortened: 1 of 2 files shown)


jkotas:
We have this support matrix replicate in number of places, e.g.: https://docs.microsoft.com/en-us/dotnet/core/install/linux-alpine

richlander:
That works.

Can you also update the .NET 5 and .NET Core 3.1 files or is this 6.0-specific?

am11:
We do not have .NET 3.1 package for `Linux-musl-arm` in this table: https://github.com/dotnet/installer#table.

I have updated 5.0, since the time64 fix was backported and latest build (`@(#)Version 5.0.1522.11506 @Commit: 3065735be79d6b7d17e8e3a723115810b43c9b3a`) does not work on Alpine 3.12. From the installers table, it is the one from `Release/5.0.4XX` column. However, the older version from `Release/5.0.2XX` column works well.

PR #7304, opened 2022-03-23
VS2019 VsDevCmd.bat makes dotnet use msbuild v 16

ql-mafre:
Not using any VsDevCmd.bat at all or using the VS2022 VsDevCmd.bat works.

### DIFF
--- a/release-notes/6.0/known-issues.md
+++ b/release-notes/6.0/known-issues.md
@@ -32,6 +32,18 @@ If you build .NET 6 projects with MSBuild 16.11, for example, you will see the f
 
 `warning NETSDK1182: Targeting .NET 6.0 in Visual Studio 2019 is not supported`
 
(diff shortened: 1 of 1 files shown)


axljoker03 (approved):

richlander:
PTAL @baronfel

baronfel (commented):
This is correct, but is a specific case of the general guidance that we have at the [SDK/Visual Studio Lifecycle](https://learn.microsoft.com/dotnet/core/porting/versioning-sdk-msbuild-vs#lifecycle) documentation, so I'm not sure how much we want to go back and patch up specific warnings.

richlander:
Given that .NET 6 is EOL, we should probably just close now. Yes?

richlander:
Closing due to response and since this is about an EOL .NET version.

PR #7388, opened 2022-04-19
adding 6.0.300sdk md

dcwhittaker:


mattjohnsonpint:
@dcwhittaker - was this put in the right folder? Seems odd to have it under `6.0.3`

mattjohnsonpint:
Pre-releases are always confusing.  6.0.202 is using the 6.0.4 runtime, so I would expect 6.0.300 to be there or higher?

FWIW, the confusion stems from .NET MAUI going to RC1 status, but requiring 6.0.300 while it's still in preview (on macOS anyway).  See https://github.com/dotnet/maui/issues/6082

PR #7424, opened 2022-05-04
Use "Close Stale Issues" for no-response.yml

jongalloway:
* Uses "official" [Close Stale Issues](https://github.com/marketplace/actions/close-stale-issues) action as it is better maintained and has a lot more features
* Sets `issues: write` permission to [allow comment / close on issues](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions)

If you want to stay with the original action, I can update this to just set the permissions.

Since this action has a few more options, I went with this policy:
* After 7 days, set a "Stale" tag and add a stale comment
* After 3 more days, close the issue with a closing comment

This seems like a nicer user experience, is more standard (since a lot of repos use this action), and allows filtering on / excluding by the Stale label.

### DIFF
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -15,12 +15,16 @@ on:
 jobs:
   noResponse:
     runs-on: ubuntu-latest
+    permissions:
+      issues: write
     steps:
-      - uses: lee-dohm/no-response@v0.5.0
+      - uses: actions/stale@v5
         with:
-          token: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


mairaw (approved):

jongalloway:
It looks like the policy for this repo needs to be updated to allow running this action. I'd suggest allowing all actions created by GitHub. If that's not acceptable, need to add `actions/stale` to the list.

image

PR #7481, opened 2022-05-24
Update Alpine supported versions

richlander:
@dotnet/linux-team


rbhanda:
Should we do this for .NET Core 3.1 and .NET 6.0 as well?

richlander:
Good call. Just did.

PR #7524, opened 2022-06-08
Add 7.0-preview5 SHA hash algorithm known issue

eerhardt:


bartonjs:
It can't happen for single-page requests... do we know the upper-bound on "safe" values?

Assume all bytes are valued greater than 99. So it's something like `{"digest":"SHA256","content":[]}` with the middle filled in with `Max(0, (4 * content.Length) - 1)` on a 1024-byte page.  If my identifiers and whitespace is right, then that's 32 overhead bytes, so `4L - 1 < (1024-32)` => `4L - 1 < 992`, `4L < 993`, `L < 248.25`, so the hash of anything 248 bytes or smaller is immune to the race.

Is that worth mentioning (after verifying the numbers)?

eerhardt:
I'm not sure it is worth mentioning. Could someone do something useful with that information?

If someone is affected by this - I would tell them to stay on 7.0-preview4 until 7.0-preview6 was released. That seems like the most appropriate action rather than trying to workaround it just for preview5.

bartonjs (approved):

PR #7545, opened 2022-06-16
Update .NET 6 known-issues.md

Rick-Anderson:
Fixes https://github.com/dotnet/aspnetcore/issues/42072

@Ogglas can you help me document this problem. Please review what I have and suggest new text.

### DIFF
--- a/release-notes/6.0/known-issues.md
+++ b/release-notes/6.0/known-issues.md
@@ -59,6 +59,38 @@ The first time SPA apps are run, the authority for the spa proxy might be incorr
 
 When using localdb (default when creating projects in VS), the normal database apply migrations error page will not be displayed correctly due to the spa proxy. This will result in errors when going to the fetch data page. Apply the migrations via 'dotnet ef database update' to create the database.
 
+### SPA template issues with Individual authentication when running in production
+
+<!-- Statement of problem here. Initially you say  but then say Tested with `Always on = true` for Azure App Service and the error actually occurred more frequently. So if it's not the app shutting down and restarting, what causes the error? Something like SPA apps on Azure that (conditions that cause the problem) return the following error `WWW-Authenticate: Bearer error="invalid_token", error_description="The issuer 'https://MyDomain.com' is invalid"`. If the app is accessed from the Azure DNS (MyDomain.azurewebsites.net), authenticaion is successful. Subsequent requests to `https://MyDomain.com` succeed until (??? the app is restarted??? but not according to always on = true). After stopping and starting the app, authenticaion succeeds.
+
+To prevent this problem without having to stop and restart the app:
+
+1. Add a new app setting which contains the target DNS address. For example, create `IdentityServer:IssuerUri` with value `https://MyDomain.com/`
+1. Add the following code to the app:
+```
+builder.Services.AddIdentityServer(options =>
+{
+    if (!string.IsNullOrEmpty(settings.IdentityServer.IssuerUri))
+    {
+        options.IssuerUri = settings.IdentityServer.IssuerUri;
+    }
+})
+```
+   Alternatively, add the following code:
+```
+builder.Services.AddDefaultIdentity<ApplicationUser>(options =>
+{
+    if (!string.IsNullOrEmpty(settings.IdentityServer.IssuerUri))
+    {
+        options.Tokens.AuthenticatorIssuer = settings.IdentityServer.IssuerUri;
+    }
+})
+```
+
+For more information, see [this GitHub issue](https://github.com/dotnet/aspnetcore/issues/42072)
+
+
+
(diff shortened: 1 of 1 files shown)


release-notes/6.0/known-issues.md
```diff
@@ -59,6 +59,38 @@ The first time SPA apps are run, the authority for the spa proxy might be incorr
 
 When using localdb (default when creating projects in VS), the normal database apply migrations error page will not be displayed correctly due to the spa proxy. This will result in errors when going to the fetch data page. Apply the migrations via 'dotnet ef database update' to create the database.
 
+### SPA template issues with Individual authentication when running in production
+
+<!-- Statement of problem here. Initially you say  but then say Tested with `Always on = true` for Azure App Service and the error actually occurred more frequently. So if it's not the app shutting down and restarting, what causes the error? Something like SPA apps on Azure that (conditions that cause the problem) return the following error `WWW-Authenticate: Bearer error="invalid_token", error_description="The issuer 'https://MyDomain.com' is invalid"`. If the app is accessed from the Azure DNS (MyDomain.azurewebsites.net), authenticaion is successful. Subsequent requests to `https://MyDomain.com` succeed until (??? the app is restarted??? but not according to always on = true). After stopping and starting the app, authenticaion succeeds.
```
Rick-Anderson, line 64:
```suggestion
<!-- Statement of problem here. Initially you say  but then say Tested with `Always on = true` for Azure App Service and the error actually occurred more frequently. So if it's not the app shutting down and restarting, what causes the error? Something like --> SPA apps on Azure that (conditions that cause the problem) return the following error `WWW-Authenticate: Bearer error="invalid_token", error_description="The issuer 'https://MyDomain.com' is invalid"`. If the app is accessed from the Azure DNS (MyDomain.azurewebsites.net), authenticaion is successful. Subsequent requests to `https://MyDomain.com` succeed until (??? the app is restarted??? but not according to always on = true). After stopping and starting the app, authenticaion succeeds.
```

Ogglas: ### SPA template issues with Individual authentication when running in production

Applications based on SPA template with Individual authentication that requires login for every page and is hosted as an Azure App Service on your own domain like https://MyDomain.com can receive the following error:

> WWW-Authenticate: Bearer error="invalid_token", error_description="The issuer 'https://MyDomain.com' is invalid"

Probably related to Azure DNS (MyDomain.azurewebsites.net) but this has not been verified. It will probably occur more frequently with `Always on` set to true for the Azure App Service but it can happen with `Always on` set to false as well. 

To prevent this problem without having to stop and restart the app when the error occurs:
1. Add a new app setting which contains the target DNS address. For example, create `IdentityServer:IssuerUri` with value `https://MyDomain.com/`
2. Add the following code to the app:

```
if (!string.IsNullOrEmpty(settings.IdentityServer.IssuerUri))
{
    builder.Services.Configure<JwtBearerOptions>(IdentityServerJwtConstants.IdentityServerJwtBearerScheme, o => o.Authority = settings.IdentityServer.IssuerUri);
}
```

below this code:

```
builder.Services.AddAuthentication()
    .AddIdentityServerJwt();
```

Then modify `AddIdentityServer` like this:

```
builder.Services.AddIdentityServer(options =>
                {
                    //Used until https://github.com/dotnet/aspnetcore/issues/42072 is fixed
                    if (!string.IsNullOrEmpty(settings.IdentityServer.IssuerUri))
                    {
                        options.IssuerUri = settings.IdentityServer.IssuerUri;
                    }
                })
    .AddApiAuthorization<ApplicationUser, ApplicationDbContext>();
```

For more information, see [this GitHub issue](https://github.com/dotnet/aspnetcore/issues/42072)

release-notes/6.0/known-issues.md
```diff
@@ -61,7 +61,7 @@ When using localdb (default when creating projects in VS), the normal database a
...
+<!-- Statement of problem here. Initially you say  but then say Tested with `Always on = true` for Azure App Service and the error actually occurred more frequently. So if it's not the app shutting down and restarting, what causes the error? Something like --> SPA apps on Azure that (conditions that cause the problem) return the following error `WWW-Authenticate: Bearer error="invalid_token", error_description="The issuer 'https://MyDomain.com' is invalid"`. If the app is accessed from the Azure DNS (MyDomain.azurewebsites.net), authentication is successful. Subsequent requests to `https://MyDomain.com` succeed until (??? the app is restarted??? but not according to always on = true). After stopping and starting the app, authentication succeeds.
```
Ogglas, line 64: See [Update known-issues.md](https://github.com/dotnet/core/pull/7545/commits/b71bdada10f4fc4b2690f8de23ceda9b39e9006c)

HaoK (approved):

PR #7517, opened 2022-06-04
Switch to Standard Term Support (STS) releases

richlander:
The "Current" term is switching to "Short Term Support (STS)". It nicely and intuitively contrasts with "Long Term Support (LTS)". Nothing else is changing (beyond the name).

Relates to:

- https://github.com/dotnet/designs/blob/main/accepted/2022/dotnet-release-labels.md
- https://github.com/dotnet/core/pull/7499

PR #7602, opened 2022-07-15
Update Current -> STS

richlander:


richlander:
I just realized that this is a dupe of #7517.

PR #7632, opened 2022-07-25
Removing x64 from iOS and Android support list - Update supported-os.md 

rbhanda:


dkurepa:
I think this should be updated for 6.0 too

richlander:
What's the thinking on this? I thought that x64 was for the simulator scenario. Is that incorrect?

leecow:
Perhaps (following Ilya's suggestion) update the row to read `x64 (simulated), Arm32, Arm64`?

richlander:
Sure, although isn't the simulator the only place you can use x64 anyway? Put another way, is this adding information solving an actual problem?

akoeplinger:
This is not correct, there are Android x64 and x86 devices out there (though they aren't widespread). You can't assume that x64 means emulator on Android.

For iOS the situation is a little different, there was only ever the iOS Simulator on x64.

release-notes/7.0/supported-os.md
```diff
@@ -86,15 +86,15 @@ OS                            | Version                   | Architectures     |
 
 OS                            | Version                 | Architectures     |
 ------------------------------|-------------------------|-------------------|
-[Android][Android]            | API 21+                 | x64, Arm32, Arm64 |
+[Android][Android]            | API 21+                 | Arm32, Arm64 |
```
akoeplinger, line 89:
```suggestion
[Android][Android]            | API 21+                 | x86, x64, Arm32, Arm64 |
```

release-notes/7.0/supported-os.md
```diff
@@ -86,15 +86,15 @@ OS                            | Version                   | Architectures     |
...
 
 [Android]: https://support.google.com/android
 
 ## iOS / tvOS
 
 OS                            | Version                 | Architectures     |
 ------------------------------|-------------------------|-------------------|
-[iOS][iOS]                    | 10.0+                   | x64, Arm32, Arm64 |
+[iOS][iOS]                    | 10.0+                   |  Arm32, Arm64 |
```
akoeplinger, line 97:
```suggestion
[iOS][iOS]                    | 10.0+                   |  x64, Arm32, Arm64 |
```

axljoker03 (approved):

akoeplinger:
Closing this as it is not needed/incorrect.

PR #7648, opened 2022-07-29
Add rollback file details

marcpopMSFT:
I need to confirm these details with the Maui team first.


rbhanda:
@marcpopMSFT Can we merge this PR?

marcpopMSFT:
@rbhanda we can now. I had to fix the rollback file details as maui shipped an additional patch after the initial release to fix some version numbers. I've tested the current values, they worked, and they appear to match what's on nuget.org so I think we're good to go.

PR #7797, opened 2022-09-14
Fix links in root readme

mderriey:
I noticed that both README files in the root directory and the `release-notes` directory have their content somewhat synced.

Some links in the root one are wrong, which this PR aims at addressing.


dcwhittaker:
Hi @mderriey - thanks for noting this, I fixed them yesterday but unsure what happened. They're showing up correctly now so I'm going to close this PR. If you find anything else amiss, let me know!

PR #7968, opened 2022-11-16
Update supported-os.md

leecow:


release-notes/7.0/supported-os.md
```diff
@@ -15,8 +15,6 @@ OS                                    | Version                 | Architectures
 [Windows Server Core][Windows-Server] | 2012+                | x64, x86        | [Windows Server][Windows-Server-lifecycle]
 [Nano Server][Nano-Server]            | Version 1809+           | x64             | [Windows Server][Windows-Server-lifecycle]
 
-**\*** Windows 7 SP1 is supported with [Extended Security Updates](https://docs.microsoft.com/troubleshoot/windows-client/windows-7-eos-faq/windows-7-extended-security-updates-faq) installed.
```
jamshedd, line 18: This removal is fine, but in addition we need to remove the Windows Client row from the table above, since both Win7 and Win8.1 are no longer supported.

jamshedd (changes requested):

PR #7972, opened 2022-11-17
Fix typo.

teo-tsirpanis:
The asterisk note about Windows 7 ESUs has been removed.


mairaw (approved):
Good catch. Thanks @teo-tsirpanis!

PR #8036, opened 2022-12-13
Fix version README links

vcsjones:
The links in the Version column of the README do not currently work. This fixes the links to point to, what I think, are the correct READMEs.


mairaw (approved):
Thanks @vcsjones!

PR #8180, opened 2023-02-08
Update glibc compat for Linux arm64

sbomer:
The update reflects the change made in https://github.com/dotnet/runtime/pull/80866.


richlander:
Thanks @sbomer!

PR #8240, opened 2023-02-27
Remove obsolete info about PowerShell version in Docker images

mthalman:


mthalman:
@rbhanda - Can you make sure this gets applied to the changes for the 7.0.4 release?

PR #8267, opened 2023-03-09
Add CA2009 regression into the known issues

buyaa-n:
There is a regression in CA2009 analyzer introduced in preview 1 that needs to be added in known issues

CC @jeffhandley

### DIFF
--- a/release-notes/8.0/known-issues.md
+++ b/release-notes/8.0/known-issues.md
@@ -4,14 +4,33 @@ You may encounter the following known issues, which may include workarounds, mit
 
 ## .NET SDK
 
-### [8.0.100-preview.1.23115.2] analyzer CA2009 throws InvalidCastException at runtime causing build failure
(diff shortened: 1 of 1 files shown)


jeffhandley (changes requested):
Please add context that shows under what conditions this will happen for a customer. Is there a code snippet that can illustrate what the analyzer is trying to process, where such a code snippet would trigger this, or does this happen on every project? How does the build failure actually surface? Does it show up as a _build error_ such that the project doesn't build successfully?

buyaa-n:
> Please add context that shows under what conditions this will happen for a customer. Is there a code snippet that can illustrate what the analyzer is trying to process, where such a code snippet would trigger this, or does this happen on every project? How does the build failure actually surface? Does it show up as a _build error_ such that the project doesn't build successfully?

Done, how it looks now?

PR #8305, opened 2023-03-18
Add support statement about old target frameworks

richlander:


krwq (approved):

release-policies.md
```diff
@@ -47,6 +47,8 @@ Patches are published at the [.NET Website](https://dotnet.microsoft.com/downloa
 
 As the end of support nears for a given .NET version, we strongly recommend you move to a newer, supported version of .NET. .NET releases that have reached end of support do not get security patches. Continuing to use an unsupported version will expose you to security vulnerabilities.
 
+End of support also applies to source that targets unsupported .NET versions. You must target a supported .NET version (via the `TargetFramework` property), for both apps and libraries. The .NET SDK produces warnings when you target out of support versions to help identify these cases. The SDK does not prevent you from targeting unsupported versions, however, your configuration will be considered unsupported.
```
jaredpar, line 50: >  You must target a supported .NET version (via the `TargetFramework` property), for both apps and libraries

This is incorrect. We specifically support targeting old TF in libraries.

terrajobst: > You must target a supported .NET version (via the `TargetFramework` property), for both apps and libraries. 

Just to add to that, we have specifically given 1P partner guidance to no waste their time upgrading .NET Framework/.NET Core version in library projects and told them that upgrading application/services/unit tests is what they should focus on.

Also, it just doesn't make sense to say "all versions of .NET Standard are supported to target in perpetuity" and simultaneously say that targeting the frameworks that implement them from the same projects is unsupported.

terrajobst (changes requested):
I feel the same was @jaredpar.

We should scope it to apps, not libraries.

jkotas:
> it just doesn't make sense to say "all versions of .NET Standard are supported to target in perpetuity" and simultaneously say that targeting the frameworks that implement them from the same projects is unsupported.

It may not make sense to you, but it is where we are today given that we are including source generators as part of netcoreapp targeting pack.

jkotas:
> we have specifically given 1P partner guidance to no waste their time upgrading .NET Framework/.NET Core versions in library projects and told them that upgrading application/services/unit tests is what they should focus on.

Yes, our guidance is to focus on upgrading the applications first. If we have guidance that targeting old .NET Core versions is supported forever for libraries, we either need to fix the guidance, or fix how we compose and service the product to be compatible with that guidance.

jaredpar (approved):

jkotas (approved):

PR #8386, opened 2023-04-11
Update supported-os.md

leecow:
A few more refinements to the supported CentOS and RHEL versions.

### DIFF
--- a/release-notes/7.0/supported-os.md
+++ b/release-notes/7.0/supported-os.md
@@ -28,13 +28,13 @@ OS                                    | Version                 | Architectures 
 OS                                    | Version               | Architectures     | Lifecycle
 --------------------------------------|-----------------------|-------------------|----------
 [Alpine Linux][Alpine]                | 3.15+                 | x64, Arm64, Arm32 | [Alpine][Alpine-lifecycle]
-[CentOS Linux][CentOS]                | 7                     | x64               | [CentOS][CentOS-lifecycle]
-[CentOS Stream Linux][CentOS]         | 8                     | x64               | [CentOS][CentOS-lifecycle]
+[CentOS][CentOS].                     | 7                     | x64               | [CentOS][CentOS-lifecycle]
+[CentOS Stream][CentOS]               | 8, 9                  | x64               | [CentOS][CentOS-lifecycle]
 [Debian][Debian]                      | 10+                   | x64, Arm64, Arm32 | [Debian][Debian-lifecycle]
 [Fedora][Fedora]                      | 36+                   | x64               | [Fedora][Fedora-lifecycle]
 [openSUSE][OpenSUSE]                  | 15+                   | x64               | [OpenSUSE][OpenSUSE-lifecycle]
-[Oracle Linux][Oracle-Linux]          | 7+                    | x64               | [Oracle][Oracle-lifecycle]
-[Red Hat Enterprise Linux][RHEL]      | 7+                    | x64, Arm64        | [Red Hat][RHEL-lifecycle]
+[Oracle Linux][Oracle-Linux]          | 8+                    | x64               | [Oracle][Oracle-lifecycle]
+[Red Hat Enterprise Linux][RHEL]      | 8+                    | x64, Arm64        | [Red Hat][RHEL-lifecycle]
 [SUSE Enterprise Linux (SLES)][SLES]  | 12 SP2+               | x64               | [SUSE][SLES-lifecycle]
 [Ubuntu][Ubuntu]                      | 18.04+                | x64, Arm64, Arm32 | [Ubuntu][Ubuntu-lifecycle]
 
--- a/release-notes/7.0/supported-os.md
+++ b/release-notes/7.0/supported-os.md
@@ -28,7 +28,7 @@ OS                                    | Version                 | Architectures 
 OS                                    | Version               | Architectures     | Lifecycle
 --------------------------------------|-----------------------|-------------------|----------
 [Alpine Linux][Alpine]                | 3.15+                 | x64, Arm64, Arm32 | [Alpine][Alpine-lifecycle]
-[CentOS][CentOS].                     | 7                     | x64               | [CentOS][CentOS-lifecycle]
+[CentOS][CentOS]                      | 7                     | x64               | [CentOS][CentOS-lifecycle]
 [CentOS Stream][CentOS]               | 8, 9                  | x64               | [CentOS][CentOS-lifecycle]
 [Debian][Debian]                      | 10+                   | x64, Arm64, Arm32 | [Debian][Debian-lifecycle]
 [Fedora][Fedora]                      | 36+                   | x64               | [Fedora][Fedora-lifecycle]
--- a/release-notes/7.0/supported-os.md
+++ b/release-notes/7.0/supported-os.md
@@ -29,7 +29,7 @@ OS                                    | Version               | Architectures   
 --------------------------------------|-----------------------|-------------------|----------
 [Alpine Linux][Alpine]                | 3.15+                 | x64, Arm64, Arm32 | [Alpine][Alpine-lifecycle]
 [CentOS][CentOS]                      | 7                     | x64               | [CentOS][CentOS-lifecycle]
-[CentOS Stream][CentOS]               | 8, 9                  | x64               | [CentOS][CentOS-lifecycle]
+[CentOS Stream][CentOS]               | 8+                    | x64               | [CentOS][CentOS-lifecycle]
 [Debian][Debian]                      | 10+                   | x64, Arm64, Arm32 | [Debian][Debian-lifecycle]
 [Fedora][Fedora]                      | 36+                   | x64               | [Fedora][Fedora-lifecycle]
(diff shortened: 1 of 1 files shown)


release-notes/7.0/supported-os.md
```diff
@@ -28,13 +28,13 @@ OS                                    | Version                 | Architectures
 OS                                    | Version               | Architectures     | Lifecycle
 --------------------------------------|-----------------------|-------------------|----------
 [Alpine Linux][Alpine]                | 3.15+                 | x64, Arm64, Arm32 | [Alpine][Alpine-lifecycle]
-[CentOS Linux][CentOS]                | 7                     | x64               | [CentOS][CentOS-lifecycle]
-[CentOS Stream Linux][CentOS]         | 8                     | x64               | [CentOS][CentOS-lifecycle]
+[CentOS][CentOS].                     | 7                     | x64               | [CentOS][CentOS-lifecycle]
```
mairaw, line 31: You probably don't need this period
```suggestion
[CentOS][CentOS]                      | 7                     | x64               | [CentOS][CentOS-lifecycle]
```

release-notes/7.0/supported-os.md
```diff
@@ -28,13 +28,13 @@ OS                                    | Version                 | Architectures
...
+[CentOS Stream][CentOS]               | 8, 9                  | x64               | [CentOS][CentOS-lifecycle]
```
mairaw, line 32: can't you say 8+ here?

```suggestion
[CentOS Stream][CentOS]               | 8+                    | x64               | [CentOS][CentOS-lifecycle]
```

release-notes/7.0/supported-os.md
```diff
@@ -28,13 +28,13 @@ OS                                    | Version                 | Architectures
...
-[CentOS Stream Linux][CentOS]         | 8                     | x64               | [CentOS][CentOS-lifecycle]
+[CentOS][CentOS]                      | 7                     | x64               | [CentOS][CentOS-lifecycle]
+[CentOS Stream][CentOS]               | 8+                    | x64               | [CentOS][CentOS-lifecycle]
 [Debian][Debian]                      | 10+                   | x64, Arm64, Arm32 | [Debian][Debian-lifecycle]
 [Fedora][Fedora]                      | 36+                   | x64               | [Fedora][Fedora-lifecycle]
 [openSUSE][OpenSUSE]                  | 15+                   | x64               | [OpenSUSE][OpenSUSE-lifecycle]
-[Oracle Linux][Oracle-Linux]          | 7+                    | x64               | [Oracle][Oracle-lifecycle]
-[Red Hat Enterprise Linux][RHEL]      | 7+                    | x64, Arm64        | [Red Hat][RHEL-lifecycle]
+[Oracle Linux][Oracle-Linux]          | 8+                    | x64               | [Oracle][Oracle-lifecycle]
+[Red Hat Enterprise Linux][RHEL]      | 8+                    | x64, Arm64        | [Red Hat][RHEL-lifecycle]
```
omajid, line 37: I think this is a scenario where Red Hat's support and Microsoft's support differ. AFAIK, Microsoft supports .NET 7 on RHEL 7 and later. Red Hat supports .NET 7 on RHEL 8 and later only. Not sure which one this table is supposed to reflect, but "8+" is the intersection of those policies :)

Falco20019: According to https://learn.microsoft.com/en-us/dotnet/core/install/linux-rhel#supported-distributions RHEL 7 is only supported in .NET 6, not in .NET 7. I would therefore go with the table from there. So I don't think there's a difference in what Microsoft supports and what RHEL supports here.

.NET 6: RHEL 7+
.NET 7: RHEL 8+

Since RHEL 7 is EOL in 2026 (way after .NET 6 is EOL), it will stay 7+ until .NET 6 is gone and the 6.0 document will not be adjusted.

Falco20019: This is an aggregation of the supported distributions together with their next upcoming EOLs that we maintain internally, based on the Microsoft Learn pages and the according `supported-os` pages.
image
Darker red means there was a change between .NET 6 and .NET 7 support, blue means EOL but according to the documentation still supported (with tickets on https://github.com/dotnet/core/labels/os-support existing).

release-notes/7.0/supported-os.md
```diff
@@ -28,13 +28,13 @@ OS                                    | Version                 | Architectures
 OS                                    | Version               | Architectures     | Lifecycle
 --------------------------------------|-----------------------|-------------------|----------
 [Alpine Linux][Alpine]                | 3.15+                 | x64, Arm64, Arm32 | [Alpine][Alpine-lifecycle]
-[CentOS Linux][CentOS]                | 7                     | x64               | [CentOS][CentOS-lifecycle]
-[CentOS Stream Linux][CentOS]         | 8                     | x64               | [CentOS][CentOS-lifecycle]
+[CentOS][CentOS]                      | 7                     | x64               | [CentOS][CentOS-lifecycle]
```
Falco20019, line 31: This one is really called `CentOS Linux`, not just `CentOS`. See https://www.centos.org/cl-vs-cs/

So please keep the name upfront `CentOS`. The lifecycle column can be `CentOS`, since it's referring to the project and is appliable to both versions.

release-notes/7.0/supported-os.md
```diff
@@ -28,13 +28,13 @@ OS                                    | Version                 | Architectures
...
+[CentOS Stream][CentOS]               | 8+                    | x64               | [CentOS][CentOS-lifecycle]
 [Debian][Debian]                      | 10+                   | x64, Arm64, Arm32 | [Debian][Debian-lifecycle]
 [Fedora][Fedora]                      | 36+                   | x64               | [Fedora][Fedora-lifecycle]
 [openSUSE][OpenSUSE]                  | 15+                   | x64               | [OpenSUSE][OpenSUSE-lifecycle]
-[Oracle Linux][Oracle-Linux]          | 7+                    | x64               | [Oracle][Oracle-lifecycle]
-[Red Hat Enterprise Linux][RHEL]      | 7+                    | x64, Arm64        | [Red Hat][RHEL-lifecycle]
+[Oracle Linux][Oracle-Linux]          | 8+                    | x64               | [Oracle][Oracle-lifecycle]
```
Falco20019, line 36: Is this really 8+ as well? Oracle Linux 7 is only EOL in July 2024. So I would expect it to be supported on .NET 6 and 7.

https://learn.microsoft.com/en-us/dotnet/core/install/linux is sadly missing a page for Oracle Linux, so I can't check that.

Falco20019 (changes requested):

release-notes/7.0/supported-os.md
```diff
@@ -28,13 +28,13 @@ OS                                    | Version                 | Architectures
...
+[Red Hat Enterprise Linux][RHEL]      | 8+                    | x64, Arm64        | [Red Hat][RHEL-lifecycle]
 [SUSE Enterprise Linux (SLES)][SLES]  | 12 SP2+               | x64               | [SUSE][SLES-lifecycle]
```
Falco20019, line 38: Not fully related but just out of curiosity:
`12 SP2+` in this case really only means `12 SP2+` & `15`.

Here are the support times for SLES:
image
image

So I would expect the following as of now: `12 SP4+, 15 SP1+`

Falco20019: We should also try to reflect changes for version that haven't been only included by fault in the `Out of support OS versions` section. So for .NET 6 and .NET 7 this would include SLES 12 SP2, SLES 12 SP3 and SLES 15 (unpatched).

release-notes/7.0/supported-os.md
```diff
@@ -28,13 +28,13 @@ OS                                    | Version                 | Architectures
 OS                                    | Version               | Architectures     | Lifecycle
 --------------------------------------|-----------------------|-------------------|----------
 [Alpine Linux][Alpine]                | 3.15+                 | x64, Arm64, Arm32 | [Alpine][Alpine-lifecycle]
-[CentOS Linux][CentOS]                | 7                     | x64               | [CentOS][CentOS-lifecycle]
-[CentOS Stream Linux][CentOS]         | 8                     | x64               | [CentOS][CentOS-lifecycle]
+[CentOS][CentOS]                      | 7                     | x64               | [CentOS][CentOS-lifecycle]
+[CentOS Stream][CentOS]               | 8+                    | x64               | [CentOS][CentOS-lifecycle]
 [Debian][Debian]                      | 10+                   | x64, Arm64, Arm32 | [Debian][Debian-lifecycle]
 [Fedora][Fedora]                      | 36+                   | x64               | [Fedora][Fedora-lifecycle]
 [openSUSE][OpenSUSE]                  | 15+                   | x64               | [OpenSUSE][OpenSUSE-lifecycle]
```
Falco20019, line 35: openSUSE lists 15+ but technically it's 15.4+ which isn't EOL. Not sure if this is intentional to signalize Microsoft support while OS is EOL.

leecow: We try not to chase minor versions as much as possible but it may make sense in this case.

Dezzj (commented):
Veiwed

Falco20019:
Any update on this?

Abdurahmon0412 (approved):

richlander:
We can safely close this PR now.

PR #8417, opened 2023-04-24
Remove .NET Core 3.1 from the roadmap.

teo-tsirpanis:


mairaw (approved):
Good catch! Thanks @teo-tsirpanis

PR #8583, opened 2023-06-28
Fix mention of 8.0 instead of 7.0

gearsetdave:
The release notes for the 8.0 previews say that .NET 8 is the next major release of .NET following .NET 8.0.

This fixes them to say it's the next major release of .NET following .NET 7.0.


mairaw (approved):
Thank you @gearsetdave

Issue #8636, opened 2023-07-20
SUSE Enterprise Linux 12 SP4 reached end-of-life on 30 - June - 2023


Falco20019:
Is it planned to be dropped for .NET 6.0 and .NET 7.0?

SLES 12 SP5 has superseded 12 SP4 and will be supported until October 2027 as oldest version receiving support. All supported-os documents (.NET 6, .NET 7 and .NET 8) currently list `12 SP2+` which makes me wonder, since SP2 was EOL even before .NET 7 was released.

See https://www.suse.com/lifecycle/#suse-linux-enterprise-server-12 for end-of-life information.

Please add os-support label.

/CC @rbhanda @richlander
/awareness @mthalman @adegeo @MichaelSimons

tarekgh:
@leecow do you know the answer to the question raised here?

leecow:
Looks like this was not properly tracked and updated as SLES 12 SP releases went EOL. I think we should update 6, 7, and 8 to SP5+.

PR #8711, opened 2023-08-24
Update supported SLES 12 SP versions

leecow:
Supported SLES SP versions have not been properly tracked. This catches us up. 

Resolve https://github.com/dotnet/core/issues/8636


Userlugard (approved):
don't leave a review

PR #8754, opened 2023-09-12
doc - Fixed servicing releases table not rendering

AptiviCEO:
Hello,

The table didn't render because there was an extra column indicator ("|:------:|") after the header. There were five indicators, when it should have been four according to the header. This is fixed by removing the extra column indicator.

Thanks!


akoeplinger (approved):

akoeplinger:
Thank you!

PR #8872, opened 2023-10-30
Update README.md to reflect correct release date for .net 7

nickdevore:
The release date (year) was wrong for .net 7.  Not critical, but likely important?


mairaw (approved):
Good catch! Thank you @nickdevore!

Dezzj (commented):
Veiwed

PR #9048, opened 2024-01-08
Update min supported macOS version

richlander:


mmitche (approved):

release-notes/8.0/supported-os.md
```diff
@@ -80,7 +80,7 @@ Other distributions are supported at best effort, per [.NET Support and Compatib
 
 OS                            | Version                   | Architectures     |
 ------------------------------|---------------------------|-------------------|
-[macOS][macOS]                | 10.15+                    | x64, Arm64        |
+[macOS][macOS]                | 12.0                    | x64, Arm64        |
```
jkotas, line 83: @richlander Do we need the same update for the macCatalyst version below?

jkotas, line 83: Nit: This should be `12+`. We have plus on all other versions. Apple dropped minor versions in the versioning scheme of their OSes.

richlander:
@jkotas Got catches.

https://github.com/dotnet/core/pull/9048

PR #9119, opened 2024-02-02
Update core-repos.md

jamesmontemagno:


JonDouglas (approved):
Adds a bit of value imo for discussions and issues.

richlander:
The first table is too broad/not helpful. It's a catch-all, which doesn't really benefit anyone.

I think of the breakdown as:

- Platform
- App Stacks
- Tools

I'd put languages in with Platform. Feel free to change the names.

If there is stuff (like `installer`) that doesn't fit well anywhere, then put that in an `Other` bucket.

jamesmontemagno:
@richlander thoughts here... I reorgainzed to: Platform, App Workloads, Libraries, Tools, Documentation

PR #9223, opened 2024-03-13
Remove outdated comment about Powershell 7.3 RC1 from release notes

jnoordsij:
The release notes for various releases have been containing a copied note about Powershell 7.3 RC1 being part of the Docker images, which was only the case for the first released Docker image in the 7.0 series and no longer true ever since. However, this message has been repeated once after and been part of all the 8.0 release notes.

This PR removes this wrong and outdated information for these cases.

See also https://github.com/dotnet/dotnet-docker/pull/4200.


jnoordsij:
@dotnet-policy-service agree company="Stichting Studiebegeleiding Leiden"

PR #9225, opened 2024-03-13
Package support statement

richlander:


jaredpar (approved):

ericstj:
@richlander do you know when we can expect to see this in https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core?

richlander:
That's a great question. That's up to @mairaw.

PR #9253, opened 2024-04-01
8.0.2xx compatibility known issue

marcpopMSFT:


mattzink:
Will VS 17.10 be released soon? It feels strange to require a pre-release version.

PR #9279, opened 2024-04-20
Fix Apple mobile versions in 9.0 supported-os.md

akoeplinger:
https://github.com/dotnet/core/pull/9160 accidentally overwrote my changes from https://github.com/dotnet/core/pull/9143 and https://github.com/dotnet/core/pull/9145


richlander:
Can you look at this @rbhanda? Looks like we might have a bit of a versioning problem to address.

Issue #8880, opened 2023-11-02
Notifications about ending OS support


Falco20019:
I didn't find an issue with the os-support label (or any else regarding this). Support for Alpine 3.15 ended yesterday. Is there already a tracking issue? Is it planned to be dropped for .NET 6.0 and .NET 7.0?

We have an aggregated internal overview that I want to keep updated and for that I would like to track that if possible :)

/CC @rbhanda @richlander

richlander:
You can consider support dropped, however, @rbhanda, do we have some paper work for that?

It is interesting to hear that you have an internal overview of this. Is there something we can do better to help with that? It is always interesting to us to hear about folks that have a structured program for such things. Feel free to respond here or mail us at dotnet@microsoft.com if you want to discuss this further.

rbhanda:
https://github.com/dotnet/core/issues/8888 created for this

Falco20019:
Hi @richlander,

thanks for your interest. We are mostly tracking the upcoming end of support per product and the differences between the version (especially what is dropped). While the `supported-os.md` files would have a `Out of support OS versions` section, those are generally not well maintained. For example, the .NET 7 page shows only Fedora 35 and SUSE LES 12 SP2-4 as out of support. But I know that also Alpine Linux 3.15, Fedora Linux 36, openSUSE Linux 15.0 - 15.3, Ubuntu Linux 18.04 and macOS 10.x are EOS. Upfront, I totally understand that it's not your focus to keep those documents up2date all the time as it involves some work. But I find it confusing that there is a section on the support pages that shows some but not all EoS things.

These are the lists we currently have internally:
![image](https://github.com/dotnet/core/assets/940619/95b753f1-0646-47bc-a110-6d0f3447e37f)
![image](https://github.com/dotnet/core/assets/940619/e456ba35-802e-47e0-9fa1-9b983cd66a86)
![image](https://github.com/dotnet/core/assets/940619/53f11059-c9d9-4703-8b43-d43d672dd28f)
![image](https://github.com/dotnet/core/assets/940619/ca7fa000-6996-4ffb-95d1-cc4eec269062)
You can ignore the light-red on .NET 7 and .NET 8 columns since these only have the internal meaning of "available without change but currently not used by us as we target .NET 6 LTS in our product". The dark-red means, there was a change in support towards the previous version.
[...]

PR #9294, opened 2024-05-01
Add Supported OS JSON files

richlander:
This is an proposal of a new schema for OS Support information. It is based on a collaboration with @Falco20019 with ideas developed in https://github.com/dotnet/core/pull/9210/.

Once we agree, the intent is to merge this PR.

Goals:

- Extend https://github.com/dotnet/core/blob/main/release-notes/releases-index.json to include OS support information, per release.
- Enable using the JSON information anywhere where OS supported is needed (including generating the markdown files, for example https://github.com/dotnet/core/blob/main/release-notes/9.0/supported-os.md
- Make it easy to atomate changes to these files based on data maintained at https://endoflife.date/.

The endoflife site is pretty cool:

- https://endoflife.date/dotnet
- https://endoflife.date/api/dotnet.json


richlander:
FYI: @martincostello

release-notes/9.0/supported-os.json
```diff
@@ -0,0 +1,277 @@
...
+                        "2012+"
+                    ]
+                }
+            ]
+        }
+    ],
+    "linux": {
+        "glibc": [
+            {
+                "architectures": [
```
martincostello, line 252: Is the architecture the key here? What I mean is, is the intent that for a given Linux distribution I use the architecture to find the source/version of glibc for that?

release-notes/9.0/supported-os.json
```diff
@@ -0,0 +1,277 @@
...
+            "distributions": [
+                {
+                    "name": "Android",
+                    "lifecycle-policy": "https://support.google.com/android",
+                    "architectures": [
+                        "x64",
+                        "arm64",
+                        "arm"
+                    ],
+                    "supported-cycles": [
```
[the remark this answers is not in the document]
Falco20019: We tried to keep the terminology close to https://endoflife.date/docs/api

release-notes/9.0/supported-os.json
Falco20019: This mostly informs on what architectures are supported. Since arm32 and x86 are not listed, those are not in support.

release-notes/9.0/supported-os.json
```diff
@@ -0,0 +1,277 @@
...
+            "family": "Windows",
+            "distributions": [
+                {
+                    "name": "Nano Server",
+                    "lifecycle-policy": "https://learn.microsoft.com/windows-server/get-started/windows-server-release-info",
+                    "architectures": [
+                        "x64"
+                    ],
+                    "supported-cycles": [
+                        "Version 1809+"
```
Falco20019, line 197: I'm not a too big fan of using wildcards but would prefer more explicit listing like in the other cases (i.e. Ubuntu). This would allow people to easier see when one is dropped.

MichaelSimons:
Looping in @joeloff as the [Microsoft.Deployment.DotNet.Releases library](https://github.com/dotnet/deployment-tools/tree/main/src/Microsoft.Deployment.DotNet.Releases) is going to need to react.

release-notes/9.0/supported-os.json
richlander: Thats in place so that this section could be generated: https://github.com/dotnet/core/blob/main/release-notes/9.0/supported-os.md#libc-compatibility

The overall doc/JSON tells you which versions that Microsoft chooses to support. The libc information could be used by a tool to provide a decent educated guess on OS versions where .NET will likely work but isn't supported by Microsoft.

release-notes/9.0/supported-os.json
richlander: I agree. I got lazy at the end. I'll try to expand these.

richlander:
Having @joeloff react could be good. I assume there is no actual pressing reason since this information is purely additive not breaking.

richlander:
I removed most of the wildcards. Two outstanding issues:

- I don't know what to include for Mac Catalyst.
- I'm sure that the Windows versions are correct.

Falco20019:
I will give it a look later today and try to fill the holes if possible :)

release-notes/9.0/supported-os.json
```diff
@@ -0,0 +1,304 @@
...
+            "family": "Apple",
+            "distributions": [
+                {
+                    "name": "iOS",
+                    "lifecycle-policy": "https://support.apple.com/ios/",
+                    "architectures": [
+                        "arm64"
+                    ],
+                    "supported-cycles": [
+                        "12.2"
```
Falco20019, line 32: I would expect here the versions of 15+. 12.2 through 14.5 are already EOL. Maybe this is also something like a min-sdk version? I haven't used iOS builds myself so far. At least that seams to correspond to the `SupportedOSPlatformVersion`.

Falco20019 (commented):
We might want to split the `supported-cycles` up. Because there is the argument that Microsoft only gives technical support for OS that are maintained and not EOL and there is also the versions supported by the compile tool chain (`SupportedOSPlatformVersion`). These are different and both are of interest.

release-notes/9.0/supported-os.json
```diff
@@ -0,0 +1,304 @@
...
+        {
+            "family": "Android",
+            "distributions": [
+                {
+                    "name": "Android",
+                    "lifecycle-policy": "https://support.google.com/android",
+                    "architectures": [
+                        "x64",
+                        "arm64",
+                        "arm"
```
akoeplinger, line 14: x86 is also supported on Android

release-notes/9.0/supported-os.json
```diff
@@ -0,0 +1,304 @@
...
+                    ],
+                    "supported-cycles": [
+                        "API 21+"
```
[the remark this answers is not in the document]
akoeplinger: While what you wrote is correct we have been using the API version instead of the Android OS version pretty much everywhere else, and it it's also what a user would set for `SupportedOSPlatformVersion` in the project file.

I'd prefer if we use API version for consistency.

release-notes/9.0/supported-os.json
akoeplinger: No, we still support iOS 12.2 as the minimum target an app can run on. It is similar to min-sdk on Android.

release-notes/9.0/supported-os.json
```diff
@@ -0,0 +1,304 @@
...
+                },
+                {
+                    "name": "Mac Catalyst",
+                    "lifecycle-policy": "https://support.apple.com/macos/",
+                    "architectures": [
+                        "x64",
+                        "arm64"
+                    ],
+                    "supported-cycles": [
+                        "15+"
```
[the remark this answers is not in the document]
akoeplinger: We can't really sync with iOS here since we support earlier iOS versions than what Mac Catalyst supports, using `15.0+` is correct here.

release-notes/9.0/supported-os.json
```diff
@@ -0,0 +1,304 @@
...
+                        "x64",
+                        "arm64"
+                    ],
+                    "supported-cycles": [
+                        "12.2",
+                        "13",
+                        "14",
+                        "15",
+                        "16",
+                        "17"
```
akoeplinger, line 64: I'm a bit confused why we use `X.Y+` for one platform and explicitly list versions for other platforms.

In the current supported-os.md document we only specify the minimum platform version that we support running on.

release-notes/9.0/supported-os.json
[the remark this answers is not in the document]
Falco20019: My main issue is what I wrote here: https://github.com/dotnet/core/pull/9294#pullrequestreview-2037531038

The current `supported-os.md` is mixing up `SupportedOSPlatformVersion` and `supported-cycles`. `SupportedOSPlatformVersion` for Windows is not even mentioned anywhere.

For me as an end-user it's pretty hard to keep everything locally in sync because our end-customers are not interested in our `SupportedOSPlatformVersion` but about what OS we support them on. And I am interested on that as well from the perspective for what OS I get support from Microsoft if I stumble into issues. The documentation is currently only covering half of that and also not very consistent.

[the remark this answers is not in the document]
akoeplinger: Yes I agree with that. We've been struggling to clearly define/separate
1. the minimum version an app will run on constrained by technical means i.e. min-sdk, min libc etc.
2. the minimum version we "officially" support which excludes EOL versions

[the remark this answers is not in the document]
Falco20019: https://github.com/dotnet/core/issues/8880 can give you a better picture about where this thing started of from.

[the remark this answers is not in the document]
Falco20019: In my opinion, the information about OS EOLs should be maintained separatedly and since the JSON files are generated according to @leecow should be taken into account when generating these to not manually maintain them in multiple locations. Otherwise the `supported-cycles` would need to be kept in sync for all .NET channels on each EOL.

release-notes/9.0/supported-os.json
Falco20019: Not according to https://github.com/dotnet/core/blob/main/release-notes/9.0/supported-os.md#android

release-notes/9.0/supported-os.json
Falco20019: Mainly because of the mixup of min-sdk and supported-cycles so far. supported-cycles will change during the lifetime of a .NET version due to EOL, min-sdk will usually not change anymore and is also only one version not a full list.

release-notes/9.0/supported-os.json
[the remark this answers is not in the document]
Falco20019: I also think this is mostly because of min-sdk and supported-cycles being mixed up. The min-sdk will be separate of iOS, the supported-cycles most probably not.

release-notes/9.0/supported-os.json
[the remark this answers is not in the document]
richlander: The most important bit of data is supported OS, which aligns with "what we test and support". We should make sure that's captured. We've run into confusion in the past when we didn't have clarity on this.

What we target (with the compiler) is also important.

release-notes/9.0/supported-os.json
```diff
@@ -0,0 +1,304 @@
...
+            {
+                "architectures": [
+                    "arm64",
+                    "x64"
+                ],
+                "version": "2.23",
+                "source": "Ubuntu 16.04"
+            },
+            {
+                "architecture": "arm32",
```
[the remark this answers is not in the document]
richlander: Done

release-notes/9.0/supported-os.json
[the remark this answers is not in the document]
richlander: It seems like https://github.com/dotnet/core/pull/9294#discussion_r1588869286 is a compromise that gives everyone what they want. I'm not seeing an obvious downside to this approach. Yes/No?

release-notes/9.0/supported-os.json
```diff
@@ -14,84 +15,44 @@
...
-            "family": "Apple",
-            "distributions": [
+            "name": "Apple",
+            "operating-system": [
                 {
+                    "id": "ios",
                     "name": "iOS",
                     "lifecycle-policy": "https://support.apple.com/ios/",
                     "architectures": [
                         "arm64"
```
Falco20019, line 33: There are differences in supported architectures for iOS and the Simulator according to https://github.com/dotnet/core/blob/b488f858df2803b353233f38b16786913d6bcbc7/release-notes/9.0/supported-os.md#ios--tvos--maccatalyst which will be lost by your change.

release-notes/9.0/supported-os.json
```diff
@@ -202,78 +170,43 @@
             ]
         },
         {
-            "family": "Windows",
-            "distributions": [
-                {
-                    "name": "Nano Server",
```
Falco20019: Where do you intend to add the information about Nano Server and Windows Server Core? They have different support information, so I'm not sure if we can just refer to any of the remaining ones.

release-notes/9.0/supported-os.json
```diff
@@ -193,16 +185,15 @@
...
                     ]
                 }
             ]
         },
         {
-            "family": "Windows",
+            "name": "Windows",
             "distributions": [
                 {
                     "name": "Nano Server",
```
Falco20019, line 199: `Windows Server Core` has the `id` set to `windows-server-core` but `Nano Server` has no `id` at all right now. I will ask the people of `endoflife.date` if there is intention to add those as well.

Falco20019: I try to add `windows-nano-server` through https://github.com/endoflife-date/endoflife.date/pull/5323. So we should use that.

richlander:
We have libc documented here. Would it make sense to document libstdc++, too?

@sbomer @jkotas

richlander:
I now have two good PoCs for using this schema. It feels pretty ergonomic and is trivial to use in conjunction with endoflife.date.

- https://github.com/dotnet/core/pull/9365/
- https://github.com/richlander/distroessed/

richlander:
After working on this a bit, I came to the following philosophy (related to the feedback).

This is document is oriented on:

- OSes not targets.
- OSes not apps.
- Straightforward interop with Endoflife.date.
- Ergonomic reading/processing.

I think the current schema achieves that pretty well.

jkotas:
> We have libc documented here. Would it make sense to document libstdc++, too?
> @sbomer @jkotas

I think documenting libc version is good enough. Versions of other libraries are effectively implied by libc version in the Linux ecosystem. Both our scheme and [manylinux](https://github.com/pypa/manylinux) depend on this simplifying assumption.

We depend on number of libraries that come with the OS. Documentation version of all these libraries would be complicated.

release-notes/9.0/supported-os.json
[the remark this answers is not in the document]
jkotas: Do we believe that it is a good name? I find this name very odd. We do not have stick with bad names. This is not 1:1 mapping with that APIs.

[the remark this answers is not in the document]
Falco20019: I found release cycle fine, but have no hard feelings about using versions instead.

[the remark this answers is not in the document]
jkotas: `supported-release-cycles` would be fine. The adjective is important to communicate the meaning.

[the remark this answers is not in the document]
jkotas: > `supported-release-cycles` would be fine.

But it is still not ideal. Release cycle is a repeatable process that produces releases. For example, take a look at how Ubuntu describes its current release cycle: https://ubuntu.com/about/release-cycle. This is not a list of the process variations. This is a list of what the process produced. So `supported-releases` or `supported-versions` would be even better.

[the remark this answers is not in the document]
richlander: `cycles` was an idea. I'm happy to switch to something more conventional.

release-notes/8.0/supported-os.json
```diff
@@ -0,0 +1,345 @@
...
+                        "Arm64",
+                        "x64"
+                    ],
+                    "supported-versions": [
+                        "24.04",
+                        "22.04",
+                        "20.04"
+                    ],
+                    "unsupported-versions": [
+                        "23.10"
```
Falco20019, line 222: This should be `23.04`. `23.10` is missing in the `supported-versions` section (still supported for 2-3 more weeks).

release-notes/8.0/supported-os.json
```diff
@@ -0,0 +1,345 @@
...
+                        "Mac Catalyst apps are supported on macOS Arm64 and x64."
+                    ]
+                },
+                {
+                    "id": "tvos",
+                    "name": "tvOS",
+                    "link": "https://developer.apple.com/tvos/",
+                    "lifecycle": "https://support.apple.com/apple-tv/",
+                    "architectures": [
+                        "Arm64",
```
Falco20019, line 93: According to https://github.com/dotnet/core/blob/v8.0.0/release-notes/8.0/supported-os.md tvOS is only supported on `Arm64`. `x64` support is only valid for the simulator (already in the notes for `macOS`).

release-notes/8.0/supported-os.json
```diff
@@ -0,0 +1,345 @@
...
+                    "link": "https://www.debian.org/",
+                    "lifecycle": "https://wiki.debian.org/DebianReleases",
+                    "architectures": [
+                        "Arm32",
+                        "Arm64",
+                        "x64"
+                    ],
+                    "supported-versions": [
+                        "12"
+                    ]
```
Falco20019, line 142: Debian is missing the `unsupported-versions` for `11` according to https://github.com/dotnet/core/blob/v8.0.0/release-notes/8.0/supported-os.md

release-notes/8.0/supported-os.json
```diff
@@ -0,0 +1,345 @@
...
+                        "Arm64",
+                        "x64"
+                    ],
+                    "supported-versions": [
+                        "40",
+                        "39"
+                    ],
+                    "unsupported-versions": [
+                        "38"
+                    ]
```
Falco20019, line 160: Fedora is mising the `unsupported-versions` entry for `37` according to https://github.com/dotnet/core/blob/v8.0.0/release-notes/8.0/supported-os.md

release-notes/8.0/supported-os.json
```diff
@@ -0,0 +1,345 @@
...
+                        "12"
+                    ]
+                },
+                {
+                    "id": "fedora",
+                    "name": "Fedora",
+                    "link": "https://fedoraproject.org/",
+                    "lifecycle": "https://fedoraproject.org/wiki/End_of_life",
+                    "architectures": [
+                        "Arm32",
```
Falco20019, line 150: According to https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md Fedora is only supported on `x64`.

release-notes/8.0/supported-os.json
```diff
@@ -0,0 +1,345 @@
...
+                        "38"
+                    ]
+                },
+                {
+                    "id": "opensuse",
+                    "name": "openSUSE Leap",
+                    "link": "https://www.opensuse.org/",
+                    "lifecycle": "https://en.opensuse.org/Lifetime",
+                    "architectures": [
+                        "Arm64",
```
Falco20019, line 168: According to https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md openSUSE Leap is only supported on `x64`.

release-notes/8.0/supported-os.json
```diff
@@ -0,0 +1,345 @@
...
+                        "Red Hat family distributions are supported per [Linux compatibility and support](../../linux-support.md)."
+                    ]
+                },
+                {
+                    "id": "sles",
+                    "name": "SUSE Enterprise Linux",
+                    "link": "https://www.suse.com/",
+                    "lifecycle": "https://www.suse.com/lifecycle/",
+                    "architectures": [
+                        "Arm64",
```
Falco20019, line 199: According to https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md SLES is only supported on `x64`.

release-notes/8.0/supported-os.json
```diff
@@ -0,0 +1,345 @@
...
+                    "supported-versions": [
+                        "40",
+                        "39"
+                    ],
+                    "unsupported-versions": [
+                        "38"
+                    ]
+                },
+                {
+                    "id": "opensuse",
```
Falco20019, line 163: `Oracle Linux` was dropped. Was this intentional?

release-notes/8.0/supported-os.json
```diff
@@ -0,0 +1,345 @@
...
+                    "id": "sles",
+                    "name": "SUSE Enterprise Linux",
+                    "link": "https://www.suse.com/",
+                    "lifecycle": "https://www.suse.com/lifecycle/",
+                    "architectures": [
+                        "Arm64",
+                        "x64"
+                    ],
+                    "supported-versions": [
+                        "15.5"
```
Falco20019, line 203: SLES still supports `12.5` as well.

Falco20019 (changes requested):
Some remarks on missing or wrong versions.

release-notes/8.0/supported-os.json
Falco20019: Should be 8 + 9 on both .NET 8 & 9. Oracle Linux 7 is still in support was never supported by any of those two .NET versions according to the file history.

release-notes/8.0/supported-os.json
```diff
@@ -0,0 +1,345 @@
+{
+    "channel-version": "9.0",
```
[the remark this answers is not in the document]
richlander: Fixed.

release-notes/8.0/supported-os.json
richlander: Fixed

release-notes/9.0/supported-os.json
```diff
@@ -84,7 +90,7 @@
                         "Arm64",
                         "x64"
```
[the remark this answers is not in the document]
richlander: Fixed.

release-notes/9.0/supported-os.json
```diff
@@ -108,7 +114,7 @@
...
                     ],
-                    "supported-cycles": [
+                    "supported-versions": [
                         "3.20",
                         "3.19",
                         "3.18",
```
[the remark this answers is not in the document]
richlander: Fixed.

release-notes/8.0/supported-os.json
richlander: I added 11 to supported (for another month).

release-notes/8.0/supported-os.json
richlander: Fixed.

release-notes/8.0/supported-os.json
richlander: Yes. Oracle Linux is just one of many RHEL-compatible distros. I added a note to accomodate this.

release-notes/8.0/supported-os.json
richlander: This is more of a historicalism. We should support all the architectures where we have an offering. There is no intent to only support x64 for any of the distros.

release-notes/8.0/supported-os.json
richlander: Same as OpenSuse.

release-notes/8.0/supported-os.json
richlander: Fixed.

release-notes/9.0/supported-os.json
```diff
@@ -139,7 +145,7 @@
                         "Arm64",
```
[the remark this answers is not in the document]
richlander: Same as stated on 8.0 review.

release-notes/9.0/supported-os.json
```diff
@@ -139,7 +145,7 @@
...
                         "x64"
                     ],
-                    "supported-cycles": [
+                    "supported-versions": [
                         "40",
                         "39"
                     ]
```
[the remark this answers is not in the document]
richlander: Same as stated for 8.0.

release-notes/8.0/supported-os.json
richlander: .NET 8+ cannot run on 12.5, due to requiring glibc 2.23.

From 12.5

```
$ ldd --version
ldd (GNU libc) 2.22
```

release-notes/9.0/supported-os.json
```diff
@@ -181,9 +190,8 @@
                         "Arm64",
                         "x64"
                     ],
-                    "supported-cycles": [
-                        "40",
-                        "39"
+                    "supported-versions": [
+                        "15.5"
```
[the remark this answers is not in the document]
richlander: .NET 8+ cannot run on 12.5, due to requiring glibc 2.23.

From 12.5

$ ldd --version  
ldd (GNU libc) 2.22

release-notes/9.0/supported-os.json
```diff
@@ -139,7 +145,7 @@
                         "Arm64",
                         "x64"
                     ],
-                    "supported-cycles": [
+                    "supported-versions": [
                         "40",
                         "39"
```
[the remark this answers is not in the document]
richlander: Fixed.

richlander:
Thanks for the feedback @Falco20019. That was very useful to get us towards more accurate info.

richlander:
Should we add a `supported-os` property at this location: https://github.com/dotnet/core/blob/763e6208a661e158e799c243f23a0a540dbdad01/release-notes/releases-index.json#L14?

If so, what should the path be? A fully-qualified GH path or to the same Azure Storage path? IMO, a GH raw path is likely good enough. The information isn't of the same secure nature as the other information and it changes outside of the patch tuesday publishing schedule.

Thoughts?

Also, I was going to skip doing this exercise for .NET 6. If someone else wants to do that, that's fine with me.

release-notes/8.0/supported-os.json
omajid: This list is about the architectures that .NET is available (supported?) on, right? IMO the list should be:

- arm64
- ppc64le
- s390x
- x64

.NET on Fedora is not available on Arm32 or x86.

release-notes/8.0/supported-os.json
```diff
@@ -0,0 +1,346 @@
...
+                    ]
+                },
+                {
+                    "id": "rhel",
+                    "name": "Red Hat Enterprise Linux",
+                    "link": "https://access.redhat.com/",
+                    "lifecycle": "https://access.redhat.com/support/policy/updates/errata/",
+                    "architectures": [
+                        "Arm64",
+                        "x64"
```
omajid, line 184: We should add s390x and ppc64le here.

richlander: I assume for 8 and 9?

release-notes/9.0/supported-os.json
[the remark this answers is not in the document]
Falco20019: That's interesting, since according to https://github.com/dotnet/core/blob/v8.0.0/release-notes/8.0/supported-os.md it was supported in the past (while the document was already mentioning glibc 2.23).

release-notes/8.0/supported-os.json
omajid: Yes.

release-notes/9.0/supported-os.json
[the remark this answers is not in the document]
richlander: It was a mistake in .NET 8, then. I probably didn't notice it initially since I'm less familiar with Suse ecosystem. This time, I actually checked since we're doing a much more thorough job.

release-notes/releases-index.json
```diff
@@ -12,7 +12,8 @@
             "support-phase": "preview",
             "eol-date": "2026-05-11",
             "release-type" : "sts",
-            "releases.json": "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/9.0/releases.json"
+            "releases.json": "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/9.0/releases.json",
+            "supported-os.json": "https://raw.githubusercontent.com/dotnet/core/main/release-notes/8.0/supported-os.json"
```
leecow, line 16: Richard and I discussed - we'll also sync the supported-os.json to the same storage location as releases.json.

leecow (approved):

release-notes/releases-index.json
richlander: Moved to https://github.com/dotnet/core/pull/9367

richlander:
https://github.com/dotnet/core/discussions/9370

PR #9296, opened 2024-05-04
Clarify Windows license information

jkotas:
- DiaSymReader is used by both runtime and SDK
- ucrtbase is no longer included in .NET 7+

### DIFF
--- a/license-information-windows.md
+++ b/license-information-windows.md
@@ -8,8 +8,8 @@ apply to your use. By using any or all of these files you agree to their associa
 The following binaries are licensed with the
 [Windows SDK License](https://learn.microsoft.com/legal/windows-sdk/license):
 
-* api-ms-\*.\* (used by .NET runtime)
-* ucrtbase.dll (used by .NET runtime)
+* api-ms-\*.\* (used by .NET runtime before .NET 7.0)
+* ucrtbase.dll (used by .NET runtime before .NET 7.0)
 * D3DCompiler_47_cor3.dll (used by WPF)
 
 The following binaries are licensed with the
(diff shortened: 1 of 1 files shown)


license-information-windows.md
```diff
@@ -8,8 +8,8 @@ apply to your use. By using any or all of these files you agree to their associa
 The following binaries are licensed with the
 [Windows SDK License](https://learn.microsoft.com/legal/windows-sdk/license):
 
-* api-ms-\*.\* (used by .NET runtime)
-* ucrtbase.dll (used by .NET runtime)
+* api-ms-\*.\* (used by .NET runtime before .NET 7.0)
+* ucrtbase.dll (used by .NET runtime before .NET 7.0)
```
richlander, line 12:
```suggestion
* api-ms-\*.\* (used by .NET runtime,  .NET 6 and earlier)
* ucrtbase.dll (used by .NET runtime ,  .NET 6 and earlier)
```

I think that's a similar approach.

richlander:
> The Windows distribution of .NET 5.0 and later releases contains files that are provided under multiple licenses. 

Can we change that to: "The Windows distribution of .NET contains files that are provided under multiple licenses."


I think the version bounding provided by ".NET 5.0" is no longer useful.

richlander (approved):

GitMensch:
Does "multiple licenses" only apply to .NET on Windows?

Shouldn't this be primarily covered in the LICENSE file? Note that this has more possible issues - see #662.

jkotas:
> Does "multiple licenses" only apply to .NET on Windows?

There are multiple licenses in play on all OSes:

On Windows, the whole package is licensed as MICROSOFT .NET LIBRARY. Parts of the package are originally licensed under variety of licenses that are compatible with MICROSOFT .NET LIBRARY.

On Linux, the whole package is licensed as MIT. Parts of the package are originally licensed under variety of licenses that are compatible with MIT. (ThirdPartyNotices.txt should have complete list, for the ones that require attribution at minimum.)

The difference between Windows distribution and non-Windows distribution is that Windows distribution includes proprietary non-OSS binaries that have licenses incompatible with MIT.

> Shouldn't this be primarily covered in the LICENSE file?

I do not know. It would be a question for lawyers on whether it would made sense. If it was possible, it would make the LICENSE file more complicated and non-standard that comes with its own set issues that are probably worse than the problem you are trying to solve.

richlander:
The .NET library license is pretty broad. I see this file as providing more insight for people that care. Microsoft is the license holder (obviously). If we believe the .NET library license is insufficient on its own, then we should be motivated to update it in some way. I don't see that happening.

If we were shipping the Windows distribution as MIT, the conversation would be different.

PR #9369, opened 2024-06-26
Update markdown formatting

richlander:


Falco20019:
There is a strange artifact in https://[raw.githubusercontent.com/dotnet/core/8a0d0c9577d9f6d864411e85b16e54d390eacc83/release-notes/7.0/supported-os.md](https://raw.githubusercontent.com/dotnet/core/8a0d0c9577d9f6d864411e85b16e54d390eacc83/release-notes/7.0/supported-os.md)

image

Looks a bit like there would be multi-threaded access on the writer?

jamesmontemagno:
@richlander let me know if those linux links are okay 

https://developer.android.com/about/versions/pie is legit so we are fine to merge if the other updates I did are good

jamesmontemagno (approved):

richlander:
I am going to make a couple more changes.

Falco20019:
@richlander Shouldn‘t the links in the JSON be changed as well? As this is where those originated from.

richlander:
Can you elaborate? I'm not following.

richlander:
I understand now. I missed that @jamesmontemagno changed the markdown. I'll fix that in another one PR. Thanks for noticed that.

Falco20019:
Yes, it seems that you reverted his changes by regenerating. He seemed to have fixed some links. And the Linux support would need to be fixed in the templates before regenerating.

richlander:
The Apple links don't seem better to me. They don't relate to an OS version. What was your thinking on that @jamesmontemagno ?

richlander:
https://github.com/dotnet/core/pull/9371

jamesmontemagno:
> The Apple links don't seem better to me. They don't relate to an OS version. What was your thinking on that @jamesmontemagno ?

https://support.apple.com/ios/ goes to a random article on how to install it
https://support.apple.com/ipados doesn't exist
 
trying to align with the other ones for macos and apple-tv... there isn't a lifecycle though

PR #9371, opened 2024-06-26
Update linux compatibility paragraph

richlander:


richlander:
I removed all the Apple support links.

@jamesmontemagno @Falco20019

jamesmontemagno (approved):

Falco20019:
I just checked https://github.com/endoflife-date/endoflife.date/blob/master/products/ios.md?plain=1 and https://github.com/endoflife-date/endoflife.date/blob/master/products/ipados.md?plain=1 but they also only link to Wikipedia for the policy. Seems like Apple is not publishing any official policy, so removing here is fine for me.

PR #9402, opened 2024-07-10
Update Linux support statements

richlander:
No change in support is intended.

- Simplifying language
- Broadening derivative support to include more than RHEL
- Fix broken links
- Remove CentOS Stream 7. I think Stream started with v8.
- Update the last-updated date
- Include last-updated date in the markdown
- Re-run generator (new lifecycle links came in)

@Falco20019 @omajid @leecow @MichaelSimons @jkotas

### DIFF
--- a/release-notes/6.0/supported-os.json
+++ b/release-notes/6.0/supported-os.json
@@ -170,8 +170,7 @@
                         "9"
                     ],
                     "unsupported-versions": [
-                        "8",
-                        "7"
+                        "8"
                     ]
                 },
                 {
@@ -249,7 +248,7 @@
                         "7"
                     ],
                     "notes": [
-                        "RHEL-compatible distributions are supported per [.NET Support](../../support.md)."
+                        "RHEL-compatible derivatives are supported per [.NET Support](../../support.md)."
                     ]
                 },
                 {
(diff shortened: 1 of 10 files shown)


release-notes/6.0/supported-os.json
```diff
@@ -249,7 +249,7 @@
                         "7"
                     ],
                     "notes": [
-                        "Red Hat family distributions are supported per [Linux compatibility and support](../../linux-support.md)."
+                        "RHEL-compatible distributions are supported per [.NET Support](../../support.md)."
```
leecow, line 252: I don't think I see anything in support.md mentioning "RHEL-compatible" support. Should that broad of a statement be clearly defined?

richlander: Great point. This is what I intend folks to look at.

https://github.com/dotnet/core/pull/9402/files#diff-017fcd31229e5b6d295ec48a4746aa15fb14eaf39ed3213ebb0e10436c73a90cR19

Three ideas:

- Use the term "RHEL-compatible derivatives" instead of "RHEL-compatible distributions"
- Use the term "Compatible derivatives" instead of "Downstream derivative support"
- Do both.

richlander: Made those changes.

richlander:
I need an approval for this change. Thanks.

omajid (approved):
No objections from me. The CentOS Stream 7 changes look correct: 7 was never a thing.

richlander:
Thanks @omajid!

PR #9412, opened 2024-07-15
Linux packages JSON format

richlander:
We have a packages file in markdown format currently. The JSON file is intended to be the source of truth and to be used to generate the markdown file (proposal in place for both formats).

I made a few design choices:

- Separate logical from physical packages.
- Document all the interesting notes (including the scenarios) on the logical packages, keeping the physical package info as lean (and duplication-free) as possible.
- Order the packages such that it is likely/possible to be able to foreach+install (like curl + jq + bash) required packages for both NAOT or CoreCLR w/o installing anything extra. This fits a bit into the "clever" category.

The intent is that any distro can be added here. NuGet packages are missing. We can add those, too.

I started with .NET 9 to prove out the idea. Adding the other .NET versions is trivial since the packages are all the same. The only difference is that the distro versions may fully, partially or not at all overlap. It's another example of a historical document.

I named this `linux-packages.json` to match `linux-packages.md`. It would be good to add platforms like FreeBSD here, too. I'm wondering if now is a good time to rename these files as `required-packages.json` or `runtime-deps.json` or similar.

Note: Here is a very similar past effort: https://github.com/dotnet/core/blob/main/release-notes/6.0/runtime-deps.json

@mthalman @ashnaga @Falco20019 @omajid @leecow


richlander:
Here's a demo of one way the JSON format can be used.

```bash
$ docker run --rm -it ubuntu:24.04
root@243a281eaab7:/# curl -s https://raw.githubusercontent.com/dotnet/core/linux-packages/release-notes/9.0/linux-packages.json | jq -r .distributions[2].releases[1].packages[0:7][].name | xargs echo                
libc6 libgcc-s1 ca-certificates libssl3t64 libstdc++6 libicu74 tzdata
root@243a281eaab7:/# apt install $(curl -s https://raw.githubusercontent.com/dotnet/core/linux-packages/release-notes/9.0/linux-packages.json | jq -r .distributions[2].releases[1].packages[0:7][].name | xargs echo)
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
libc6 is already the newest version (2.39-0ubuntu8.2).
libgcc-s1 is already the newest version (14-20240412-0ubuntu1).
ca-certificates is already the newest version (20240203).
ca-certificates set to manually installed.
libssl3t64 is already the newest version (3.0.13-0ubuntu3.1).
libstdc++6 is already the newest version (14-20240412-0ubuntu1).
The following NEW packages will be installed:
  libicu74 tzdata
0 upgraded, 2 newly installed, 0 to remove and 2 not upgraded.
Need to get 11.1 MB of archives.
After this operation, 38.5 MB of additional disk space will be used.
```

mthalman (commented):
My expectation is that this file would eventually be used as the source for generating more than just the release markdown file.

For other reviewers: what is being proposed here is essentially an implementation of https://github.com/dotnet/core/issues/5646 which is part of a larger epic: https://github.com/dotnet/core/issues/5651. There are a multitude of areas that need to derive this dependency data (read about them in the https://github.com/dotnet/core/issues/5651 epic).

I'm fine with constraining the scope here for now to just target the release markdown files.

> I'm wondering if now is a good time to rename these files as required-packages.json or runtime-deps.json or similar.

I don't have too much of an opinion here other than to say the naming should be specific to Linux and not Windows since that's the scope of this file. My schema in https://github.com/dotnet/core/issues/5646 had attempted to combine Linux and Windows dependencies with a combined description but, looking back, it seemed unnecessary and made things ugly. If we ever want to describe Windows dependencies, that should be done in a separate file with its own schema.

richlander:
I really just meant that it would be good to have the option to include FreeBSD (and the like) in this list. I agree that including Windows is a step too far. Also, the product doesn't have dependencies on macOS or Windows.

Yes, there are other opportunities to use dependency lists. One of the things I learned from my other efforts is that generating markdown is sufficient for validating the schema and ensuring that the data is maintained.

release-notes/9.0/linux-packages.json
```diff
@@ -0,0 +1,385 @@
...
+                        }
+                    ]
+                }
+            ]
+        },
+        {
+            "name": "Debian",
+            "commands": [
+                "sudo apt-get update",
+                "sudo apt-get install -y --no-install-recommends"
```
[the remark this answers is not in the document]
mthalman: Good point. I agree that it would be more clear to have things separated out. I think something like the following should be sufficient:

```json
"commands": {
  "pre-install": "sudo apt-get update"
  "install": "sudo apt-get install -y --no-install-recommends"
}
```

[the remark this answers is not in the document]
omajid: We solved a similar pre-install/install/query command problem in https://github.com/JL03-Yue/vscode-dotnet-runtime/blob/JL03-Yue/red-hat-sdk-install/vscode-dotnet-runtime-library/distro-data/distro-support.json. Could we re-use/adapt that?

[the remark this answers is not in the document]
richlander: Yes. I think we could just take that as-is.

release-notes/9.0/linux-packages.json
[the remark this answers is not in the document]
richlander: Like this?

https://github.com/dotnet/core/blob/main/release-notes/releases-index.json#L2

release-notes/9.0/linux-packages.json
```diff
@@ -0,0 +1,385 @@
...
+            ],
+            "required": [
+                "Globalization"
+            ]
+        },
+        {
+            "id": "krb5",
+            "name": "Kerberos",
+            "references": [
+                "http://web.mit.edu/kerberos"
```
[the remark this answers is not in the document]
richlander: Oops!

release-notes/9.0/linux-packages.json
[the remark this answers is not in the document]
mthalman: That would be the best option. But even a markdown file that describes it would be good.

release-notes/9.0/linux-packages.json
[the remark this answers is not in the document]
richlander: How does this look?

https://github.com/dotnet/core/pull/9412/commits/d513a42d018ccd46ace0a72bcfd7e22d4029042d

I haven't updated the markdown generate code yet. Wanted to validate the schema change first.

release-notes/9.0/linux-packages.json
```diff
@@ -0,0 +1,411 @@
...
+                "Kerberos networking"
+            ]
+        }
+    ],
+    "distributions": [
+        {
+            "name": "Alpine",
+            "commands": [
+                "sudo apk add"
+            ],
```
mthalman, line 96: Remove

release-notes/9.0/linux-packages.json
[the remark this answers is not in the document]
richlander: I can likely do both.

Might be a good opportunity to try out this new feature: https://github.com/dotnet/core/blob/main/release-notes/9.0/preview/preview6/libraries.md#jsonschemaexporter

billgothacked:
Thanks: )

release-notes/9.0/linux-packages.json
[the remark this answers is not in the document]
richlander: I updated the markdown generator to the new schema.

richlander:
I added a formats description based on the request from @mthalman for more information on the new format. I'm sure that more can be added to the descriptions. I'll add more as I go. Feel free to do same.

https://github.com/dotnet/core/pull/9412/commits/489d5cdb1b58264e2900fc8e9a633a4e080ac9d2

release-notes/formats.md
```diff
@@ -0,0 +1,101 @@
+# Release note formats
+
+Release notes are available in a combination of markdown and JSON formats. The intent is to make all release notes available in markdown and also in JSON if there is a  scenario that requires structured data. This content is [licensed](./license-information.md) for broad use.
+
+Our approach has evolved over time. The following descripions are for our latest practices.
```
mthalman, line 5:
```suggestion
Our approach has evolved over time. The following descriptions are for our latest practices.
```

release-notes/formats.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Monthly patch release notes (markdown)
+
+We typically release an update each month. These often include security fixes for vulberabilities (AKA CVEs) disclosed on the same day.
```
mthalman, line 9:
```suggestion
We typically release an update each month. These often include security fixes for vulnerabilities (AKA CVEs) disclosed on the same day.
```

release-notes/formats.md
```diff
@@ -0,0 +1,101 @@
...
+We typically release a preview for the next major version each month. These are much the same as the monthly patches, however, will not include CVE information until the new major release is supported (starting with RC1).
+
+Examples:
+
+- [.NET 9 Preview 1](./9.0/preview/preview1/README.md)
+- [.NET 9 Preview 6](./9.0/preview/preview6/README.md)
+
+These markdown files are primarily composed of dense feature descriptions. This content is used as source material for [What's New](https://learn.microsoft.com/dotnet/core/whats-new/dotnet-9/overview) pages in official Microsoft docs, which are [often updated](https://github.com/dotnet/docs/pulls?q=is%3Apr+What%27s+New) on the same day as a preview release.
+
+Preview release notes are always in a `preview` folder. This approach was adopted so that preview releases do distract from stable releases once preview releases are no longer relevant.
```
mthalman, line 34:
```suggestion
Preview release notes are always in a `preview` folder. This approach was adopted so that preview releases do not distract from stable releases once preview releases are no longer relevant.
```

release-notes/formats.md
```diff
@@ -0,0 +1,101 @@
...
+## Releases (json)
+
+The `releases.json` file contains expansive release information for the life of a major release. It includes much the same information as the Monthly and preview markdown release notes.
+
+Examples:
+
+- [8.0 `releases.json`](./8.0/releases.json)
+- [9.0 `releases.json`](./9.0/releases.json)
+
+This files can grow to be quite large. We are considering producing release-specific versions of this file, for example for `8.0/8.0.1/release.json`.
```
mthalman, line 77:
```suggestion
These files can grow to be quite large. We are considering producing release-specific versions of this file, for example for `8.0/8.0.1/release.json`.
```

mthalman, line 77: Is there an issue to point to here?

release-notes/9.0/linux-packages.json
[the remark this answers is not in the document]
mthalman: I'm not sure what schema type this is but it differs from the releases-index schema at https://json.schemastore.org/dotnet-releases-index.json. They should use the same format.

The description/comment fields that are provided in the releases-index schema are particularly helpful. That was my original intent: to have an explanation of what things meant within the model.

release-notes/README.md
```diff
@@ -14,12 +14,16 @@ You can find release notes for all releases, including out-of-support releases,
...
 
 * [Download .NET](https://dotnet.microsoft.com/download/dotnet)
-* [Releases Index][releases-index.json] -- Index for all release channels in JSON format
-* [dotnet-install scripts](https://learn.microsoft.com/dotnet/core/tools/dotnet-install-script)
 * [Installation docs](https://learn.microsoft.com/dotnet/core/install/)
+* [dotnet-install scripts](https://learn.microsoft.com/dotnet/core/tools/dotnet-install-script)
+
+## Release notes in JSON format
+
+* [Releases index][releases-index.json]
```
leecow, line 25: It feels like this should remain in the 'Release Info' section.

release-notes/README.md
```diff
@@ -14,12 +14,16 @@ You can find release notes for all releases, including out-of-support releases,
 [8.0.7]: 8.0/8.0.7/8.0.7.md
 [6.0.32]: 6.0/6.0.32/6.0.32.md
 
-## Release Information
+## Download .NET
 
 * [Download .NET](https://dotnet.microsoft.com/download/dotnet)
-* [Releases Index][releases-index.json] -- Index for all release channels in JSON format
-* [dotnet-install scripts](https://learn.microsoft.com/dotnet/core/tools/dotnet-install-script)
```
leecow, line 21: What's the thinking behind removing the install script link? I think I understand that it's the only item directly related to installation so is a bit out of place.

release-notes/README.md
```diff
@@ -14,12 +14,16 @@ You can find release notes for all releases, including out-of-support releases,
...
 * [Installation docs](https://learn.microsoft.com/dotnet/core/install/)
+* [dotnet-install scripts](https://learn.microsoft.com/dotnet/core/tools/dotnet-install-script)
+
+## Release notes in JSON format
```
leecow, line 23: Does this need it's own section? Also, the formats file spends as much time on markdown as it does on json though I think I get wanting to call attention to the new json files.

release-notes/formats.md
richlander: > vulberabilities

These are the truly bad ones!

release-notes/README.md
richlander: I think separating out download info from extra stuff is useful. I suspect 90%+ of people looking at this page are not interesting in advanced scenarios.

richlander:
Good question for @leecow.

https://github.com/dotnet/core/pull/9412#discussion_r1681212234

leecow:
I don't feel strongly that they should match but it's worth looking at. I do agree that any schema we post should include description and example fields.

release-notes/9.0/linux-packages.json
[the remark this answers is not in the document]
richlander: I just asked @eiriktsarpalis if there is a way to use the new feature w/descriptions/comments. I love how easy the schema generator is to use. It's basically a one-liner. Ideally, we'd move the existing releases.json schema over to this scheme if there is a way to preserve the extra information (like with attributes or `///`).

richlander:
@eiriktsarpalis explained to me how to add descriptions with the generated schema. I'll give that a try. Might be a good customer sample!

@mthalman and I chatted about the name. We agreed that `linux-packages.json` isn't the best name because we want to file to be a natural home for BSD packages, too. How about `package-deps.json`? It's not a perfect name, but I also don't have another suggestion. I'd then go with a matching MD file for .NET 9+ and have mis-matched files for earlier versions (to avoid a needless breaking change).

richlander:
I created an issue for the release-specific files.

https://github.com/dotnet/core/issues/9417

release-notes/README.md
richlander: I just moved it up.

richlander:
@mthalman -- How does this look?

https://github.com/dotnet/core/pull/9412/commits/03af9b669257a2bc4603bb95c5a7e8f0dee2f727

The schema is still generated. I used `DescriptionAttribute` on each type in the model (plus a little code) and this is the result. Quite an effective pattern.

release-notes/9.0/linux-packages.json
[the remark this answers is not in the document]
nagilson: +1 for adding a json schema file. We also use a schema for our repo to track linux install commands and it's been rather helpful. https://github.com/dotnet/vscode-dotnet-runtime/blob/main/Documentation/json-schema/distro-support-schema.json

This way VS Code's editor will pick up the schema and validate the format whenever changes are made.

release-notes/9.0/linux-packages.json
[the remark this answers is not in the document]
nagilson: That seems good. The file looks slightly different on our end now but I'm not sure you'd want to include the dpkg lock waits, it's a bit specific to our scenario. https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/distro-data/distro-support.json

It's interesting that the documents for installing .NET on a specific distro are also quite tied to this. It would be good to keep these in parity. I don't know if that necessitates having a repo or some sort of service like the releases.json server to host it, but perhaps.

release-notes/formats.md
richlander: https://github.com/dotnet/core/issues/9417

release-notes/schema/dotnet-required-packages.json
```diff
@@ -0,0 +1,170 @@
...
+            "type": "string"
+          },
+          "References": {
+            "type": [
+              "array",
+              "null"
+            ],
+            "items": {
+              "type": [
+                "string",
```
mthalman, line 37: The set of values used here should be constrained. I was thinking this would be an enum type.

release-notes/schema/dotnet-required-packages.json
mthalman, line 1: Can you also provide description fields for the properties?

One point of clarification that the descriptions could make is to point out the relationship between `packages[].id` and `distributions[].releases[].packages[].id`

release-notes/schema/dotnet-required-packages.json
richlander: `References` is intended to be markdown strings, as you can see in `linux-packages.json`. What would that look like?

mthalman: Sorry, I meant this to be under `Required`. I added my comment there instead: https://github.com/dotnet/core/pull/9412/files/03af9b669257a2bc4603bb95c5a7e8f0dee2f727#r1682969587.

release-notes/schema/dotnet-required-packages.json
```diff
@@ -0,0 +1,170 @@
...
+            }
+          },
+          "Required": {
+            "type": [
+              "array",
+              "null"
+            ],
+            "items": {
+              "type": [
+                "string",
```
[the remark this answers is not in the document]
richlander: Right. That is more what I was expecting you to care about. Ya. I can add an enum there. We can always extend it if we find more scenarios.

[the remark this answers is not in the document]
richlander: FYI: I also have the casing wrong. I need to generate this as `KebabCaseLower`.

release-notes/schema/dotnet-required-packages.json
```diff
@@ -0,0 +1,170 @@
...
+                },
+                "CommandRoot": {
+                  "type": "string"
+                },
+                "CommandParts": {
+                  "type": "array",
+                  "items": {
+                    "type": [
+                      "string",
+                      "null"
```
nagilson, line 98: Is `"null"` necessary to allow an empty array?

release-notes/schema/dotnet-required-packages.json
```diff
@@ -0,0 +1,170 @@
...
+      "type": "string"
+    },
+    "EolDate": {
+      "type": "string",
+      "format": "date"
+    },
+    "Packages": {
+      "type": "array",
+      "items": {
+        "description": "A logical package that will exist in various distros with different package names. Includes the scenarios for which the package is required.",
```
nagilson, line 18: What is meant by a 'logical package'?

release-notes/formats.md
```diff
@@ -0,0 +1,101 @@
+# Release note formats
+
+Release notes are available in a combination of markdown and JSON formats. The intent is to make all release notes available in markdown and also in JSON if there is a  scenario that requires structured data. This content is [licensed](./license-information.md) for broad use.
```
nagilson, line 3:
```suggestion
Release notes are available in a combination of markdown and JSON formats. The intent is to make all release notes available in markdown and also in JSON if there is a scenario that requires structured data. This content is [licensed](./license-information.md) for broad use.
```

release-notes/formats.md
```diff
@@ -0,0 +1,101 @@
...
+These markdown files include the following information:
+
+- Links to binaries
+- Notable changes, including CVEs
+- Compatibility information
+- Updates packages
+
+## Monthly preview release notes (markdown)
+
+We typically release a preview for the next major version each month. These are much the same as the monthly patches, however, will not include CVE information until the new major release is supported (starting with RC1).
```
nagilson, line 25:
```suggestion
We typically release a preview for the next major version each month. These are much the same as the monthly patches; however, they will not include CVE information until the new major release is supported (starting with RC1).
```

release-notes/formats.md
```diff
@@ -0,0 +1,101 @@
...
+Examples:
+
+- [8.0 `releases.json`](./8.0/releases.json)
+- [9.0 `releases.json`](./9.0/releases.json)
+
+These files can grow to be quite large. We are considering producing release-specific versions of this file, for example for `8.0/8.0.1/release.json`.
+
+## Supported OS (json and markdown)
+
+Supported OS information is published for each release. This information indicates to users which OSes they can expect an app (or the .NET SDK) to run on for a given .NET version. It also indicates which OSes are supported and when they transitioned to EOL status.
```
nagilson, line 81: These sections are very well documented here, the language is very clear. I think people will be excited to have the supported-os and linux-package json together.

release-notes/9.0/linux-packages.json
```diff
@@ -0,0 +1,408 @@
...
+                        {
+                            "id": "krb5",
+                            "name": "libgssapi-krb5-2"
+                        }
+                    ]
+                }
+            ]
+        },
+        {
+            "name": "Ubuntu",
```
nagilson, line 276: I think it would be ideal to include Red Hat for the launch as well.

release-notes/schema/dotnet-required-packages.json
richlander: I was also thinking of "nominal package". Open to better terms.

I was trying to do few things:

- Provide additional information on each package in one place, per DRY rules.
- Provide a canonical name for each package since most package names differ by distro and/or distro version.
- Make it possible to write a single script that works across distros w/o having any knowledge of the particular distro and their package names.

release-notes/schema/dotnet-required-packages.json
richlander: That property is defined like this: `IList<string> CommandParts` (AKA, not nullable).

What is the reason for that design choice @eiriktsarpalis?

release-notes/formats.md
richlander: Great! Together, meaning "both"?

release-notes/9.0/linux-packages.json
richlander: Good call. Can you contribute those @omajid?

After we merge this PR, I'll add .NET 6/8 versions, too.

richlander: Also, we want as many distros as possible. I just started with some handy ones to prove out the idea.

release-notes/formats.md
nagilson: Indeed :)

release-notes/schema/dotnet-required-packages.json
nagilson: I like nominal package better! I think saying 'The canonical name for the package that will exist...' works great, but unfortunately there are 2 relevant interpretations of Canonical in this instance. Perhaps 'statutory package name'? 😄

release-notes/9.0/linux-packages.json
```diff
@@ -0,0 +1,408 @@
...
+                            "name": "tzdata"
+                        },
+                        {
+                            "id": "krb5",
+                            "name": "libgssapi-krb5-2"
+                        }
+                    ]
+                }
+            ]
+        },
```
omajid, line 274:
```suggestion
        },
                {
                    "name": "Fedora 40",
                    "release": "40",
                    "packages": [
                        {
                            "id": "libc",
                            "name": "glibc"
                        },
                        {
                            "id": "libgcc",
                            "name": "libgcc"
                        },
                        {
                            "id": "ca-certificates",
                            "name": "ca-certificates"
                        },
                        {
                            "id": "libssl",
                            "name": "openssl-libs"
                        },
                        {
                            "id": "libstdc++",
                            "name": "libstdc++"
                        },
                        {
                            "id": "libicu",
                            "name": "libicu"
                        },
                        {
                            "id": "tzdata",
                            "name": "tzdata"
                        },
                        {
                            "id": "krb5",
                            "name": "krb5-libs"
                        }
                    ]
                }
            ]
        },
```

release-notes/9.0/linux-packages.json
```diff
@@ -0,0 +1,408 @@
...
+            "name": "CA Certificates",
+            "references": [
+                "https://www.redhat.com/sysadmin/ca-certificates-cli"
+            ],
+            "required": [
+                "HTTPS"
+            ]
+        },
+        {
+            "id": "libssl",
```
omajid, line 39: Thoughts on making it more explicit that this is OpenSSL, not another SSL library?

```suggestion
            "id": "openssl",
```

release-notes/9.0/linux-packages.json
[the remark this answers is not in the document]
adegeo: Nope, no automation! But with the addition of these files, I'll be able to write automation for the docs which will help a lot.

richlander:
Back to naming. I don't want to continue with `linux-packages` because "linux" is too narrow.

I propose "os-packages.json". It sort of fits with `supported-os.json` where the context is an OS. It also means that we can add any OS to this that we see fit, Linux, Unix, or something else.

release-notes/schema/dotnet-required-packages.json
richlander: Yup. The double meaning of "canonical" definitely came to mind.

release-notes/schema/dotnet-required-packages.json
richlander: Done.

release-notes/9.0/linux-packages.json
richlander: Thanks!

release-notes/9.0/linux-packages.json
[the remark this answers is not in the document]
richlander: The supported-os.json file might also help.

richlander:
I think we're getting close to a final review on this PR. I believe I've addressed the bulk of the feedback to date.

release-notes/schemas/dotnet-os-packages.json
```diff
@@ -0,0 +1,190 @@
...
+    "eol-date": {
+      "description": "End of life date of product version.",
+      "type": "string",
+      "format": "date"
+    },
+    "packages": {
+      "description": "Set of nominal packages used by product, with descriptions.",
+      "type": "array",
+      "items": {
+        "description": "A nominal package is a distro-agnostic representation of a package, including the scenarios for which the package is required. A nominal package will be referenced a distribution package, with a distribution-specific package name.",
```
mthalman, line 21:
```suggestion
        "description": "A nominal package is a distro-agnostic representation of a package, including the scenarios for which the package is required. A nominal package will be referenced by a distribution package, with a distribution-specific package name.",
```

release-notes/schemas/dotnet-os-packages.json
```diff
@@ -0,0 +1,190 @@
...
+        "type": [
+          "object",
+          "null"
+        ],
+        "properties": {
+          "id": {
+            "description": "Id of nominal package.",
```
mthalman, line 28:
```suggestion
            "description": "ID of nominal package.",
```

release-notes/schemas/dotnet-os-packages.json
```diff
@@ -0,0 +1,190 @@
...
+                "null"
+              ]
+            }
+          }
+        },
+        "required": [
+          "id",
+          "name",
+          "required-scenarios",
+          "references"
```
mthalman, line 68: `references` probably doesn't need to be required.

release-notes/schemas/dotnet-os-packages.json
```diff
@@ -0,0 +1,190 @@
...
+                      "string",
+                      "null"
+                    ]
+                  }
+                }
+              },
+              "required": [
+                "run-under-sudo",
+                "command-root",
+                "command-parts"
```
mthalman, line 121: It's possible that a command would not require any parts.

release-notes/schemas/dotnet-os-packages.json
```diff
@@ -0,0 +1,190 @@
+{
+  "description": "The set of packages required by a given product version for a set of distros.",
+  "type": [
+    "object",
+    "null"
```
mthalman, line 5: What does this `null` stuff even mean for the type? Does that mean its nullable? How would that be represented in the JSON? Is it applicable in all cases?

release-notes/schemas/dotnet-os-packages.json
```diff
@@ -0,0 +1,190 @@
...
+                  "type": "array",
+                  "items": {
+                    "description": "A distro archive package to install, with a reference to a logical package with more information.",
+                    "type": [
+                      "object",
+                      "null"
+                    ],
+                    "properties": {
+                      "id": {
+                        "description": "Reference to nominal package id, to providing access to required scenarios and other information.",
```
mthalman, line 153:
```suggestion
                        "description": "Reference to nominal package id, providing access to required scenarios and other information.",
```

release-notes/9.0/os-packages.json
```diff
@@ -0,0 +1,458 @@
+{
+    "$schema": "../schemas/dotnet-os-packages.json",
+    "channel-version": "9.0",
+    "eol-date": "2026-05-11",
```
mthalman, line 4: What's the need for this? This seems to be overloading the data represented here with data whose source of truth is in a separate file. It seems irrelevant to package descriptions.

richlander: I wondered about that, too. I'll remove it.

release-notes/schemas/dotnet-os-packages.json
richlander: I don't know how that works. @eiriktsarpalis  can likely explain.

release-notes/schemas/dotnet-os-packages.json
richlander: Agreed. It is `IList<string>?`.

https://github.com/richlander/distroessed/blob/f180c74df71817868df497d58dfd3ecce50c4e1c/src/DotnetRelease/PackageRecords.cs#L32

I'm not sure what to do.

release-notes/schemas/dotnet-os-packages.json
richlander: Made it nullable. We'll see if that helps.

release-notes/schemas/dotnet-os-packages.json
eiriktsarpalis: It means that the underlying .NET type is a reference type and reflects the fact that the serializer will accept `null` JSON values when deserializing. There was a long debate as to whether we should be exposing `null` in the schema and ultimately we chose correctness over conciseness.

release-notes/schemas/dotnet-os-packages.json
eiriktsarpalis: Making the corresponding parameter optional should do the trick (i.e. assign to `null`)

release-notes/schemas/dotnet-os-packages.json
eiriktsarpalis: Requiredness and nullability are orthogonal traits. This is true both in JSON schema and in C# semantics where you can have `required` fields that are nullable.

eiriktsarpalis: Conversely, you can have non-nullable parameters that are optional, e.g.
```C#
record MyRecord(string value = "default");
```

richlander: I get it now. It's not exactly intuitive, but I guess it makes sense. In any case, your suggested pattern works. Thanks!

release-notes/schema/dotnet-required-packages.json
richlander: Fixed that with a schema update.

richlander:
One step closer. I applied the patterns provided by @eiriktsarpalis, which helped a lot.

How is it looking @mthalman?

release-notes/9.0/os-packages.md
```diff
@@ -0,0 +1,154 @@
...
+```bash
+sudo apk add \
+    libgcc \
+    ca-certificates \
+    libssl3 \
+    libstdc++ \
+    icu-libs \
+    icu-data-full \
+    tzdata \
+    libgssapi-krb5-2
```
mthalman, line 44: I acknowledge the order that you have the packages listed in the JSON. But the generated output here should be sorted.

richlander: That's fair. I assume alpha order?

mthalman: yep

richlander: Done.

richlander: Good call.

mthalman (approved):

release-notes/9.0/linux-packages.json
```diff
@@ -0,0 +1,408 @@
...
+                            "name": "tzdata"
+                        },
+                        {
+                            "id": "krb5",
+                            "name": "libgssapi-krb5-2"
+                        }
+                    ]
+                },
+                {
+                    "name": "Debian 11 (Bullseye)",
```
lbussell: Should Bullseye be included here if it's EOL before .NET 9's release?

release-notes/schema/dotnet-required-packages.json
```diff
@@ -0,0 +1,170 @@
...
+          "object",
+          "null"
+        ],
+        "properties": {
+          "Id": {
+            "type": "string"
+          },
+          "Name": {
+            "type": "string"
+          },
```
lbussell: Should we consider the case where functionality in some distros could be split across multiple packages? It's not the best example since it's not really a separate distro, but Ubuntu Chiseled splits packages into multiple slices. However I don't think we've needed to install multiple slices of one package at once yet though.

billgothacked:
Hi: )

release-notes/9.0/linux-packages.json
richlander: Good call. Definitely ironic after our discussions about Debian 11 this week. Removed.

release-notes/schema/dotnet-required-packages.json
richlander: I think that should be fine. I think we already have that case. Yes?

https://github.com/dotnet/core/blob/0babb5dc73c7ecc8e835da8efc8055d5b45d84b6/release-notes/9.0/os-packages.json#L121-L128

release-notes/schemas/dotnet-os-packages.json
```diff
@@ -0,0 +1,184 @@
...
+      "type": "array",
+      "items": {
+        "description": "An operating system distribution, with required package install commands and specific packages for distribution releases.",
+        "type": [
+          "object",
+          "null"
+        ],
+        "properties": {
+          "name": {
+            "description": "Name of the distribution, matching ID in /etc/os-release,however, the expectation is that this value starts with a capital letter (proper noun).",
```
jkotas, line 78:
```suggestion
            "description": "Name of the distribution, matching ID in /etc/os-release, however, the expectation is that this value starts with a capital letter (proper noun).",
```

richlander:
I added a new `min-version` concept for packages. @bartonjs and @vcsjones are considering increasing the min-version of OpenSSL to 1.1.1 (which is likely a good idea). I wanted to try that out. I updated the schema, included this particular example in the JSON, and included this data in the "Notes" section of the markdown table.

https://github.com/dotnet/core/blob/3a8b35f332e4185340b4a3eae54a1a5936c8b180/release-notes/9.0/os-packages.json#L40

release-notes/schemas/dotnet-os-packages.json
richlander: Thanks! I updated the object model with this change and then re-generated the JSON schema.

release-notes/schemas/dotnet-os-packages.json
richlander: Thanks!

billgothacked:
Will you sponsor me. ***@***.*** Thanks-peace out

richlander:
I plan to merge this PR on Monday.

release-notes/schema/dotnet-required-packages.json
lbussell: Didn't notice that before. Looks good!

release-notes/9.0/os-packages.json
```diff
@@ -0,0 +1,421 @@
...
+                        {
+                            "id": "krb5",
+                            "name": "libgssapi-krb5-2"
+                        }
+                    ]
+                }
+            ]
+        },
+        {
+            "name": "Debian",
```
mthalman, line 182: The removal of Debian 11 made me realize that this content should be in sync with https://github.com/dotnet/core/blob/main/release-notes/9.0/supported-os.json. In other words, one could imagine a validation script that checks that all the distros listed here can be cross-referenced to supported-os.json. I see that supported-os.json has both an `id` and `name` for each distro. What's the need for both? Should this file reference by `id` instead?

Similarly for the release, you list a friendly name like `Debian 12 (Bookworm)`. Why is that detailed information presented here (for a file that is package-centric) when that information isn't even contained in supported-os.json? Really, I see that information as being superfluous. You just need to know the distro id/name and version. In fact, I would argue that `release` should be renamed to `version` to more clearly indicate the relationship between these two files.

release-notes/9.0/os-packages.json
```diff
@@ -0,0 +1,421 @@
...
+                            "id": "icu",
+                            "name": "icu-data-full"
+                        },
+                        {
+                            "id": "tzdata",
+                            "name": "tzdata"
+                        },
+                        {
+                            "id": "krb5",
+                            "name": "libgssapi-krb5-2"
```
mthalman, line 137:
```suggestion
                            "name": "krb5"
```

release-notes/9.0/os-packages.json
```diff
@@ -0,0 +1,421 @@
...
```
mthalman, line 175:
```suggestion
                            "name": "krb5"
```

release-notes/9.0/os-packages.json
richlander: The friendly name is so that it is easy to generate markdown. Certainly, we could use the following implementation: $"{distro} {version}". I like having the codenames in the markdown since we use both the version numbers and codenames in different parts of our infra. It also enables a long distro name like "Red Hat Enterprise Linux", which doesn't fit very nicely in an "id" field.

https://github.com/dotnet/core/pull/9412/files#diff-c3805f9c7f5bc85d9d552816365e70fad4a2243ba220a607e2a1c1360e44fc17R33

In the other file, the `id` field can be thought of as `endoflifedate-id`. Those values match what endoflife.date uses: https://endoflife.date/. The tools we wrote query that database extensively. The `name` field is what is used in the generated markdown.

Excellent point on keeping in sync. I'm going to be writing some more tools related to staying in sync. That's a good scenario to include.

richlander: I haven't published a schema for `supported-os.json` yet. I'll make sure the connection with endoflife.date is documented.

release-notes/schemas/dotnet-os-packages.json
eiriktsarpalis: This PR doesn't include the code that was used to generate this schema, correct?

release-notes/9.0/os-packages.json
```diff
@@ -134,7 +134,7 @@
                         },
                         {
                             "id": "krb5",
-                            "name": "libgssapi-krb5-2"
+                            "name": "krb5"
```
mthalman, line 137: This will require regeneration of the markdown.

release-notes/schemas/dotnet-os-packages.json
richlander: It does not. I haven't figured out the long-term home for these tools.

Schema generation: https://github.com/richlander/distroessed/blob/linux-packages/src/GenerateJsonSchemas/Program.cs

OM: https://github.com/richlander/distroessed/blob/linux-packages/src/DotnetRelease/PackageRecords.cs

release-notes/9.0/os-packages.json
richlander: Good call!

mthalman (approved):

arrowd:
On FreeBSD we have `libc` as part of the base system, so we need a way to somehow hide dependencies based on OS. The same goes for `libgcc` as we have LLVMs `libcompilerrt` as a replacement.

richlander:
@arrowd -- I don't think that will be a problem. The nominal packages list is intended to be a superset view of packages. Each distro then is able to subscribe to that list, with a partial or full set. When the markdown is generated, it uses the distro packages only. So, I think the scheme already works as you need. Feel free to contribute to FreeBSD packages and I think it will just work.

arrowd:
Sure, I'll try to find time for a follow-up after this PR lands.

Issue #9417, opened 2024-07-17
Create `release.json` files per patch release


richlander:
We should create patch-release-specific `release.json` files.

With such a scheme, I'd expect to see:

- `release.json` at this location https://github.com/dotnet/core/tree/main/release-notes/8.0/8.0.1
- `release-index.json` at this location https://github.com/dotnet/core/tree/main/release-notes/8.0

The primary motivator for this is that `releases.json` can get very large:

- https://github.com/dotnet/core/blob/main/release-notes/6.0/releases.json
- https://github.com/dotnet/core/blob/main/release-notes/8.0/releases.json

This demonstrates the problem (and .NET 6 still has a ways to go):

```bash
$ curl -s https://raw.githubusercontent.com/dotnet/core/main/release-notes/6.0/releases.json | wc -l
   24680
```

That's 24k lines.

My hypothesis for JSON users:

- Most are satisfied with `release-index.json` (just looking for version numbers and the CVE bool).
- The vast majority of the remaining users only want to know about the latest release.
- Many of the users that need to look at all releases would prefer to read/process one release at a time to better control memory usage.

We could write a tool to backfill all the existing directories. That would be pretty easy. That could be done in parallel with establishing the scheme.

Related: https://github.com/dotnet/core/issues/9205

@leecow

leecow:
This sounds like a good improvement. I'll open a tracking issue in our internal tooling repo to track the necessary work. 

/cc @rbhanda

richlander:
Do you have any thoughts on the index file?

richlander:
Here is a first take on the index format.

```json
{
    "channel-version": "8.0",
    "support-phase": "active",
    "release-type": "lts",
    "release-date": "2023-11-08",
    "eol-date": "2026-11-10",
    "announcement": "https://devblogs.microsoft.com/dotnet/announcing-dotnet-8/",
    "release-notes": "https://github.com/dotnet/core/blob/main/release-notes/8.0/README.md",
    "lifecycle-policy": "https://aka.ms/dotnetcoresupport",
    "releases.json": "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/8.0/releases.json",
    "supported-os.json": "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/8.0/supported-os.json",
    "releases": [
        {
            "version": "8.0.7",
            "date": "2024-07-09",
            "security": true,
[...]

PR #9427, opened 2024-07-23
Add release-specific release.json files

richlander:
Fixes https://github.com/dotnet/core/issues/9417.

Improvements:

- Add READMEs for new content
- Generate `releases-index.json`
- Generate `patch-releases-index.json`, per release
- Generate `release.json`, per patch release
- Update `releases.json`, per major release, to include all the required information for generating the other files
- Generate/add schemas for all JSON formats
- Add `$schema` reference to JSON files

Notes: 

- I only generated the fancy new files for releases in-support. That had the benefit of making it so that the algorithm doesn't continue to re-generate these files for releases no longer generating new content.
- I didn't add a schema reference to the generated files, both because it doesn't seem necessary (no edits are intended) and I do not know how.
- I'll add `os-packages.json` files for .NET 6 and 8 in a later PR.

@Falco20019 @martincostello @leecow


release-notes/9.0/patch-release-index.json
```diff
@@ -0,0 +1,41 @@
+{
+  "channel-version": "9.0",
+  "releases": [
+    {
+      "release-version": "9.0.0-preview.6",
+      "release-date": "2024-07-09",
+      "security": false,
+      "release.json": "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata//9.0/preview/preview6/release.json"
```
martincostello, line 8: Just thinking out loud - would it be better to name this something as to what it is, rather than just the name of the file (`release-info-uri` or something)?

release-notes/9.0/patch-release-index.json
```diff
@@ -0,0 +1,41 @@
+{
+  "channel-version": "9.0",
```
martincostello, line 2: Is it worth having an explicit `latest-version` or similar property to act as a key into `releases`, rather than relying on `releases[0]` always being the latest?

release-notes/formats.md
```diff
@@ -1,35 +1,47 @@
...
+- [9.0/releases.json](./9.0/releases.json)
+- [9.0/patch-release-index.json](./9.0/patch-release-index.json)
+- [Patch release overview](./9.0/preview/preview1/release.json)
 
-We typically release a preview for the next major version each month. These are much the same as the monthly patches; however, they will not include CVE information until the new major release is supported (starting with RC1).
+## Monthly preview release notes
+
+We typically release a preview for the next major version each month. These are much the same as the monthly patches.
+
+They priumarily differ in two ways:
```
martincostello, line 34:
```suggestion
They primarily differ in two ways:
```

release-notes/formats.md
```diff
@@ -1,35 +1,47 @@
...
+
+- Dense feature descriptions
+- No CVE information (since the releases are not supported)
```
martincostello, line 37: Is this still true when it gets to RC stage?

release-notes/schemas/dotnet-patch-releases-index.json
```diff
@@ -0,0 +1,50 @@
...
+            "description": "Version of release.",
+            "type": "string"
+          },
+          "release-date": {
+            "description": "Date of release.",
+            "type": "string",
+            "format": "date"
+          },
+          "security": {
+            "description": "Security status of release.",
```
martincostello, line 29: This description only makes sense to me if it were an enum - what about something like _"Whether the release contains any CVE fixes"_?

release-notes/schemas/dotnet-patch-releases-index.json
```diff
@@ -0,0 +1,50 @@
...
+            "type": "boolean"
+          },
+          "release.json": {
+            "description": "Url to detailed description of release.",
```
martincostello, line 33:
```suggestion
            "description": "URL to detailed description of release.",
```

release-notes/schemas/dotnet-patch-releases-index.json
```diff
@@ -0,0 +1,50 @@
...
+    "releases": {
+      "type": "array",
+      "items": {
+        "type": [
+          "object",
+          "null"
+        ],
+        "properties": {
+          "release-version": {
+            "description": "Version of release.",
```
martincostello, line 20: Maybe be more specific about what the version is (it's the "marketing" or "brand" version, right?)

release-notes/schemas/dotnet-patch-release.json
```diff
@@ -0,0 +1,409 @@
...
+          "description": "The date of the patch release.",
+          "type": "string",
+          "format": "date"
+        },
+        "release-version": {
+          "description": "The version of the release",
+          "type": "string"
+        },
+        "security": {
+          "description": "The security status of the release.",
```
martincostello, line 25: Same comment here.

release-notes/9.0/patch-release-index.json
richlander: Great topic. I am not a fan of the current naming scheme. I just copied what was already there for consistency.

Example: https://github.com/dotnet/core/blob/a6569a1037f9405b678b30c03ece7df475c04c55/release-notes/releases-index.json#L16

An argument could be made that keeping each file consistent makes sense, but that this is a new file so an opportunity to do better. Compelling?

release-notes/formats.md
richlander: Good catch. I had that text in there before, but it got lost. I can add it back.

release-notes/9.0/patch-release-index.json
martincostello: I can see both arguments - consistency is nice, but holding back improvement for the sake of it isn't great either. I could go either way 😄

baronfel: I'll say that as a consumer of this file having property keys with `.` in them is a _giant pain_.

richlander: How does that show up @baronfel? 

We could also come up with a new scheme, enable those everywhere and keep the only properties for compat.

baronfel: For .NET, when defining binding models for this property you have to annotate the property with the JsonPropertyName attribute, that's the main pain point. Some bindings in other libraries or tools (like `jq` path patterns) are also tricky to encode compared to property keys with dashes or underscores.

richlander: OK. I'm with you. I'm convinced. I'll address this. I'm very much not a fan of this naming scheme.

richlander:
I think I got all of that @martincostello.

The formats file still needs some help.

release-notes/9.0/patch-release-index.json
```diff
@@ -1,41 +1,43 @@
...
   "channel-version": "9.0",
+  "latest-release": "9.0.0-preview.6",
+  "latest-release-date": "2024-07-09",
   "releases": [
     {
       "release-version": "9.0.0-preview.6",
       "release-date": "2024-07-09",
       "security": false,
-      "release.json": "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata//9.0/preview/preview6/release.json"
+      "release-info-uri": "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata//9.0/preview/preview6/release.json"
```
Falco20019: The URI seems to be a bit strange, maybe an issue in the generator (if you generated it).
```suggestion
      "release-info-uri": "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/9.0/preview/preview6/release.json"
```

release-notes/9.0/patch-releases-index.json
```diff
@@ -0,0 +1,43 @@
+{
+  "channel-version": "9.0",
+  "latest-release": "9.0.0-preview.6",
+  "latest-release-date": "2024-07-09",
+  "releases": [
+    {
+      "release-version": "9.0.0-preview.6",
+      "release-date": "2024-07-09",
+      "security": false,
+      "release-info-uri": "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata//9.0/preview/preview6/release.json"
```
martincostello, line 10: Double `//`s in these URIs.

release-notes/9.0/patch-release-index.json
richlander: Thanks. Will check the generator.

richlander: Fixed.

release-notes/9.0/patch-releases-index.json
richlander: Fixed.

martincostello (approved):

richlander:
I believe that this PR is ready to go.

Much of this is based on an OM I built, which is the basis of the schema generation and the serializer for `release.json`. I did a test where I deserialized and then serialized `releases.json` and then used a json diff tool to see how before and after compared. The test passed. I also validated the files with these new schema. That also seems to pass (you can test this yourself). 

We need to wait for @leecow to validate the following changes:

- Updates to `releases-index.json` and `releases.json` files
- Addition of `patch-releases-index.json` and `release.json` files

These all need to be uploaded/updated in blob storage before merging.

We should also double check that everyone is good with the new property and file names.

release-notes/releases-index.json
Falco20019, line 1: Add the `$schema`.

Falco20019 (commented):
See some comments on https://github.com/richlander/distroessed/pull/5#pullrequestreview-2207417510 for changes to the generated schemas. Usually, schemas also do set `"additionalProperties": false` to make it clear that there are no additional fields allowed in the JSON. Not fully required, but would make intention more clear.

release-notes/releases-index.json
eiriktsarpalis: By default STJ allows additional properties which is reflected in the generated schema. This setting can be overridden using the `JsonUnmappedMemberHandlingAttribute`.

richlander: I made this update.

https://github.com/dotnet/core/pull/9427/commits/e86853259f2e70b601edfc30b8707c3bda372f72

richlander:
I had to do process a merge conflict due to https://github.com/dotnet/core/pull/9424. I knew that would happen, but that's how it goes.

Falco20019 (approved):

bigdaddykane18 (approved):

bigdaddykane18 (approved):

richlander:
This work will be broken into multiple PRs.

PR #9434, opened 2024-07-29
os-packages.json: Add FreeBSD section

arrowd:
@richlander

### DIFF
--- a/release-notes/9.0/os-packages.json
+++ b/release-notes/9.0/os-packages.json
@@ -286,6 +286,33 @@
             ]
         },
         {
+            "name": "FreeBSD",
+            "install-commands":
+            [
+                {
+                    "run-under-sudo": true,
+                    "command-root": "pkg",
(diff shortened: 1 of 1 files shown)


release-notes/9.0/os-packages.json
```diff
@@ -416,6 +416,33 @@
...
+                    "run-under-sudo": true,
+                    "command-root": "pkg",
+                    "command-parts": ["install", "-A", "{packageName}"]
+                }
+            ],
+            "releases": [
+                {
+                    "name": "FreeBSD 14.1",
+                    "release": "14.1",
+                    "packages": [
```
richlander, line 434: Does this mean that the other packages (like OpenSSL and ca-certs) are installed by default? If not, what's the story on that?

richlander:
Thanks!

richlander:
The intent is for the distros/OSes to be in alpha order. Can you place FreeBSD just after Fedora?

PR #9441, opened 2024-08-09
Remove invalid versions

Falco20019:
I removed some invalid versions that sadly found their way into the JSON files.

/CC @richlander


release-notes/6.0/supported-os.json
```diff
@@ -338,7 +338,6 @@
                         "11-21h2-e",
                         "10-21h2-e-lts",
                         "10-21h2-iot-lts",
-                        "10-20h2-e-lts",
```
Falco20019, line 341: Windows 10 20H2 was only a semi-anual version and never a LTS. Therefore there was also no information on EndOfLife.date.

release-notes/6.0/supported-os.json
```diff
@@ -393,11 +392,9 @@
                         "x86"
                     ],
                     "supported-versions": [
-                        "23H2",
```
Falco20019, line 396: According to https://mcr.microsoft.com/en-us/product/windows/servercore and https://learn.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle there was never a release of 23H2 or 2012.

Falco20019: And besides https://learn.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle stating that 1607 is only Semi-Anual and released on Nano Server, whereas 2016 is the only Long-Term on Server Core:
image

https://mcr.microsoft.com/en-us/product/windows/servercore shows that it's instead handled like the Server-Core case with 2019 and 1809:
image

release-notes/6.0/supported-os.json
richlander: You are right. Not sure what happened there. Likely a copy/paste error.

release-notes/6.0/supported-os.json
richlander: Server 2012: https://learn.microsoft.com/en-us/lifecycle/products/windows-server-2012

richlander:
LGTM. Add 2012 back and then we can merge.

Thanks for noticing this!

Falco20019:
I only removed 2012 from windows-server-core and not windows-server as there isn't an image:
image

release-notes/6.0/supported-os.json
Falco20019: But since there are no images available, there is nothing to have support for 🤔

Falco20019: Ok, I found https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/hh846313(v=vs.85) so I missed to add 2012 to EoL.date.

Will have a PR there tomorrow and fix it here now.

richlander:
My point was that Windows Containers started with 2016 so this search isn't testing anything. Server Core pre-dates containers.

Falco20019:
My point was about which version server core was introduced with as installation option. Seems to be 2012, right? So adding that to EoL.date should be enough to cover all versions?

richlander:
Got it. Sorry about that.

richlander (approved):

Falco20019:
I will add 2012 here in 1H once I'm on the office and make a PR for it on EoL.date :)

Falco20019:
@richlander I re-added it to .NET 6 and 8 and added it to .NET 7 and 9 where it had been missing in the first place. I did so, because `windows-server` lists 2012 explicitly on those two as well. Is this correct?

richlander:
That all tracks for me.

We don't add OSes to unsupported if we are still supporting them, like the ESU case you call out. For clarity, the decision to support an OS longer than standard EOL is arbitrary (not documented by policy) and not specific to MS-owned OSes. We have on occasion chosen to support Ubuntu longer than standard EOL since we had multiple customers who needed that. I am sure this will happen again.

That makes these files all the more valuable as a source of truth.

release-notes/6.0/supported-os.json
richlander: Shouldn't `10-20h2-e` and `10-20h2-2` be added to the unsupported list?

release-notes/9.0/supported-os.json
```diff
@@ -294,10 +294,11 @@
                         "x86"
                     ],
                     "supported-versions": [
-                        "23H2",
                         "2022",
                         "2019",
-                        "2016"
+                        "2016",
+                        "2012-R2",
```
richlander, line 300: I'm asking for insight on 2012/R2 on .NET 9 support, from @jamshedd.

Falco20019: Same for 2008 then since it's in ESU4: https://learn.microsoft.com/de-de/lifecycle/products/windows-server-2008-r2
Was never listed as supported in the past (also not on the markdown before having the JSON).

richlander: I got confirmation that we still support 2012/R2, including for .NET 9.

richlander:
Is that it? Read to merge?

release-notes/9.0/supported-os.json
Falco20019: Question about 2008 is still open.

richlander: I was saying it is correct that 2008 was never listed. We never supported it, so it should not be in either supported or unsupported lists. This one isn't a function of ESU. It's that we don't test it and it may not work. We do test Server 2012.

release-notes/6.0/supported-os.json
Falco20019: 10-20h2-e and 10-20h2-w are already on the unsupported list. Not sure what you mean by 10-20h2-2.

release-notes/9.0/supported-os.json
Falco20019: Thanks for the clarification. Then this should be fine. EoL.date was merged, so I will have another run of distrossed and the core-support tooling to see if any more might be off. But should be fine to merge otherwise :)

richlander:
Merging. There will also be opportunity for more changes!

Thanks for the contribution.

PR #9455, opened 2024-08-13
WPF .NET 9 Preview 7 Known Issues

harshit7962:
The following is a doc mentioning known-issues in WPF in .NET 9 preview 7 release.

### DIFF
--- a/release-notes/9.0/known-issues.md
+++ b/release-notes/9.0/known-issues.md
@@ -2,4 +2,42 @@
 
 You may encounter the following known issues, which may include workarounds, mitigations, or expected resolution timeframes.
 
-**There are currently no known issues.**
+## .NET WPF
+
+#### Usage of incorrect types as DynamicResource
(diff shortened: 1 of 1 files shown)


jamesmontemagno:
@richlander  is this where we want these? I would think if it is for a specific release it would be somewhere about that release?

harshit7962:
I think it should have ideally been placed in specific folder of preview 7, but I did not see any of the previous versions with a known-issue file. 
Open to including this in preview 7 folder if needed.

richlander:
This is the preferred pattern (known issues at root). We document one set of known issues as a living document. This means the statement should be removed when we ship RC1 if that's appropriate. If people want more historical information, they can use git history.

This policy is a bit different from the others. The reason we do it this way is so that there is only ever one place to look, search engines don't serve up old known issues, and we push users to latest. We don't support any one preview or patch. We support the major version.

jamesmontemagno (approved):

PR #9462, opened 2024-08-14
GetTempPath known issues documentation

rbhanda:


release-notes/6.0/known-issues.md
```diff
@@ -2,6 +2,26 @@
...
+
+An observable difference between the GetTempPath and GetTempPath2 Win32 APIs is that they return different values for SYSTEM and non-SYSTEM processes. When calling this function from a process running as SYSTEM it will return the path %WINDIR%\SystemTemp, which is inaccessible to non-SYSTEM processes. This return value for SYSTEM processes cannot be overridden by environment variables. For non-SYSTEM processes, GetTempPath2 will behave the same as GetTempPath, respecting the same environment variables to override the return value.
```
GrabYourPitchforks, line 8: Recommend striking these sentences. We discussed it offline, but the gist is that the statements are not 100% accurate, and we (.NET) should not make ourselves appear as the authoritative source of information for a Windows API.

~~When calling this function from a process running as SYSTEM it will return the path %WINDIR%\SystemTemp, which is inaccessible to non-SYSTEM processes. This return value for SYSTEM processes cannot be overridden by environment variables.~~

If you want to call attention to the specific problem people are having, you could couch it in point-in-time terms, like:

As of this writing (August 2024), when running as a SYSTEM-level process, environment variables like `%TEMP%` are not honored by the `GetTempPath2` Win32 API. This could cause behavioral differences for applications which run as SYSTEM and which previously relied on the `%TEMP%` environment variable controlling the return value of the `Path.GetTempPath` API.

release-notes/6.0/known-issues.md
```diff
@@ -2,6 +2,26 @@
...
+**Temporary Workaround**
+
+Warning: The opt-out will disable the security fix for the elevation of privilege vulnerability detailed in [CVE 2024-38081](https://msrc.microsoft.com/update-guide/vulnerability/CVE-2024-38081). The opt-out is only for temporary workaround if you're sure that the software is running in secure environments. Microsoft does not recommend applying this temporary workaround.
+  - Option#1: Opt out in a command window by setting the environment variable
+	  -- COMPlus_Disable_GetTempPath2=true
+  -	Opiton#2: [Opt out globally on the machine by creating a system-wide environment variable and rebooting to ensure all processes observe the change](https://stackoverflow.com/questions/2365307/what-determines-the-return-value-of-path-gettemppath). 
+   --	System-wide environment variables can be set by running “sysdm.cpl” from a cmd window, and then navigating to “Advanced -> Environment variables -> System variables -> New”. 
+
+**Resolution**
+The API behavior change is by design to address the elevation of privilege vulnerability. Any affected software or application is expected to make code change to adapt to this new design change.
```
GrabYourPitchforks, line 23: > Any affected software or application is expected to make code change to adapt to this new design change.

This seems... hostile? Who needs to make a code change, and what type of code change are they expected to make?

release-notes/6.0/known-issues.md
TaraOverfield: @GrabYourPitchforks - Are you suggesting that we add the extra call out in the beginning or the end of the known issues section?  What do you think of something like this: 

The .NET Framework servicing update released on [July 2024 security and quality rollup - .NET Framework](https://learn.microsoft.com/en-us/dotnet/framework/release-notes/2024/07-09-july-security-and-quality-rollup) contains a security fix addressed an elevation of privilege vulnerability detailed in [CVE 2024-38081](https://[msrc.microsoft.com/update-guide/vulnerability/CVE-2024-38081](https://msrc.microsoft.com/update-guide/vulnerability/CVE-2024-38081)). The fix changed System.IO.Path.GetTempPath method return value. If windows version exposes the GetTempPath2 Win32 API, this method invokes that API and returns the resolved path. See [the Remarks section](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w#remarks) of the [GetTempPath2](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w) documentation for more information on how this resolution is performed, including how to control the return value through the use of environment variables. The GetTempPath2 API may not be available on all versions of Windows.
 
An observable difference between the GetTempPath and GetTempPath2 Win32 APIs is that they return different values for SYSTEM and non-SYSTEM processes.  As of this publishing, in August 2024, when running as a SYSTEM-level process, environment variables like `%TEMP%` are not honored by the `GetTempPath2` Win32 API. This could cause behavioral differences for applications which run as SYSTEM and which previously relied on the `%TEMP%` environment variable controlling the return value of the P`ath.GetTempPath` API.  For non-SYSTEM processes, GetTempPath2 will behave the same as [GetTempPath](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppatha), respecting the same environment variables to override the return value.  
 
In some scenarios it may be possible to redirect the Temp folder to a different folder using environment variables or other means. Please refer to the official documentation for the [GetTempPath2](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w) Win32 API for the most up to date information on this behavior.

release-notes/6.0/known-issues.md
TaraOverfield: I agree the statement seems harsh.  There was much discussion about the resolution prior to making the initial post.  It seemed necessary to give customers a direction of what they need to do for a permanent fix other than opting out of the security changes.  

Can you suggest a better way to inform customers that they should really make a code change?  It seems a little light in content to leave the resolution as "The API behavior change is by design to address the elevation of privilege vulnerability."

release-notes/6.0/known-issues.md
```diff
@@ -2,6 +2,26 @@
...
+
+Please reference to the [System.IO.Path.GetTempPath API](https://learn.microsoft.com/dotnet/api/system.io.path.gettemppath?view=netframework-4.8.1&tabs=windows) for more information.
+
+**Temporary Workaround**
+
+Warning: The opt-out will disable the security fix for the elevation of privilege vulnerability detailed in [CVE 2024-38081](https://msrc.microsoft.com/update-guide/vulnerability/CVE-2024-38081). The opt-out is only for temporary workaround if you're sure that the software is running in secure environments. Microsoft does not recommend applying this temporary workaround.
+  - Option#1: Opt out in a command window by setting the environment variable
+	  -- COMPlus_Disable_GetTempPath2=true
+  -	Opiton#2: [Opt out globally on the machine by creating a system-wide environment variable and rebooting to ensure all processes observe the change](https://stackoverflow.com/questions/2365307/what-determines-the-return-value-of-path-gettemppath). 
+   --	System-wide environment variables can be set by running “sysdm.cpl” from a cmd window, and then navigating to “Advanced -> Environment variables -> System variables -> New”.
```
jeffhandley, line 20: I don't think we should include "Option#2" here, for a couple reasons:

1. It doesn't apply to the breaking change, since it only discusses environment variables that used to be honored and are still honored when running as non-SYSTEM. I found this misleading and contradictory.
2. It links out to a StackOverflow post instead of official documentation.

We can also change the environment variable guidance to not explicitly state how to set it, as setting it through a command window might not be applicable.

```suggestion
Warning: The opt-out will disable the security fix for the elevation of privilege vulnerability detailed in [CVE 2024-38081](https://msrc.microsoft.com/update-guide/vulnerability/CVE-2024-38081). The opt-out is only for temporary workaround if you're sure that the software is running in secure environments. Microsoft does not recommend applying this temporary workaround.

To apply the opt-out, set the `COMPlus_Disable_GetTempPath2` environment variable to `true` in the necessary context. System-wide environment variables can be set by running `sysdm.cpl` and navigating to "Advanced -> Environment variables -> System variables -> New".
```

release-notes/8.0/known-issues.md
```diff
@@ -2,6 +2,26 @@
 
 You may encounter the following known issues, which may include workarounds, mitigations, or expected resolution timeframes.
 
+## .NET Servicing July 2024 update
```
jeffhandley, line 5: Once we settle on the text above, we'll of course need to replicate it into this file too.

release-notes/6.0/known-issues.md
GrabYourPitchforks: Yeah, I'm honestly not quite sure what to put here. The reason we made the code change is that we wanted customers to be secure without them needing to modify their own code. :)

If we felt compelled to write something, maybe something like "if you know your app depends on (some env var), query that specific env var"? That's generally good practice for limiting code fragility anyway. It makes the code more closely tied to the thing it cares about rather than bouncing through an indirection and all the implementation details that might come with it.

jeffhandley:
@rbhanda I just realized this PR is still open with unresolved feedback

release-notes/8.0/known-issues.md
Supermenwild: ok

Supermenwild (commented):
gh pr checkout 9462

richlander:
Should we close this now?

PR #9479, opened 2024-08-22
Update 9.0 OS Packages

richlander:
Contributes to https://github.com/dotnet/core/issues/9417 and borrows from https://github.com/dotnet/core/pull/9427.

@Falco20019 @leecow @omajid

### DIFF
--- a/release-notes/9.0/os-packages.md
+++ b/release-notes/9.0/os-packages.md
@@ -101,6 +101,36 @@ sudo pkg install -A \
     krb5
 ```
 
+## RHEL
+
+### RHEL 8
+
+```bash
+sudo dnf install -y \
+    ca-certificates \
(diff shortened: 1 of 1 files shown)


Falco20019 (approved):
Looks good. I missed your PR and did some of the same stuff on https://github.com/dotnet/core/pull/9481 (all but the `os-packages.md`) based on the updated JSON files. So I would merge this one before making the PR on mine to reduce the change-set to review.

richlander:
We can also make this one just about os-packages so that they don't conflict. I am happy to do that if you want.

richlander:
I updated/narrowed the commit and forced pushed to make it only about the packages file. That will avoid conflicts https://github.com/dotnet/core/pull/9481.

PR #9481, opened 2024-08-23
Fix supported-os.{md,json}

Falco20019:
Contributes to #9417 and includes generated parts similar to #9479. So I would recommend doing this PR after #9479 is merged to reduce the change-set in the PR.

* Moved Debian 11 from supported to unsupported on .NET ~6, 7~, 8
* Moved RHEL 7 from supported to unsupported on .NET 6, ~7~
* Moved Ubuntu 23.10 from supported to unsupported on .NET 6, ~7~, 8
* Added SUSE Enterprise Linux 12.5 to supported on .NET 8
* Added SUSE Enterprise Linux 15.4 to unsupported on .NET 8
* ~Moved SUSE Enterprise Linux 15.4 from supported to unsupported on .NET 7~
* Added SUSE Enterprise Linux 15.6 to supported on .NET 6, 8, 9
* Regenerated supported-os.md which also fixed `Windows Server Core` entries (already in #9479)

~I did update .NET 7 even though technically not required since it's EOL anyway. Just did it to also see if I covered everything in my adjusted distroessed tooling. If you like to keep it on the old state, I can revert the changes. I did not add any new versions like SLES 15.6 to .NET 7 of course :)~

I compared all versions to the last-before-being-generated ones and according to https://github.com/dotnet/core/blob/56290a0acf2e15e4aa8b0ffa0a02f41501fc088e/release-notes/8.0/supported-os.md we seem to have dropped SLES some versions on .NET 8 (which I re-added here).

/CC @richlander @leecow @omajid

### DIFF
--- a/release-notes/6.0/supported-os.json
+++ b/release-notes/6.0/supported-os.json
@@ -1,6 +1,6 @@
 {
     "channel-version": "6.0",
-    "last-updated": "2024-07-11",
+    "last-updated": "2024-08-23",
     "families": [
         {
             "name": "Android",
@@ -184,10 +184,10 @@
                         "x64"
                     ],
                     "supported-versions": [
-                        "12",
-                        "11"
+                        "12"
                     ],
                     "unsupported-versions": [
+                        "11",
                         "10"
                     ]
                 },
@@ -244,7 +244,9 @@
                     ],
                     "supported-versions": [
                         "9",
-                        "8",
+                        "8"
+                    ],
+                    "unsupported-versions": [
                         "7"
                     ],
                     "notes": [
@@ -261,6 +263,7 @@
                         "x64"
                     ],
                     "supported-versions": [
+                        "15.6",
                         "15.5",
                         "12.5"
                     ],
@@ -285,11 +288,11 @@
                     ],
                     "supported-versions": [
                         "24.04",
-                        "23.10",
                         "22.04",
                         "20.04"
                     ],
                     "unsupported-versions": [
+                        "23.10",
                         "23.04",
                         "22.10",
                         "21.10",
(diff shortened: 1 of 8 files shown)


richlander:
For .NET 7 (and any other EOL version), the intent is to leave the information as accurate at the time of EOL, so that we have a historical document. IMO, it is harmful to end up with a document that says "no OSes supported". What's useful is to know which OSes .NET 7 was supported on (at EOL), for historical interest, 5 years from now.

So, it is fine to continue to update the .NET 7 document as long as we're making it accurate to that goal. We shouldn't move OSes to unsupported that were supported at EOL.

richlander:
PR looks good, other than the w comments I made. Thanks!

Falco20019:
Will do it on Monday since I‘m out of office for the weekend and doing it by mobile isn‘t that confy if that‘s early enough for you :)

richlander:
Monday is more than early enough! I hope you enjoy your time away.

Falco20019:
Thanks for the clarification. I re-checked .NET 7 and the document was correct to the state of EOL (no missing OS in JSON). So I reverted it back to the state before and only regenerated the MD files to fix the missing WinServ 2012(R2) OS that you already considered with #9479.

I also added the exceptions for Debian 11 and 12 to https://github.com/richlander/distroessed/pull/8 to not get confused in the future again :) I moved Debian 11 back to supported and updated the initial post to clarify that.

Falco20019:
@richlander I created two PRs showing how the exceptional versions COULD be maintained in the `supported-os.json` if we want to add it into there. We could of-course also maintain it in a separate JSON, but since the schemas are not finalized yet, it might be better to not duplicate the main-structure.
* https://github.com/Falco20019/core/pull/2
* https://github.com/Falco20019/distroessed/pull/1

richlander (approved):
Thanks!

Issue #9470, opened 2024-08-20
[macOS] Add support for macOS Sequoia


rbhanda:
This issue is created to track progress for the work to enable [macOS sequoia](https://www.apple.com/newsroom/2024/06/macos-sequoia-takes-productivity-and-intelligence-on-mac-to-new-heights/) support for .NET 8.0 and .NET 9.0

### Product and Test updates
- [ ] [All teams] Do a compatibility check for the product with the new OS
- [x] ~#9471~ Runtime does not need RID changes starting with 8.0.
- [ ] #9472
- [x] ~[Infra] Replace docker tags used in builds across repos and branches~ Docker is not applicable to macOS.
- [ ] [CLI team] Update CI legs to latest updated Dockerfile 
- [ ] #9474
- [x] #9473

## Release team:
- [x]  Notify .NET Core teams regarding new OS release
- [ ] Update the supported OS content on GitHub and marketing sites. 

## Document updates:
- [ ] Add OS prerequisites to the dependencies document
- [x] Update distro install page

mikkeljohnsen:
When will support for macOS Sequoia available ?

My app keeps crashing, and not sure this could be the problem. It is working fine on macOS < 15

xtqqczze:
https://github.com/dotnet/docs/issues/42365

adegeo:
Completed in the MS Learn docs install page via https://github.com/dotnet/docs/pull/43045

PR #9547, opened 2024-10-10
Fix OS support according to current EOL information

Falco20019:
Relates to:
* https://github.com/dotnet/core/issues/9523
* https://github.com/dotnet/core/issues/9470
* https://github.com/dotnet/core/issues/9550

Also had a look at:
* https://github.com/dotnet/core/issues/9488 (this is not released yet, so I skipped it for now)

The following things have changed:
* Android released version 15 (added to .NET 6, 8 and 9)
* iOS, iPadOS and tvOS released version 18 (added to .NET 6, 8 and 9)
* iPadOS never had version 12 (removed from .NET 6, was iOS before on iPads)
* macOS released version 15 (added to .NET 6, 8 and 9)
* macOS ended 12 (moved to unsupported in .NET in .NET 6 and 8, removed from .NET 9)
* Windows 11 released version 24H2 (added to .NET 6, 8 and 9)
* Windows 11 ended 22H2 (W) and 21H2 (E) (moved to unsupported in .NET 6 and 8, removed from .NET 9)

Open points:
* Do we really want to have tvOS 12.2, 13 and 14 listed as supported? They are EOS but still supported to compile against. On Android we had a note mentioning the minimum possible target but only listed non-EOS versions as supported.
* Are new versions to be added to .NET 6 (since it's in Maintenance phase) or only EOS versions moved to unsupported?

/CC @richlander 

Maybe tag with the `os-support`


vitek-karas:
@dalexsoto for the Apple related version changes...
@jonpryor for the Android related version changes...

richlander:
Thanks! In general, LGTM.

Questions for the team:

- Are we supporting the new mobile OSes with .NET 6? I'm assuming not.
- It does appear that macOS 12 is EOL and that we should stop supporting it. Have me moved testing to macOS13+?

dalexsoto:
> Thanks! In general, LGTM.
> 
> Questions for the team:
> 
> * Are we supporting the new mobile OSes with .NET 6? I'm assuming not.
> * It does appear that macOS 12 is EOL and that we should stop supporting it. Have me moved testing to macOS13+?

Hello! .NET 6 is no longer supported for mobile as it follows this support policy https://dotnet.microsoft.com/en-us/platform/support/policy/maui

Falco20019:
> Hello! .NET 6 is no longer supported for mobile as it follows this support policy https://dotnet.microsoft.com/en-us/platform/support/policy/maui
> 

Good point, I will then revert the changes to Apple and Android for .NET 6 and only remove the faulty iPadOS 12 entry there.

richlander:
@Falco20019 -- We need to unsupport all Android and iOS versions for .NET 6. The rest looks good. Thanks!

matouskozak:
> * It does appear that macOS 12 is EOL and that we should stop supporting it. Have me moved testing to macOS13+?

_A brownout will take place on November 4, 14:00 UTC - November 5, 00:00 UTC to raise awareness of the upcoming macOS-12 environment removal. For more details, see https://github.com/actions/runner-images/issues/10721_

In dotnet/runtime I think we already moved to `macos-latest` for internal builds and we will shortly have to move the public ones to 13+ as well.

Falco20019:
> @Falco20019 -- We need to unsupport all Android and iOS versions for .NET 6. The rest looks good. Thanks!

I unsupported .NET 6 MAUI and manually adjusted the `Out of support OS versions` to reference the MAUI policy since `distroessed` is currently not knowing about MAUI lifecycle requirements.

xtqqczze:
Should macOS 12 be considered a previously supported OS for .NET 9, given that .NET 9 RC1 was supported from 2024-09-10 to 2024-10-08, but macOS 12 reached its end-of-life on 2024-09-16?

richlander:
> Should macOS 12 be considered a previously supported OS for .NET 9,

Either option works for me. I think that less than a month of support doesn't require us to do anything special. It's more academic than anything.

richlander:
Thanks for the changes.

Falco20019:
In my opinion, we don't need to add those supported before preview too not clutter the JSON and MD files with Out of Support notes already on release. Personally, I would not even add them in the first place during previews and RCs, but that's also fine either way.

PR #9562, opened 2024-10-23
Update windows license information

richlander:
Contributes to: https://github.com/dotnet/runtime/issues/108905

We were given the guidance that the license we were using for `vcruntime140_cor3.dll` was wrong and that we could use ".NET Library License" instead for it.

I also took the liberty to do some cleanup to make this document easier to read.

We can merge when @jkotas approves.

### DIFF
--- a/license-information-windows.md
+++ b/license-information-windows.md
@@ -12,7 +12,11 @@ The following binaries are licensed with the
 * coreclr.dll and .NET runtimes included in binaries published as single-file (due to [extra telemetry](https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/dwreport.cpp) included by .NET runtime in Watson crash reports)
 * Microsoft.DiaSymReader.Native.{x86|amd64|arm|arm64}.dll (used by .NET runtime and SDK)
 * PresentationNative_cor3.dll (used by WPF)
+* vcruntime140_cor3.dll (used by WPF)
 * wpfgfx_cor3.dll (used by WPF)
+
+Note: vcruntime140_cor3.dll is the same binary as
+vcruntime140.dll, which is included in Visual Studio, relicensed under .NET Library License by Microsoft.
 
 The following binaries are licensed with the
 [Windows SDK License](https://learn.microsoft.com/legal/windows-sdk/license):
--- a/license-information-windows.md
+++ b/license-information-windows.md
@@ -9,7 +9,7 @@ This document is provided for informative purposes only, and is not itself a lic
 The following binaries are licensed with the
 [.NET Library License](https://dotnet.microsoft.com/dotnet_library_license.htm)
 
-* coreclr.dll and .NET runtimes included in binaries published as single-file (due to extra telemetry included by .NET runtime in Watson crash reports)
+* coreclr.dll and .NET runtimes included in binaries published as single-file (due to [extra telemetry](https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/dwreport.cpp) included by .NET runtime in Watson crash reports)
 * Microsoft.DiaSymReader.Native.{x86|amd64|arm|arm64}.dll (used by .NET runtime and SDK)
 * PresentationNative_cor3.dll (used by WPF)
 * wpfgfx_cor3.dll (used by WPF)
@@ -20,7 +20,7 @@ The following binaries are licensed with the
 * D3DCompiler_47_cor3.dll (used by WPF)
 
 All other binaries and files are licensed with the
-[.NET Library License](https://dotnet.microsoft.com/dotnet_library_license.htm).
+[MIT license](https://github.com/dotnet/core/blob/main/LICENSE.TXT).
 
 See [license information](./license-information.md) for information about
 other operating systems.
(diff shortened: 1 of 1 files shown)


jkotas:
There is more offline discussion about this

Lorcon1 (approved):
*

license-information-windows.md
```diff
@@ -6,26 +6,13 @@ apply to your use. By using any or all of these files you agree to their associa
...
 * D3DCompiler_47_cor3.dll (used by WPF)
 
-.NET 6 and earlier:
-
-* api-ms-\*.\* (used by .NET runtime)
-* ucrtbase.dll (used by .NET runtime)
-
 All other binaries and files are licensed with the
-[MIT license](https://github.com/dotnet/core/blob/main/LICENSE.TXT).
+[.NET Library License](https://github.com/dotnet/core/blob/main/LICENSE.TXT).
```
jkotas, line 15: The link still points to MIT

richlander: Oops. Not sure how that happened. I had intended to make that change. Will resolve.

license-information-windows.md
```diff
@@ -2,30 +2,25 @@
...
-[Windows SDK License](https://learn.microsoft.com/legal/windows-sdk/license):
-
-* api-ms-\*.\* (used by .NET runtime, .NET 6 and earlier)
-* ucrtbase.dll (used by .NET runtime, .NET 6 and earlier)
-* D3DCompiler_47_cor3.dll (used by WPF)
-
 The following binaries are licensed with the
 [.NET Library License](https://dotnet.microsoft.com/dotnet_library_license.htm)
 
+* coreclr.dll and .NET runtimes included in binaries published as single-file (due to extra telemetry included by .NET runtime in Watson crash reports)
```
jkotas, line 12: We also need to include `vcruntime140_cor3.dll (used by WPF)` that started all this discussion in the .NET Library License list. For extra clarity, we may add a note that it is same binary as `vcruntime140.dll` included in Visual Studio redist relicensed under .NET Library License with Microsoft permission.

jkotas, line 12:
```suggestion
* coreclr.dll and .NET runtimes included in binaries published as single-file (due to [extra telemetry](https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/dwreport.cpp) included by .NET runtime in Watson crash reports)
```
Link to the code for more clarity

license-information-windows.md
```diff
@@ -2,30 +2,29 @@
...
-[Windows SDK License](https://learn.microsoft.com/legal/windows-sdk/license):
-
-* api-ms-\*.\* (used by .NET runtime, .NET 6 and earlier)
-* ucrtbase.dll (used by .NET runtime, .NET 6 and earlier)
-* D3DCompiler_47_cor3.dll (used by WPF)
-
 The following binaries are licensed with the
 [.NET Library License](https://dotnet.microsoft.com/dotnet_library_license.htm)
 
+* coreclr.dll and .NET runtimes included in binaries published as single-file (due to [extra telemetry](https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/dwreport.cpp) included by .NET runtime in Watson crash reports)
```
jkotas, line 12:
```suggestion
* coreclr.dll and .NET runtimes included in binaries published as single-file (due to [extra telemetry](https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/dwreport.cpp) included by .NET runtime in Windows Error Reporting crash reports)
```
Use official name for Watson. It turns Watson crash reporting is something completely different: https://www.datadriven.com/crash-reporting-software/

jkotas (approved):
LGTM!

license-information-windows.md
```diff
@@ -2,30 +2,29 @@
...
-[Windows SDK License](https://learn.microsoft.com/legal/windows-sdk/license):
-
-* api-ms-\*.\* (used by .NET runtime, .NET 6 and earlier)
-* ucrtbase.dll (used by .NET runtime, .NET 6 and earlier)
-* D3DCompiler_47_cor3.dll (used by WPF)
-
 The following binaries are licensed with the
 [.NET Library License](https://dotnet.microsoft.com/dotnet_library_license.htm)
 
+* coreclr.dll and .NET runtimes included in binaries published as single-file (due to [extra telemetry](https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/dwreport.cpp) included by .NET runtime in Windows Error Reporting crash reports)
```
js6pak, line 12: Why? The linked `dwreport.cpp` file seems to be MIT-licensed?

richlander: The code itself is MIT. The resultant binary is licensed as .NET Library license because of data collection, per the comment. .NET app crashes on Windows generate memory dumps that get collected by Watson. Those dumps are used to fix bugs in .NET. Many Microsoft products are integrated with Watson. I don't know how many product fixes are made based on Watson information but it has to be a lot. .NET Framework is the same.

There is a clause in the .NET Library License about data collection. The .NET SDK is the other component that I'm aware of that relies on this clause. One can certainly imagine using MIT with software that has a data collection scenario. We decided not to do that.

I remember using Office when I was a kid and seeing the "Dr. Watson" dialog. Yes, Sherlock Holmes themed. Office was the originator of Watson.

PR #9563, opened 2024-10-23
Correct architecture support for Ubuntu 24.04

leecow:
.NET arm32 is not supported on Ubuntu 24.04. See https://github.com/dotnet/runtime/issues/101444

@richlander @rbhanda


richlander:
FYI: this will just get deleted the next time I generate the markdown files.

PR #9641, opened 2024-11-29
Update support statements

Falco20019:
Follow-up for #9604

I removed iOS 15 and iPadOS 15 because they ended before .NET 9 was released (like @richlander did with Ubuntu 20.04). Feel free to instead add it to the `unsupported-versions` section if preferred.

Please check if the versions marked as supported are really already supported. I only updated the JSON now as discussion base based on what the currently available and non-EOL OS are.

* .NET 8
  * Moved from `supported-versions` to `unsupported-versions`
    * Alpine 3.17
    * Fedora 39
    * SUSE Enterprise Linux 12.5
    * iOS 15
    * iPadOS 15
* .NET 9
  * Removed from `supported-versions` since EOL before release
    * iOS 15
    * iPadOS 15
* ~~.NET 8 + .NET 9~~
  * ~~Added to `supported-versions`~~
    * ~~Fedora 41~~
    * ~~Windows Server 2025~~
    * ~~Windows Server Core 2025~~
    * ~~Nano Server 2025~~

/CC @richlander @sbomer @leecow

PR #9647, opened 2024-12-06
Add .NET 10 support information

richlander:
We are building up infra for .NET 10. We need to know what the min supported version is for each OS/distro.

Please comment if these versions should change, now or in the future. I've made some initial changes w/rt to .NET 8 and 9. It's all open for discussion. Note that the versions should be based on what makes sense to support as of late 2025 not late 2024 (now).

Compare with:

- https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md
- https://github.com/dotnet/core/blob/main/release-notes/9.0/supported-os.md

Note: We're moving to two digits (before the `.`). I'm sure that will break someone's assumptions in their code.

Related: https://github.com/dotnet/runtime/issues/109939

@Falco20019 @leecow @rbhanda

### DIFF
--- a/release-notes/10.0/supported-os.json
+++ b/release-notes/10.0/supported-os.json
@@ -1,0 +1,356 @@
+{
+    "channel-version": "9.0",
+    "last-updated": "2024-12-06",
+    "families": [
+        {
+            "name": "Android",
+            "distributions": [
+                {
+                    "id": "android",
+                    "name": "Android",
+                    "link": "https://www.android.com/",
+                    "lifecycle": "https://support.google.com/android",
+                    "architectures": [
+                        "Arm32",
+                        "Arm64",
+                        "x64"
+                    ],
+                    "supported-versions": [
+                        "15",
+                        "14",
+                        "13",
+                        "12.1",
+                        "12"
+                    ],
+                    "notes": [
+                        "API 21 is used as the minimum SDK target."
+                    ]
+                }
+            ]
+        },
+        {
+            "name": "Apple",
+            "distributions": [
+                {
+                    "id": "ios",
+                    "name": "iOS",
+                    "link": "https://developer.apple.com/ios/",
+                    "architectures": [
+                        "Arm64"
+                    ],
+                    "supported-versions": [
+                        "18",
+                        "17",
+                        "16"
+                    ],
+                    "notes": [
+                        "iOS 12.2 is used as the minimum SDK target."
+                    ]
+                },
+                {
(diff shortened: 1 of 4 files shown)


release-notes/10.0/supported-os.json
```diff
@@ -0,0 +1,356 @@
+{
+    "channel-version": "9.0",
```
WeihanLi, line 2: Should this be `10.0`?

richlander: Ooops! Good catch.

```suggestion
    "channel-version": "10.0",
```

matouskozak (approved):
LGTM!

The only change from .NET 9 looks to be the cut of iOS 15 support which should be already out of support. @dalexsoto do you know if we should expect iOS 16 to get OOF by late 2025?

@jonpryor is there anything from Android side that we should change?

dalexsoto (approved):

richlander:
I can removed iOS 15 from earlier releases as a separate PR. Good call.

Falco20019:
@richlander My PR #9641 would already remove them. Should I do some more adjustments there or is it ready for merging?

Falco20019:
@richlander Just out of curiosity: .NET 11 (expected ~Nov 10th 2026) is going to drop Windows Server 2012 (+R2) since ESU is expected [to end in October 2026](https://learn.microsoft.com/de-de/lifecycle/products/windows-server-2012-r2), right?

richlander:
Those assumptions seem sound. In general, if an OS has <6 months of support left when a new .NET release ships, we don't add it to our support list (for that .NET release). That's why Alpine 3.20 isn't listed as supported for .NET 10, even though Alpine 3.20 is a currently supported release.

I don't have a lot of depth on the ESU program. I expect that we'll continue to list ESU releases in supported-os.json/md, like you say.

> With some LTS being longer than the ESU anyway.

I'm not sure what you mean by that.

Falco20019:
> > With some LTS being longer than the ESU anyway.
> 
> I'm not sure what you mean by that.

The ESU is a paid program. So people can pay to continue using `Windows 10 22H2 ESU` per year for up to 3 years (until 2028 latest) once it reached general End-of-Support in 2025. Or they could downgrade to/stay on `Windows 10 21H2 Enterprise LTS` which has support until 2027. Or downgrade even more to `Windows 10 1809 Enterprise LTS` which has support until 2029 without paying (even longer than the 3 paid years of 22H2).

So yeah, talking about straight-forward licensing :)

PR #9665, opened 2024-12-17
Fix broken links and update

Falco20019:
@richlander I found that some of the links are broken and that .NET 6 was still shown as LTS although EOL.


richlander (approved):

richlander:
Thanks!

PR #9666, opened 2024-12-17
Moving 9.0.200 preview to 9.0.0 family

rbhanda:


richlander:
Can we use this as an opportunity to come up with a format that makes SDKs easy to discover?

There are two patterns for that:

- Add links from `9.0.0.md` to each of the SDKs.
- Add a `README.md` that links to each of the relevant releases.

I think the second one might actually be the easiest and most intuitive.

richlander (approved):

Issue #9638, opened 2024-11-28
.NET OS Support Tracking


richlander:
.NET is supported on multiple operating systems, which have their own releases and lifecycles. "Incoming" and "Outgoing" releases are tracked in this issue. Supported releases are documented at [.NET Supported OS Policy](https://github.com/dotnet/core/blob/main/os-lifecycle-policy.md).

## Incoming

"Incoming" releases will soon be supported.

- https://github.com/dotnet/core/issues/9583
- https://github.com/dotnet/core/issues/9577
- https://github.com/dotnet/core/issues/9523
- https://github.com/dotnet/core/issues/9488
- https://github.com/dotnet/core/issues/9470
- https://github.com/dotnet/core/issues/9374

## Outgoing

"Outgoing" will soon be dropped from support.

- https://github.com/dotnet/core/issues/9625
- https://github.com/dotnet/core/issues/9571
- https://github.com/dotnet/core/issues/9564
- https://github.com/dotnet/core/issues/9549
- https://github.com/dotnet/core/issues/9548

PR #9679, opened 2025-01-07
Remove iOS 16+17 and openSUSE Leap 15.5

Falco20019:
Relates to #9638

I removed iOS 16+17 because they ended before .NET 10 was released. iPadOS 16 and 17 is still supported by Apple.

Please check if the versions marked as supported are really already supported. I only updated the JSON now as discussion base based on what the currently available and non-EOL OS are.

* .NET 8 + 9
  * Moved from supported-versions to unsupported-versions
    * openSUSE Leap 15.5
    * iOS 16 + 17
* .NET 10
  * Removed from supported-versions since EOL before release
    * iOS 16 + 17

/CC @richlander @sbomer @leecow


richlander:
FYI @jonathanpeppers @Redth

jonathanpeppers:
@rolfbjarne and @dalexsoto can comment on the support status of iOS 16 and 17.

It seems like we probably support iOS 12.2 and higher, though?

* https://learn.microsoft.com/dotnet/maui/supported-platforms

richlander:
This isn't a question of compiler targeting, but what we commercially support. Apple doesn't support users running iOS 12. 

This same distinction is clarified in this recent issue: https://github.com/dotnet/runtime/issues/109939

rolfbjarne:
> Context: [endoflife.date/ios](https://endoflife.date/ios)

I'm not sure this is the best place to check what Apple supports or not (no idea what would be though), because it seems older iOS can get resurrected.

Example: iOS 15 lost active support in September 2022, but still got a security update in July 2024, almost two years later.

Also removing iOS 17 seems rather harsh, since iOS 18 is only a few months old. I wouldn't be surprised if Apple decides to skip new security updates to iOS 17 in the future.

Dropping iOS 16 seems fair though, a year has passed since last time.

Falco20019:
According to the discussion on https://github.com/endoflife-date/endoflife.date/issues/6444 there are known unresolved CVEs for iOS 17 and Apple seems to push people towards 18 instead of fixing it. iPadOS 16 also might be EOL (which I didn't include right now).

Apple is never officially setting anything EOL, so checking when something is last updated is usually the only real indicator. The only really EOL devices and not the OS.

Redth:
I'm not seeing any supporting references on this end of life website, not sure where it gets its source of truth from, and I'd be very hesitant to base our decisions on it for iOS/Android.  I don't think either Apple or Google do a great job of articulating which OS versions are actually 'supported'.  

### iOS 
As @rolfbjarne mentioned, Apple can still push out service releases for older versions and I'm not sure they actually declare anything officially as to which versions are 'supported'.  The only thing I can really find from Apple is: [Note: Because of dependency on architecture and system changes to any current version of Apple operating systems (for example, macOS 15, iOS 18, and so on), not all known security issues are addressed in previous versions (for example, macOS 14, iOS 17, and so on).](https://support.apple.com/en-ca/guide/deployment/depc4c80847a/web).  The reality is there's a not insignificant amount of devices on older iOS versions and our customers expect we reasonably support the majority of the addressable market.  

### Android
I'm less certain about Google, but there's security patch dates in Android now, and I think vendors generally decide how long they support a given OS / device for, so the decision I believe is less up to Google in this case.  For their Pixel devices the support time frame has changed over the years, some had 3 years, it looks like starting with Pixel 8, Google is promising 7 years of security updates - however they don't state which Android OS versions this will apply to.  There's also updates to things like AndroidX and Google Play Services which arguably make the concept of a support OS version even more ambiguous.

-------------------------

I think the end result here is we need to make some reasonable decisions based on what the runtimes are capable of and what the market share / distribution looks like for the platforms.

For .NET 10, I think my recommendation would be:

- iOS: 15+
- Android: 23+

However, I think there are other stakeholders that should have input to the final decision there too.

@steveisok @dalexsoto @jonpryor @davidortinau and any others ya'll would suggest?

Falco20019:
> I also thought that removing iOS 17 seemed quick. Can app devs still upload apps for iOS 17?

Starting 2025 Apple will limit store apps to target using iOS 18 SDK. Most probably limiting it to Xcode 16.

While Xcode 16 still supported deploying for iOS 12, Xcode 16.1 already dropped it and moved up to iOS 13.

Falco20019:
@Redth iOS 15 was already removed from .NET 8+9 in the past, so 16+ would be the better fit then. Android is currently on 12+ which according to https://apilevels.com/ means 68.7% of all devices (cummulative usage), while 13+ means 55.6%.

I would therefore adjust the change to the following instead if that matches your expectations:
* .NET 8 + 9
  * Moved from supported-versions to unsupported-versions
    * openSUSE Leap 15.5
    * iOS 16
    * iPadOS 16
* .NET 10
  * Removed from supported-versions since EOL before release
    * iOS 16
    * iPadOS 16

jonttm:
Hello, I give some context from https://github.com/endoflife-date/endoflife.date/issues/6444

Apple doesn't publish any information about what versions they support but we can learn from history. The newest version of iOS receives security updates for all devices. The previous version only receives security updates for all devices until December. Beyond December, previous versions only receive security updates for devices that cannot update to the newest version. All devices on iOS 17 are capable of updating to iOS 18, therefore iOS 17 will not receive support. iOS 16 might receive a future surprise update because the iPhone X and 8 are stuck on it.

Falco20019:
> I would therefore adjust the change to the following instead if that matches your expectations:
> 
> * .NET 8 + 9
>   * Moved from supported-versions to unsupported-versions
>     * openSUSE Leap 15.5
>     * iOS 16
>     * iPadOS 16
> * .NET 10
>   * Removed from supported-versions since EOL before release
>     * iOS 16
>     * iPadOS 16

I applied them as discussed.

richlander (approved):

richlander:
I'm sorry these are taking so long to merge. I'm hoping we can get faster at this.

Issue #9675, opened 2024-12-31
Use `builds.dotnet.microsoft.com` CDN in releases.json files


richlander:
`releases.json` and `release.json` files currently reference the VS CDN. That's a bad practice since it requires that both `builds.dotnet.microsoft.com` and `download.visualstudio.microsoft.com` are up. Is either is down, then users cannot download files via this flow. That naturally means that the JSON files should reference files on the `builds.dotnet.microsoft.com` CDN. Also, the VS CDN doesn't have stable URLs, making it impossible to have a stable link to the JSON files should we want to locate them there instead.

Example JSON file:

https://github.com/dotnet/core/blob/218ef74621d76d73c0a562893df875d37ce928ad/release-notes/8.0/releases.json#L23-L33

We are currently validating that the Azure Front Door and Akamai CDNs that we'll be using provide good latency/service to various large countries.

This change also means that we will no longer be publishing new URLs for the VS CDN. It will be used as a private implementation detail for .NET delivery via Visual Studio. All users will need to move to using the dotnet CDN.

FYI: We're going to call `builds.dotnet.microsoft.com` the "dotnet CDN". Feel free to use that terminology. We intend for it to have a very long future.

agocke:
It's actually worse than that: the current behavior is inconsistent and, as far as I can see, cannot be made consistent.

What I mean by consistent is the data in one of the JSON files shouldn't be contradicted by the links and data in referenced files. So, for instance, a URL in the original JSON file should never 404. There may be network issues that prevent delivery, but a 404 would indicate that the content is inconsistent.

As far as I can see, we may have some limited ability to ensure consistency in one CDN. For instance we could stage our releases-index.json such that the referenced URLs are uploaded before we publish the index.

But across CDNs I think this is just impossible. So, an example where this is broken is the CDN-delivered releases-index.json file. The file at https://builds.dotnet.microsoft.com/dotnet/release-metadata/releases-index.json looks like

```json
{
    "$schema": "https://json.schemastore.org/dotnet-releases-index.json",
    "releases-index": [
        {
            "channel-version": "9.0",
            "latest-release": "9.0.0",
            "latest-release-date": "2024-12-03",
            "security": true,
            "latest-runtime": "9.0.0",
            "latest-sdk": "9.0.101",
            "product": ".NET",
            "support-phase": "active",
            "eol-date": "2026-05-12",
            "release-type": "sts",
[...]

PR #9706, opened 2025-01-30
Switch links to use new domain

richlander:
Related: #9675

This is a test to determine if the new CDN has better or worse latency for users. Please share feedback.


leecow (approved):
Looks good for the test.

PR #9707, opened 2025-01-31
Add package servicing doc

richlander:


release-policies.md
```diff
@@ -55,22 +57,13 @@ Additionally, customers targeting an out-of-support .NET version in their projec
...
-* A new version of the package is available. Some packages support multiple patches versions at once, one per major version matching a supported major .NET version.
-* The package exclusively includes implementations for out-of-support .NET versions.
-
-Nuget.org includes [version](https://www.nuget.org/packages/System.Text.Json/#versions-body-tab) and [supported framework](https://www.nuget.org/packages/System.Text.Json/#supportedframeworks-body-tab) information that can be used to determine support status for packages.
-
 ## Support requirements
 
 To remain supported, you must do the following:
 
 * Use a supported SDK
```
artl93, line 64: Suggest linking to [the .NET landing page](https://dotnet.microsoft.com/en-us/download) so that customers can see which versions are currently in support.

Documentation/policies/package-servicing.md
```diff
@@ -0,0 +1,50 @@
...
+- [System.Text.Json](https://www.nuget.org/packages/System.Text.Json)
+- [System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource)
+
+## Support
+
+Users must be on the latest patch version of the package or a runtime-band to be supported.
+
+## Compatibility
+
+Packages retain compatibility for supported .NET versions. A package will not drop support for a .NET version that is in support. When a .NET version goes out of support, packages released after that point may drop support for that .NET version.
```
maddymontaquila, line 39: We aren't sure this is going to be true for Aspire 10 and .NET 8 yet, and we won't be making that call for a bit, so a disclaimer is needed.

maddymontaquila (approved):
just the one comment and otherwise lgtm!

Documentation/policies/package-servicing.md
richlander:
```suggestion
In general, packages retain compatibility for supported .NET versions. A package will not drop support for a .NET version that is in support. When a .NET version goes out of support, packages released after that point may drop support for that .NET version.

Some packages may have a different policy, which will be separately documented.
```

ericstj: I think Aspire must retain compatibility and serviceability - either through branching or multi-targeting.  It cannot break the LTS promise otherwise folks can't use it and expect to stay on LTS.  cc @eerhardt @joperezr 

I agree it's probably nuanced - ASP.NET does the branching thing rather than multi-targeting + compatibility.

@richlander do you think these two modes (branching vs multi-targeting) are important enough to call out here?

richlander: Yes. I think we can add those. On one hand, it is an implementation detail. On the other hand, I think it is useful to explain how we achieve this so that others can understand and repeat the pattern. We're not doing any special or magic. It's just well-trodden patterns.

Documentation/policies/package-servicing.md
```diff
@@ -0,0 +1,50 @@
...
+
+Some packages match a library (in name and content) in the runtime. In this case, the team publishes  fixes for the library in the runtime and the package at the same time. The version numbers will match. Users can upgrade to the new latest patch version to get the fix.
+
+Users can upgrade to the latest library version (like 8.0.x) that matches a supported runtime (like 8.0). Users can also move to the latest patch of a newer major version (like 9.0.x) provided that the runtime is still in support.
+
+The advantage of matching versions is that the runtime library will be used in place of the package, resulting in smaller deployments and servicing being provided by the runtime. The advantage of using a newer package version is getting access to newer functionality.
+
+Apps only ever need a package reference to one of these libraries to use a newer major version. It is more common for libraries to have a reference, particularly .NET Standard libraries.
+
+The following libraries use this servicing model:
```
eerhardt, line 27: Should we add EF Core here?

cc @AndriySvyryd @SamMonoRT

eerhardt, line 27: Would it help to add dotnet/aspnetcore packages here as well? For example, `Microsoft.AspNetCore.SignalR.Client` and `Microsoft.Extensions.Diagnostics.HealthChecks`

Documentation/policies/package-servicing.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Latest version servicing
+
+The default package servicing model is “latest version”. The team publishes an updated package with a security and/or functionality fix. Users can upgrade to the new latest version to get the fix.
+
+The following libraries use this servicing model:
+
+- [.NET Aspire](https://github.com/dotnet/aspire)
+- [ML.NET](https://github.com/dotnet/machinelearning)
+- [Maintenance-packages](https://github.com/dotnet/maintenance-packages)
```
eerhardt, line 15: It would be nice to add dotnet/extensions here as well.

One recurring point of confusion is that we have `Microsoft.Extensions.*` packages with different support models based on the repo it comes from. For MS.Ext libraries coming from runtime and aspnetcore, it uses Runtime-band servicing. For MS.Ext libraries in dotnet/extensions, it uses "Latest version servicing".

Documentation/policies/package-servicing.md
richlander: The lists are not intended to be exhaustive. However, 3-6 examples would be great. Feel free to add in more via suggestion. It seems like an exhaustive list should be provided via a different approach.

Documentation/policies/package-servicing.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Runtime-band servicing
+
+Some packages match a library (in name and content) in the runtime. In this case, the team publishes  fixes for the library in the runtime and the package at the same time. The version numbers will match. Users can upgrade to the new latest patch version to get the fix.
+
+Users can upgrade to the latest library version (like 8.0.x) that matches a supported runtime (like 8.0). Users can also move to the latest patch of a newer major version (like 9.0.x) provided that the runtime is still in support.
+
+The advantage of matching versions is that the runtime library will be used in place of the package, resulting in smaller deployments and servicing being provided by the runtime. The advantage of using a newer package version is getting access to newer functionality.
+
+Apps only ever need a package reference to one of these libraries to use a newer major version. It is more common for libraries to have a reference, particularly .NET Standard libraries.
```
BrennanConroy, line 25: This statement was confusing had to read it many times to understand. Maybe being more explicit about the user writing a library (particularly with a .NET Standard target) will have explicit package references.

Documentation/policies/package-servicing.md
BrennanConroy: Idk about HealthChecks, but SignalR.Client is not part of the shared framework, so it always requires a package reference.

Documentation/policies/package-servicing.md
eerhardt, line 15:
```suggestion
- [.NET Aspire](https://github.com/dotnet/aspire)
- [.NET Platform Extensions](https://github.com/dotnet/extensions)
- [ML.NET](https://github.com/dotnet/machinelearning)
- [Maintenance-packages](https://github.com/dotnet/maintenance-packages)
```

Documentation/policies/package-servicing.md
```diff
@@ -0,0 +1,50 @@
...
+
+The following libraries use this servicing model:
+
+- [Microsoft.Extensions.DependencyInjection](https://www.nuget.org/packages/Microsoft.Extensions.DependencyInjection)
+- [System.Text.Json](https://www.nuget.org/packages/System.Text.Json)
+- [System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource)
```
eerhardt, line 31:
```suggestion
- [Microsoft.Extensions.DependencyInjection](https://www.nuget.org/packages/Microsoft.Extensions.DependencyInjection)
- [System.Text.Json](https://www.nuget.org/packages/System.Text.Json)
- [System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource)
- [ASP.NET Core Packages](https://github.com/dotnet/aspnetcore)
- [Entity Framework Core Packages](https://github.com/dotnet/efcore)
```

Documentation/policies/package-servicing.md
eerhardt: One idea we had last week in a meeting in this area was:

1. Mark old package versions as "deprecated" on nuget.org. (for example, all System.Text.Json 7.0.x versions would be marked as deprecated today.)
2. Add tooling to let you know when you are using something is not supported any longer.

This way, you could see that the 8.x and 9.x versions were still in support. But the 7.x version isn't.

For "Latest version servicing" like .NET Aspire all 8.x versions would be marked as "deprecated".

eerhardt: > exhaustive list 

I'm not suggesting this list be exhaustive of all packages. What I am suggesting is to document the _groups_ of packages that follow the same model. All Runtime, ASP.NET Core, EF Core follow this model. All Extensions, Aspire, ML.NET follow latest version model.

Documentation/policies/package-servicing.md
richlander: I agree that it is confusing. Better?

```suggestion
It common for .NET Standard libraries to have a reference to these libraries. Libraries and app projects only need a reference if they need access to newer features in a later major version (than their target framework exposes).
```

Documentation/policies/package-servicing.md
richlander: I would love for support level to be exposed via the NuGet gallery and the NuGet API/client. That's clearly the best model.

@OliaG @aortiz-msft

Documentation/policies/package-servicing.md
richlander:
```suggestion
Packages retain compatibility for supported .NET versions. A package will not drop support for a .NET version that is in support. When a .NET version goes out of support, packages released after that point may drop support for that .NET version.

Mechanically, a package can support multiple target frameworks via either of two approaches:

- Multi-targeting: A project supports and is built multiple target frameworks, resulting in assets supporting multiple .NET versions in a NuGet package.
- Branching: A project exists and is built in multiple branches, each targeting a specific target framework and result in a single asset supporting a specific .NET version.

The differences in these approaches are user observable and require users to adopt different patterns with package references. They are also patterns that users can adopt for themselves with their own packages.
```

eerhardt: Is this documentation for library/package authors? Or for consumers of our packages?

Documentation/policies/package-servicing.md
AndriySvyryd: > The advantage of matching versions is that the runtime library will be used in place of the package, resulting in smaller deployments and servicing being provided by the runtime.

This does not apply to EF, perhaps there should be a subsection for packages in the shared framework.

Documentation/policies/package-servicing.md
```diff
@@ -0,0 +1,53 @@
...
+## Compatibility
+
+Packages retain compatibility for supported .NET versions. A package will not drop support for a .NET version that is in support. When a .NET version goes out of support, packages released after that point may drop support for that .NET version.
+
+## End of Life
+
+.NET packages are no longer supported when either of the following occurs:
+
+- A new patch version of the package is available.
+- The package exclusively includes implementations for out-of-support .NET versions.
```
AndriySvyryd, line 49: Is a package still supported if one of its (transitive) dependencies is not?

@GrabYourPitchforks

Documentation/policies/package-servicing.md
ericstj: This doc cannot become a complex map of all the packages and how they are different.  It needs to describe some basic principles that folks can understand and get value from.

Documentation/policies/package-servicing.md
ericstj: I think that would be a problem for that package to fix - more like an SDL issue for that package.  Not something the customer is supposed to worry about.

Documentation/policies/package-servicing.md
ericstj: I think this is OK, though I hesitate to add this level of detail.  The principle we need to communicate is that we have a promise with a package that installs on a supported framework - that package, or a compatible package version, will continue to be supported as long as that framework is supported.

ericstj (approved):
It's a difficult topic that will always have more we can say.  I think this does a good job of giving an outline without trying to explain the world.

Documentation/policies/package-servicing.md
AndriySvyryd: Sure. But right now it's factually incorrect.

richlander: Feel free to write a doc for EF (in the EF repo) and we can link to it.

aortiz-msft: > I would love for support level to be exposed via the NuGet gallery and the NuGet API/client. That's clearly the best model. That's a good idea to follow merging this PR.
> 
> @OliaG @aortiz-msft

@richlander Long thread here. Would you please summarize the ask for us and open an issue in NuGet/Home? 

cc @OliaG

richlander: Ask: There should be an E2E for deprecated/unsupported packages from producer marking existing packages as such and consumers getting warnings via the gallery, VS, and the NuGet/dotnet CLI. A team like ours isn't going to use the gallery to mark our packages as deprecated. We need an API for that (we're talking to @joelverhagen about that part). I can open an issue if there isn't one on that.

AndriySvyryd: https://github.com/NuGet/Home/pull/13598 ?

richlander: Yes. That appears to be a match and I remember it now. Good find.

FYI: @JonDouglas

PR #9709, opened 2025-02-01
Update Linux versions

richlander:
@Falco20019 @siramvikram 

FYI: I've been forgetting to update the `last-updated` date. Good to remember for later. I did it this time.

- Using fixes from: https://github.com/richlander/distroessed/pull/13
- I have "format on save" enabled in VS Code. Some of the diffs are from that.

### DIFF
--- a/release-notes/10.0/supported-os.md
+++ b/release-notes/10.0/supported-os.md
@@ -1,18 +1,15 @@
 # .NET 10 - Supported OS versions
 
-Last updated: 2025-01-31
+Last Updated: 2025-02-03
+Support phase: Preview
 
-[.NET 10](README.md) is a [Long Term Support (LTS)](../../release-policies.md) release and [is supported](../../support.md) on multiple operating systems per their lifecycle policy.
-
-Note: .NET 10 is currently in development and not supported. Supported OS versions are subject to change.
+[.NET 10](README.md) is an [LTS](../../release-policies.md) release and [is supported](../../support.md) on multiple operating systems per their lifecycle policy.
 
 This file is generated from [supported-os.json](supported-os.json) and is based on support information from [endoflife.date](https://endoflife.date/).
 
 ## Android
 
-OS                              | Versions                    | Architectures         | Lifecycle
-------------------------------- | --------------------------- | --------------------- | ----------------------
-[Android][0]                    | 15, 14, 13, 12.1, 12        | Arm32, Arm64, x64     | [Lifecycle][1]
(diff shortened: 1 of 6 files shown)


release-notes/10.0/supported-os.md
```diff
@@ -47,9 +47,9 @@ Notes:
 OS                              | Versions                    | Architectures         | Lifecycle
 ------------------------------- | --------------------------- | --------------------- | ----------------------
 [Alpine][6]                     | 3.21                        | Arm32, Arm64, x64     | [Lifecycle][7]
-[Azure Linux][8]                | 3.0                         | Arm64, x64            | None
```
Falco20019, line 50: Seems like distroessed changed here? Shouldn't make a hard difference.

Falco20019 (approved):
Sorry, I also forgot to change the last update timestamp on my PRs. As there might be some delays between PR creation and merge, we might want to do it briefly before merging.

Maybe we should also note in the Schema that it should be checked by comparing for IF it changed since last checked and not that the time is greater when the real time used when checked. Otherwise one might miss changes as the timestamp might be "before checked" at a time after one checked it.

release-notes/10.0/supported-os.md
richlander: Looks like some logic got lost in the tool. It needs to be updated. I can do that later. Good catch.

richlander: It is fixed now.

leecow (approved):
Looks good

release-notes/9.0/supported-os.md
```diff
@@ -42,81 +43,82 @@ Notes:
...
 
-OS                      | Version       | Date
------------------------ | ------------- | ----------------------
-iOS                     | 16            | [2024-08-07](https://support.apple.com/HT213407)
-iPadOS                  | 16            | [2024-08-07](https://developer.apple.com/documentation/ios-ipados-release-notes/ipados-16-release-notes)
-openSUSE Leap           | 15.5          | 2024-12-31
+| OS                    | Version       | Date                 |
+| --------------------- | ------------- | -------------------- |
+| iOS                   | 16            | [2024-08-07](https://support.apple.com/HT213407) |
+| iPadOS                | 16            | [9999-12-31](https://developer.apple.com/documentation/ios-ipados-release-notes/ipados-16-release-notes) |
```
richlander, line 123: Looks like this bad date slipped through @Falco20019. We'll have to address that in a distroessed update.

PR #9749, opened 2025-02-18
Update supported-os.md

jamesmontemagno:


richlander:
We've had this conversation a few times. This document about supported OSes, not min SDKs. I know that's not how the mobile team thinks about it, hence the recurring conversation. Min SDKs isn't really a support concept. You can see, however, that we document Linux "min SDKs" in that same file.

Also, we generated the .md file from the .json file.

release-notes/9.0/supported-os.md
```diff
@@ -23,11 +23,11 @@ Notes:
...
 | OS                            | Versions                    | Architectures         | Lifecycle            |
-| ----------------------------- | --------------------------- | --------------------- | -------------------- |
-| [iOS][2]                      | 18, 17                      | Arm64                 | None                 |
-| [iPadOS][3]                   | 18, 17                      | Arm64                 | None                 |
-| [macOS][4]                    | 15, 14, 13                  | Arm64, x64            | None                 |
-| [tvOS][5]                     | 18, 17, 16, 15, 14, 13, 12.2 | Arm64                | None                 |
+| ----------------------------- | ---------------------------- | --------------------- | -------------------- |
+| [iOS][2]                      | 12.2 to 18                   | Arm64                 | None                 |
+| [iPadOS][3]                   | 12.2 to 18                   | Arm64                 | None                 |
+| [macOS][4]                    | Venture (13) to Sequoia (15) | Arm64, x64            | None                 |
```
paulirwin, line 29: There is a macOS version typo here: "Venture (13)" should be "Ventura (13)"

jamesmontemagno:
@richlander  gotcha!

Android has the concept of API Level and OS version.... they don't align unlike other platforms. I'll ping ya Rich on this on teams

richlander:
I think that API level and OS version is the same as glib version that we have with Linux.

PR #9769, opened 2025-02-26
.NET 10 Preview 2 Release Notes

jamesmontemagno:


release-notes/10.0/preview/preview2/sdk.md
```diff
@@ -0,0 +1,24 @@
+# .NET SDK in .NET 10 Preview 2 - Release Notes
```
Youssef1313, line 1: @mariam-abdulla We should make sure to announce the new `dotnet test` experience as part of SDK release notes. Not sure if we should do that for preview 2 release notes, or wait for preview 3 when it's more tested and stabilized.

mariam-abdulla: I think it's better we wait until it's more test and stabilized.

jamesmontemagno: If it is something shipping in P2 then put it in p2 and add more in p3. These are point in time of what is in the release and not announcements.

PR #9823, opened 2025-03-27
.NET 10 Preview 3 - Runtime

jamesmontemagno:


AndyAyersMS:
@amanasifkhalid are you planning on adding the codegen notes?

amanasifkhalid:
@AndyAyersMS yes, I'll have something up shortly...

release-notes/10.0/preview/preview3/runtime.md
```diff
@@ -8,6 +8,94 @@ Here's a summary of what's new in the .NET Runtime in this preview release:
...
+       add      rsp, 32
+       pop      rbx
+       pop      r15
+       pop      rbp
+       ret
+```
+
+For more information on stack allocation improvements in the JIT compiler, check out [dotnet/runtime #108913](https://github.com/dotnet/runtime/issues/108913).
+
+## Improved Code Layout
```
amanasifkhalid, line 90: @AndyAyersMS I'm not sure how we want to highlight this work externally. Right now, it reads pretty bookish, though I'm not sure if there's a way around that.

release-notes/10.0/preview/preview3/runtime.md
```diff
@@ -8,6 +8,94 @@ Here's a summary of what's new in the .NET Runtime in this preview release:
...
+{
+    string[] words = {"Hello", "World!"};
+    foreach (var str in words)
+    {
+        Console.WriteLine(str);
+    }
+}
+```
+
+The lifetime of `words` is scoped to the `Print` method, and the JIT can avoid heap-allocating the strings `"Hello"` and `"world!"` already. However, the fact that `words` is an array of `strings`, a reference type, would previously stop the JIT from stack-allocating it. Now, the JIT can eliminate every heap allocation in the above example. At the assembly level, the code for `Print` used to look like this:
```
AndyAyersMS, line 26: I believe the strings are actually heap allocated here, but that happens when the method is JITed, and their addresses are stable and can be incorporated into the code.

release-notes/10.0/preview/preview3/runtime.md
AndyAyersMS: If you want to keep this at a higher level we could create a benchmark with similar pattern and use the `[MemoryDiagnoser]` to show there is now less allocation.

Though it might also run a bit slower since we're now paying to zero init the array in the method rather than getting zeroed memory "for free" from the heap. Structuring the benchmark so it properly reflects when this cost is paid on .NET 9 might be tricky.

amanasifkhalid: Sorry, I meant the 3-opt writeup.

release-notes/10.0/preview/preview3/runtime.md
amanasifkhalid: Fixed

release-notes/10.0/preview/preview3/runtime.md
AndyAyersMS: I think the 3-opt writeup is ok.

jamesmontemagno:
Please ping me when ready to merge

amanasifkhalid:
@jamesmontemagno I don't have anything else to add

jamesmontemagno:
@AndyAyersMS anything else?

PR #9869, opened 2025-05-01
Update file URLs in releases.json files

richlander:
- download.visualstudio.microsoft.com -> builds.dotnet.microsoft.com
- Updated using https://github.com/richlander/releases-json-url-rewrite
- Double-checked with the script at the same location
- These changes should be the final set of these domain/URL updates
- `release.json` (singular) and `.md` files were not updated.


richlander:
@leecow @lbussell

release-notes/10.0/releases.json
```diff
@@ -1,1733 +1,1733 @@
...
+{
+  "channel-version": "10.0",
+  "latest-release": "10.0.0-preview.3",
+  "latest-release-date": "2025-04-10",
+  "latest-runtime": "10.0.0-preview.3.25171.5",
+  "latest-sdk": "10.0.100-preview.3.25201.16",
+  "support-phase": "preview",
+  "release-type": "lts",
+  "lifecycle-policy": "https://aka.ms/dotnetcoresupport",
+  "releases": [
```
lbussell, line 10: What is the diff here? Line endings? I was expecting only the "url" properties to change.

leecow (approved):

release-notes/10.0/releases.json
richlander: Probably. I ran my tool on Ubuntu.

PR #9889, opened 2025-05-13
Onboard the dotnet/issue-labeler for 'area-' label prediction on issues

jeffhandley:
The [dotnet/issue-labeler](https://github.com/dotnet/issue-labeler/wiki) uses an ML.NET model trained from this repository to automatically predict and apply labels to new issues. This onboards the labeler into the dotnet/core repository, configuring it to predict `area-` labels on new issues. Pull requests will not be labeled.


RussKie (approved):

jeffhandley:
> 23 |         type: number

That is a bogus lint error from the actionlinter. `number` is a valid type for workflow dispatch inputs:
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputsinput_idtype

So far, I'm striking out figuring out how to configure superlinter to ignore that warning and/or ignore the workflows.

The yaml lint errors are also... irritating, but presumably could be addressed.

jeffhandley:
Bumping superlinter to v7 addressed those lint issues. But now there are others to address. Will continue Wednesday.

.github/workflows/labeler-cache-retention.yml
```diff
@@ -0,0 +1,36 @@
...
+#
+# Regularly restore the prediction model from cache to prevent cache eviction
+name: "Labeler: Cache Retention"
+
+# For more information about GitHub's action cache limits and eviction policy, see:
+# https://docs.github.com/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
+
+on:
+  schedule:
+    - cron: "24 12 * * *" # 12:24 every day (arbitrary time daily)
```
tarekgh: Is this in pacific time zone?

.github/workflows/labeler-promote.yml
```diff
@@ -0,0 +1,36 @@
...
+        default: "staged"
+      backup_key:
+        description: "The cache key suffix to use for backing up the currently active model. Defaults to 'backup'."
+        default: "backup"
+
+permissions:
+  actions: write
+
+jobs:
+  promote-issues:
```
tarekgh, line 27: >promote-issues: [](http://example.com/codeflow?start=2&length=15)

We don't need to have similar check here:

github.repository_owner == 'dotnet'?

tarekgh: Oh, looks this specific to the model if I understand it correctly

.github/workflows/labeler-cache-retention.yml
jeffhandley: UTC (I didn't think they had it documented, but it seems they do).
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

.github/workflows/labeler-promote.yml
jeffhandley: Since the workflow is only triggered manually, we don't need to check that. We put the guard in place for workflows that run automatically (so that forks don't get red workflows).

.github/workflows/labeler-cache-retention.yml
tarekgh: Would it help if we add this to the comment?

jeffhandley:
```suggestion
    - cron: "24 12 * * *" # 12:24 UTC every day (arbitrary time daily)
```

PR #9946, opened 2025-06-23
.NET 10 P6 - FSharp

jamesmontemagno:
This PR updates the fsharp preview6 release notes with the latest manual changes as requested.


T-Gro:
Thanks for the ping @jamesmontemagno .
We do not have anything mentionable this time, that is  for the diff between p5 and p6 - I already spread all the good news in the previous post.

PR #9982, opened 2025-07-17
Remove note about Current

richlander:
Seems like this is more confusing at this point than helpful.

@leecow


Frulfump:
Yeah I guess so, well especially since it didn't reference the rationale for the change https://github.com/dotnet/designs/blob/main/accepted/2022/dotnet-release-labels.md#currently-used-labels . I've linked to that document in the past to argue why the name "Current" is confusing in similar contexts i.e https://github.com/Azure/azure-sdk-for-net/issues/51170#issuecomment-3078932861

PR #10016, opened 2025-08-05
.NET Aug2025 8A servicing release Updates

victorisr:


leecow (approved):
releases-index and releases.json files look good

PR #10019, opened 2025-08-06
Remove non-existing OS

am11:
AFAIK, Windows Server 2008 was the last edition to support x86 arch. Also, Windows Server 2025 is the first one to support arm64 (currently insider preview).


am11:
cc @richlander

richlander:
I appreciate the change. The markdown files are generated from the JSON files using tools from https://github.com/richlander/distroessed/. I'm happy to them for you.

`x86` is intended to mean 32-bit, either via native or WoW64. This raises the question of what we'll say for x64 w/rt macOS. Maybe you are right.

> macOS: The x64 emulator (Rosetta 2) is supported on macOS Arm64.

That's probably a good model. We can consider WoW64 to be similar. And remove (when the time comes) x64 from macOS and x86 from Windows.

Fair?

am11:
Sounds good. Another possible direction could be to add a separate column for Processor vs. Process architecture (like how C# distinguishes them).

MARSHinshipUNITY-Byte:
Deploy

richlander:
Sorry for the delay! I appreciate the nudge on this change.

Addressed in https://github.com/dotnet/core/pull/10137.

PR #10063, opened 2025-08-19
API Diff .NET 10.0 Preview 7 -> RC1

ericstj:
- [ ] ASP.NET - @dotnet/aspnet-api-review 
- [x] Winforms -- @merriemcgaw @JeremyKuhne @KlausLoeffelmann 
- [ ] Libraries - @artl93 @jeffschwMSFT @jeffhandley @karelz @SamMonoRT 

cc @tannergooding @bartonjs @jkoritzinsky @StephenMolloy @jkotas @jcouv @sbomer

### DIFF
--- a/release-notes/10.0/preview/rc1/api-diff/Microsoft.AspNetCore.App/10.0-rc1.md
+++ b/release-notes/10.0/preview/rc1/api-diff/Microsoft.AspNetCore.App/10.0-rc1.md
@@ -1,0 +1,15 @@
+# API difference between .NET 10.0 Preview 7 and .NET 10.0 RC 1
+
+API listing follows standard diff formatting.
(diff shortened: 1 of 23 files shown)


bartonjs:
System.Security LGTM

tannergooding:
@ericstj, sorry I didn't get to this sooner.

Is it expected that `System.Numerics.Tensors` isn't showing up here? It's an out of band package, but such packages are likely still interesting to include in the diffs.

ericstj:
These diffs only include shared frameworks, not additional OOB packages.  It might be interesting to actually generate the diffs as part of the package itself cc @ViktorHofer. 

I could see us auto-generating both the API shape and the diff and putting it in the package readme.  We might need to choose a single framework to represent though.

PR #10073, opened 2025-09-08
Add dotnet.config to known RC1 issues

Youssef1313:


release-notes/10.0/known-issues.md
```diff
@@ -26,3 +26,25 @@ Add the following to a `Directory.Build.props` file that is imported in all MAUI
 ```
 
 The behavior will be **fixed in the .NET 10 preview 5** release. The workaround should be removed after that release.
+
+
+## 'dotnet.config' will be removed in favor of 'global.json'
```
Evangelink:
```suggestion
## 'dotnet.config' will be removed in favor of 'global.json'
```

release-notes/10.0/known-issues.md
```diff
@@ -26,3 +26,25 @@ Add the following to a `Directory.Build.props` file that is imported in all MAUI
 ```
 
 The behavior will be **fixed in the .NET 10 preview 5** release. The workaround should be removed after that release.
+
+
+## 'dotnet.config' will be removed in favor of 'global.json'
+
+In .NET 10, we introduced 'dotnet.config' file that is used to set the test runner and is the way to opt-in using `dotnet test` for Microsoft.Testing.Platform. This is already shipping in RC1, but will be removed in RC2.
```
Evangelink: You are using single quotes ' here but back-tilt ` below. I would standardize to `.

I don't know if that's allow in titles as it could cause issues with hyperlinks.

release-notes/10.0/known-issues.md
```diff
@@ -26,3 +26,25 @@ Add the following to a `Directory.Build.props` file that is imported in all MAUI
...
+
+With `global.json`, this becomes:
+
+```json
+{
+  "test": {
+    "runner": "Microsoft.Testing.Platform"
+  }
+}
+```
```
Evangelink, line 49: I would also add an example with a version of .NET being pinned just for extra clarity.

PR #10094, opened 2025-09-15
Update efcore.md for RC 2

jongalloway:
Please update the release notes here as needed for RC 2.\n\n/cc @roji


roji:
@jongalloway pushed a commit with some of the EF rc.2 highlights.

PR #10098, opened 2025-09-15
[RC2] SDK release notes

jongalloway:
Add RC2 scaffold section to SDK release notes file.


jongalloway:
@baronfel Thanks for the progress on the SDK RC2 release notes! If possible, it would be great to have these wrapped up by end of week to support the 10/14 release.

baronfel:
I think this might be _it_ for us - will confirm today hopefully.

PR #10126, opened 2025-10-17
Update supported OSes

Falco20019:
The supported-os files seem to be a bit outdated. Therefore many changes. I hope I listed all of them.

Additions:
* Android 16
* iOS 26
* iPadOS 26
* macOS 26
* tvOS 26
* Debian 13
* openSUSE Leap 16.0
* SUSE Enterprise Linux 15.7
* Ubuntu 25.10
* Windows 11 25H2

Removals:
* Android 12 / 12.1
* iOS 17
* macOS 13
* tvOS 17
* tvOS 18
* Windows 11 22H2 (E)
* Windows 10 22H2

NOTE: I updated the MDs using the latest distroessed, but it seems the formatting is off. Do I have to adjust something or use a config?

/CC @richlander

PR #10137, opened 2025-10-28
Updated Supported OSes

richlander:
Builds on, resolves, and reflects: 

- https://github.com/dotnet/core/pull/10126
- https://github.com/dotnet/core/pull/10019
- https://github.com/dotnet/runtime/issues/121140


richlander:
Markdown table formatting still isn't right. That largely doesn't matter. I have a new implementation that resolves this, but it is isn't integrated. When I integrate it, the tables will be pretty.

leecow (approved):

am11:
@richlander does it make sense to emphasize libc-based portability first in docs, then present distros as examples rather than the definitive list of supported OSes? i.e. https://github.com/dotnet/core/pull/10140

PR #10140, opened 2025-10-29
Reorient Linux support docs to portability-first model

am11:
Restructured Linux section to center on libc-based portability. Distributions are presented as examples meeting that baseline, not as the defining scope of Linux support.

### DIFF
--- a/release-notes/10.0/supported-os.md
+++ b/release-notes/10.0/supported-os.md
@@ -40,6 +40,16 @@ Notes:
 
 ## Linux
 
+Microsoft-provided [Portable Linux builds](../../linux.md) define minimum compatibility primarily via libc version:
+
+| Libc  | Version | Architectures | Source       |
+| ----- | ------- | ------------- | ------------ |
+| glibc | 2.27    | Arm64, x64    | Ubuntu 18.04 |
+| glibc | 2.35    | Arm32         | Ubuntu 22.04 |
+| musl  | 1.2.3   | Arm32, Arm64, x64 | Alpine 3.17 |
+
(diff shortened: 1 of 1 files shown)


release-notes/10.0/supported-os.md
```diff
@@ -49,7 +59,7 @@ Notes:
 | [Fedora][13]        | 42       | Arm32, Arm64, x64 | [Lifecycle][14] |
 | [openSUSE Leap][15] | 16.0, 15.6 | Arm64, x64      | [Lifecycle][16] |
 | [Red Hat Enterprise Linux][17] | 10, 9 | Arm64, ppc64le, s390x, x64 | [Lifecycle][18] |
-| [SUSE Enterprise Linux][19] | 16.0, 15.7, 15.6 | Arm64, x64 | [Lifecycle][20] |
+| [SUSE Linux Enterprise][19] | 16.0, 15.7, 15.6 | Arm64, x64 | [Lifecycle][20] |
```
am11, line 62: Product name: https://www.suse.com/products/server/

richlander:
This is an interesting idea.

One one hand, this makes perfect sense. A leads to B. However, it's also theory leads to concrete application. I'm not sure how many users care about the theory. I rarely get asked about the glibc aspects. It's almost always low-level developers like yourself or other providers like AWS.

I've also considered this section as "left to the reader" content. 

What do you think?

am11:
Got it 🙂 I agree, the detailed libc info is mostly for curious readers, so having it after the distros works. I’ll close this PR.

I've opened a separate PR for SUSE Linux Enterprise name correction https://github.com/dotnet/core/pull/10176.

PR #10123, opened 2025-10-15
Update NET 9.0 EOL Date

Varorbc:

PR #10154, opened 2025-11-11
Update NET 9.0 EOL Date

ToreDemant:
Correct NET 9.0 EOL date that was forgotten in the original PR:
https://github.com/dotnet/core/pull/10123


victorisr:
Hello, EOL date for .NET 9.0 has been extended to November 11, 2026, therefore this PR is not valid. 

Closing this and as always, thank you for your contribution!

ToreDemant:
@victorisr isn't the May 12 date then wrong?

ToreDemant:
@victorisr and are you sure it is November **11**? That would be different from 8.0.x

victorisr:
> @victorisr isn't the May 12 date then wrong?
Actually Yes, You are indeed correct. Apologies for the confusion. Your updates has been made effective

PR #10170, opened 2025-11-20
Editorial pass over the dotnet.config file removal

baronfel:
Tweaking this known issue for wording and clarity based on feedback from users.


richlander (approved):
Thanks. This is much clearer to me.

PR #10223, opened 2026-02-05
Update containers.md for Preview 1

jamesmontemagno:
Please update the release notes here as needed for Preview 1.

/cc @lbussell


lbussell:
No changes for .NET 11 P1.

PR #10226, opened 2026-02-05
Update efcore.md for Preview 1

jamesmontemagno:
Please update the release notes here as needed for Preview 1.

/cc @roji


SamMonoRT:
https://github.com/dotnet/EntityFramework.Docs/pull/5256 should have the details needed

roji:
Yep, I'll take care of this soon.

roji:
@jamesmontemagno done. I've done slightly different this time; we already fully have all preview.1 docs ready to go in a `preview-1` branch in the EF docs repo; [here's the page](https://github.com/dotnet/EntityFramework.Docs/blob/preview-1/entity-framework/core/what-is-new/ef-core-11.0/whatsnew.md) that's ready to go live. Rather than duplicating that information here, I'm just linking from here to there, where there's full documentation with code samples and everything. The build is currently failing because that page isn't live yet, but once we push live on our side (on the 10th) everything should work here.

What do you think, does this make sense?

jamesmontemagno:
@roji usually for the first release it is the same... but like for P2 we want to see just the diff, so having it her would be nicer IMHO unless you plan on doing diffs. But the hope is show diffs here and then point to the what's new for source of truth in latest.

roji:
@jamesmontemagno oh 100%, the idea is to only post new features for each preview release in these release notes (so diffs); but I'm proposing to just link to the section in the EF docs that describes those new changes, rather than copying-and-pasting a version of them into here (as I did in the 10 preview release notes).

In any case, I'll push the EF-side preview.1 docs tomorrow during the European day, so everything should be online by the time you're up and links in this PR should resolve without any errors.

jamesmontemagno:
@roji  sounds good, trust ya, you may need to use the full urls for /ef/core/what-is-new/ef-core-11.0/whatsnew.md#complex-types-and-json-columns-on-entity-types-with-tpttpc-inheritance

PR #10241, opened 2026-02-08
RunApiDiff script and documentation improvements

jeffhandley:
Running the API Diff for 11.0.0-preview.1 revealed some opportunities to make this script more user-friendly. The first hurdle faced was the search index for the staging NuGet feed for preview.1 did not have any results, and this led to needing to find all of the version parameter values through other means. Copilot CLI figured it all out, but it was difficult to achieve by hand, and its results weren't fully deterministic. Once the logic was understood, this led to a series of Copilot-authored improvements to the script and documentation with each step captured as its own commit.

A lot of the documentation and logic was based on running the diff _after_ a release was published, but we began running it _before_ the releases during .NET 10 and we intend to continue that approach going forward.

### DIFF
--- a/release-notes/RunApiDiff.md
+++ b/release-notes/RunApiDiff.md
@@ -7,7 +7,7 @@ The [`RunApiDiff.ps1`](./RunApiDiff.ps1) script automatically generates an API c
 - PowerShell 7.0 or later
 - The Microsoft.DotNet.ApiDiff.Tool. Use `-InstallApiDiff` to have the script install it automatically from the transport feed constructed from the current version's major version (`https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet{MAJOR}-transport/nuget/v3/index.json`), or install it manually:
 
-```
+```shell
 dotnet tool install --global Microsoft.DotNet.ApiDiff.Tool --source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet{MAJOR}-transport/nuget/v3/index.json --prerelease
 ```
 
--- a/release-notes/RunApiDiff.ps1
+++ b/release-notes/RunApiDiff.ps1
@@ -1,4 +1,4 @@
-# This script allows running API-diff to generate the dotnet/core report that compares the APIs introduced between two previews, in the format expected for publishing in the dotnet/core repo.
+﻿# This script allows running API-diff to generate the dotnet/core report that compares the APIs introduced between two previews, in the format expected for publishing in the dotnet/core repo.
 
 # Prerequisites:
 # - PowerShell 7.0 or later
@@ -329,6 +329,7 @@ Function DiscoverVersionFromFeed {
 }
 
 Function Write-Color {
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSAvoidUsingWriteHost', '')]
     Param (
         [ValidateNotNullOrEmpty()]
         [string] $newColor
@@ -544,7 +545,7 @@ Function GetPreviewFolderPath {
         Return [IO.Path]::Combine($prefixFolder, "$dotNetVersion.$previewNumberVersion", $apiDiffFolderName)
     }
 
-    $releaseKindFolderName = GetReleaseKindFolderName $dotNetVersion $releaseKind $previewNumberVersion
+    $releaseKindFolderName = GetReleaseKindFolderName -dotNetVersion $dotNetVersion -releaseKind $releaseKind -previewNumberVersion $previewNumberVersion
     Return [IO.Path]::Combine($prefixFolder, "preview", $releaseKindFolderName, $apiDiffFolderName)
 }
 
@@ -963,7 +964,10 @@ If ([System.String]::IsNullOrWhiteSpace($CoreRepo)) {
     try {
         $CoreRepo = git -C $scriptDir rev-parse --show-toplevel 2>$null
     }
-    catch { }
+    catch {
+        # git command may fail if not in a repo; ignore
+        $null = $null
+    }
 
     If ([System.String]::IsNullOrWhiteSpace($CoreRepo)) {
         Write-Error "Could not determine the git repository root from '$scriptDir'. Please specify -CoreRepo explicitly." -ErrorAction Stop
@@ -997,7 +1001,7 @@ If ([System.String]::IsNullOrWhiteSpace($CurrentMajorMinor) -and [System.String]
         Write-Color cyan "Latest existing api-diff: $latestDesc"
 
         # Probe the feed for the next version after the latest api-diff
-        $next = GetNextVersionFromFeed $latestApiDiff.MajorMinor $latestApiDiff.PrereleaseLabel $DotNetPublicFeedUrl
+        $next = GetNextVersionFromFeed -majorMinor $latestApiDiff.MajorMinor -prereleaseLabel $latestApiDiff.PrereleaseLabel -feedUrl $DotNetPublicFeedUrl
 
         If ($next) {
             $CurrentMajorMinor = $next.MajorMinor
@@ -1105,7 +1109,7 @@ $InstallApiDiffCommand = "dotnet tool install --global Microsoft.DotNet.ApiDiff.
 if ($InstallApiDiff) {
     Write-Color white "Installing ApiDiff tool..."
     Write-Color yellow $InstallApiDiffCommand
-    Invoke-Expression $InstallApiDiffCommand
+    & dotnet tool install --global Microsoft.DotNet.ApiDiff.Tool --source $transportFeedUrl --prerelease
 }
 
 $apiDiffCommand = get-command "apidiff" -ErrorAction SilentlyContinue
@@ -1119,7 +1123,7 @@ $apiDiffExe = $apiDiffCommand.Source
 
 ## Create api-diff folder in core repo folder if it doesn't exist
 
-$previewFolderPath = GetPreviewFolderPath $CoreRepo $CurrentMajorMinor $CurrentReleaseKind $CurrentPreviewRCNumber $IsComparingReleases
+$previewFolderPath = GetPreviewFolderPath -rootFolder $CoreRepo -dotNetVersion $CurrentMajorMinor -releaseKind $CurrentReleaseKind -previewNumberVersion $CurrentPreviewRCNumber -IsComparingReleases $IsComparingReleases
 If (-Not (Test-Path -Path $previewFolderPath))
 {
     Write-Color white "Creating new diff folder: $previewFolderPath"
@@ -1129,11 +1133,11 @@ If (-Not (Test-Path -Path $previewFolderPath))
 ## Run the ApiDiff commands
 
 # Example: "10.0-preview2"
-$currentDotNetFullName = GetDotNetFullName $IsComparingReleases $CurrentMajorMinor $CurrentReleaseKind $CurrentPreviewRCNumber
+$currentDotNetFullName = GetDotNetFullName -IsComparingReleases $IsComparingReleases -dotNetVersion $CurrentMajorMinor -releaseKind $CurrentReleaseKind -previewNumberVersion $CurrentPreviewRCNumber
 
 # Examples: ".NET 10 Preview 1" and ".NET 10 Preview 2"
-$previousDotNetFriendlyName = GetDotNetFriendlyName $PreviousMajorMinor $PreviousReleaseKind $PreviousPreviewRCNumber
-$currentDotNetFriendlyName = GetDotNetFriendlyName $CurrentMajorMinor $CurrentReleaseKind $CurrentPreviewRCNumber
+$previousDotNetFriendlyName = GetDotNetFriendlyName -DotNetVersion $PreviousMajorMinor -releaseKind $PreviousReleaseKind -PreviewNumberVersion $PreviousPreviewRCNumber
+$currentDotNetFriendlyName = GetDotNetFriendlyName -DotNetVersion $CurrentMajorMinor -releaseKind $CurrentReleaseKind -PreviewNumberVersion $CurrentPreviewRCNumber
 
 $sdksToProcess = @()
 If (-Not $ExcludeNetCore) { $sdksToProcess += "NETCore" }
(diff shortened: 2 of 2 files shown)


release-notes/RunApiDiff.md
```diff
@@ -1,27 +1,96 @@
...
+The [`RunApiDiff.ps1`](./RunApiDiff.ps1) script automatically generates an API comparison report between two .NET versions, in the format expected for publishing in the dotnet/core repo.
 
-## Instructions
+## Prerequisites
 
-1. Clone the dotnet/core repo. Let's assume you clone it into `D:\core`.
-2. Create a temporary directory. Let's assume you create it in `D:\tmp`.
-3. Install or update your copy of the APIDiff tool (use the latest transport feed):
+- PowerShell 7.0 or later
+- Azure CLI (`az login`) if using authenticated Azure DevOps feeds
```
ericstj: This should not be required.  We don't want to generate from internal feeds - only nightly feeds which should be public.

release-notes/RunApiDiff.md
```diff
@@ -1,27 +1,96 @@
...
+```
+dotnet tool install --global Microsoft.DotNet.ApiDiff.Tool --prerelease
+```
+
+## Quick Start
+
+The simplest usage only requires the NuGet feed URL for the "current" (after) version, which will compare the staged release against the most recent publicly released version.
+
+```powershell
+.\RunApiDiff.ps1 <current-nuget-feed-url>
```
ericstj: Feed URL is not a great UX.  These are not intuitive.

release-notes/RunApiDiff.md
```diff
@@ -1,27 +1,96 @@
...
+| Parameter | Description |
+|---|---|
+| `ExcludeNetCore` | Skip the Microsoft.NETCore.App comparison |
+| `ExcludeAspNetCore` | Skip the Microsoft.AspNetCore.App comparison |
+| `ExcludeWindowsDesktop` | Skip the Microsoft.WindowsDesktop.App comparison |
+| `InstallApiDiff` | Install or update the ApiDiff tool before running |
+
+## Examples
+
+Simplest — auto-discover everything from the feeds (compares latest on `PreviousNuGetFeed` against latest on `CurrentNuGetFeed`):
```
ericstj: duplicative

release-notes/RunApiDiff.md
```diff
@@ -1,27 +1,96 @@
...
+
+```powershell
+.\RunApiDiff.ps1 <current-nuget-feed>
+```
+
+Explicit version parameters (comparing .NET 10.0 Preview 7 to RC 1):
```
ericstj: If I'm wishing, I think the best UX is just to specify a current version.  If it's pre-release then we find the previous pre-release.  If it's not, then we find the previous released version.  Only specify params if we can't infer.  That said, It's fine to specify all these.

.github/skills/api-diff/SKILL.md
```diff
@@ -0,0 +1,35 @@
+---
+name: api-diff
```
ericstj, line 2: Do we really need all these markdown files to make the script callable by CCA?  I like the idea of enabling this to run in relnotes PR, but I would hope the we could avoid duplication just by pointing to the docs?  I guess it's not too big of a deal if the same thing refactoring scripts will keep these docs up to date.

release-notes/RunApiDiff.md
```diff
@@ -1,27 +1,87 @@
...
 
 ```powershell
+.\RunApiDiff.ps1
+```
+
+## Parameters
+
+### Version Parameters
+
+All version parameters can be auto-discovered. When no version information is provided, the script scans existing `api-diff` folders in the repository to find the latest version and infers the next one in the progression (preview.1 → preview.2 → ... → preview.7 → rc.1 → rc.2 → GA → next major preview.1). When `PreviousVersion` or `CurrentVersion` is provided, the `MajorMinor` and `PrereleaseLabel` values are extracted from it automatically.
```
ericstj:
```suggestion
By default the script assumes a diff will be produced for the next preview. When no version information is provided, the script scans existing `api-diff` folders in the repository to find the latest version and infers the next one in the progression (preview.1 → preview.2 → ... → preview.7 → rc.1 → rc.2 → GA → next major preview.1). When `PreviousVersion` or `CurrentVersion` is provided, the `MajorMinor` and `PrereleaseLabel` values are extracted from it automatically.
```

release-notes/RunApiDiff.md
```diff
@@ -1,27 +1,87 @@
...
+| `PreviousPrereleaseLabel` | Prerelease label for the "before" version (e.g., `preview.7`, `rc.1`). Omit for GA. | Inferred from api-diffs, extracted from `PreviousVersion`, or discovered from `PreviousNuGetFeed` |
+| `CurrentMajorMinor` | The "after" .NET major.minor version (e.g., `10.0`) | Inferred from api-diffs, extracted from `CurrentVersion`, or discovered from `CurrentNuGetFeed` |
+| `CurrentPrereleaseLabel` | Prerelease label for the "after" version (e.g., `preview.7`, `rc.1`). Omit for GA. | Inferred from api-diffs, extracted from `CurrentVersion`, or discovered from `CurrentNuGetFeed` |
+
+### Feed Parameters
+
+| Parameter | Description | Default |
+|---|---|---|
+| `CurrentNuGetFeed` | NuGet feed URL for downloading "after" packages | Constructed from `CurrentMajorMinor`: `https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet{MAJOR}/nuget/v3/index.json` |
+| `PreviousNuGetFeed` | NuGet feed URL for downloading "before" packages | `https://api.nuget.org/v3/index.json` |
```
ericstj: Nit: consider using dotnet-public feed for this to avoid the untrusted endpoint.

release-notes/RunApiDiff.ps1
```diff
@@ -89,59 +76,214 @@ Param (
...
+
+    $serviceIndex = Invoke-RestMethod -Uri $feedUrl
+    $flatContainer = $serviceIndex.resources | Where-Object { $_.'@type' -match 'PackageBaseAddress' } | Select-Object -First 1
+
+    If (-not $flatContainer) {
+        Write-Error "Could not find PackageBaseAddress endpoint in feed '$feedUrl'. Please specify -${label}MajorMinor and -${label}PrereleaseLabel explicitly." -ErrorAction Stop
+    }
+
+    $baseUrl = $flatContainer.'@id'
+    $versionsUrl = "${baseUrl}microsoft.netcore.app.ref/index.json"
```
ericstj: From .NET 10.0 onward it's the case that we'll have the same version for NETCore.App, WindowsDesktop, ASP.NET -- however that hasn't always been the case.

release-notes/RunApiDiff.ps1
```diff
@@ -89,59 +76,214 @@ Param (
...
+    $versionsResult = Invoke-RestMethod -Uri $versionsUrl
+
+    If (-not $versionsResult.versions -or $versionsResult.versions.Count -eq 0) {
+        Write-Error "No versions of Microsoft.NETCore.App.Ref found on feed '$feedUrl'. Please specify -${label}MajorMinor and -${label}PrereleaseLabel explicitly." -ErrorAction Stop
+    }
+
+    $latestVersion = $versionsResult.versions | Select-Object -Last 1
+    Write-Color cyan "Latest version on feed: $latestVersion"
```
ericstj: Sometimes branding won't happen in time and `main` might be producing newer versions of the previous release.  I can't think of a better way to handle this, but just calling that out as a challenge with auto-discovery of latest version.  

We have "releases.json" stuff, but that's not ready soon enough.  It's possible that "darc" might have a data feed here.  All that said, this is probably good enough for our cases.

release-notes/RunApiDiff.ps1
```diff
@@ -89,59 +76,214 @@ Param (
...
+        }
+    }
+
+    If ($entries.Count -eq 0) { Return $null }
+
+    Return ($entries | Sort-Object { $_.SortKey } | Select-Object -Last 1)
+}
+
+## Given a version milestone, return the next one in the progression
+## preview.1 → preview.2 → ... → preview.7 → rc.1 → rc.2 → ga → {next major}.0 preview.1
```
ericstj: Until it's not 😆  I wonder if we can get the next version by just probing the feed?

.github/skills/api-diff/SKILL.md
jeffhandley: You're right; I got carried away with the SKILL here. I'm going to substantially minimize the skill and instead of it having interactivity for running locally, focus on having it map the user's intent to the appropriate arguments to the script. That way it has the focus of invoking from an agent rather than invoking locally.

release-notes/RunApiDiff.ps1
jeffhandley: Thanks. Updated to query each package independently, with the CLI argument version number being understood at the Microsoft.NETCore.App.Ref version.

release-notes/RunApiDiff.ps1
jeffhandley: Implemented. I was tempted to just leave this as-is and anticipate having to change this code whenever that sequence changes, but you're right that this will surprise us at some point in the future. Discovering from the feed is a good idea. Thanks.

release-notes/RunApiDiff.ps1
```diff
@@ -764,22 +958,166 @@ if ($PSVersionTable.PSVersion.Major -lt 7) {
...
+## Infer current and previous versions from existing api-diffs if not provided
+If ([System.String]::IsNullOrWhiteSpace($CurrentMajorMinor) -and [System.String]::IsNullOrWhiteSpace($CurrentPrereleaseLabel) -and [System.String]::IsNullOrWhiteSpace($CurrentNuGetFeed)) {
+    $latestApiDiff = FindLatestApiDiff $CoreRepo
+    If ($latestApiDiff) {
+        $latestDesc = If ($latestApiDiff.PrereleaseLabel) { "$($latestApiDiff.MajorMinor)-$($latestApiDiff.PrereleaseLabel)" } Else { "$($latestApiDiff.MajorMinor) GA" }
+        Write-Color cyan "Latest existing api-diff: $latestDesc"
+
+        # Probe the feed for the next version after the latest api-diff
+        $latestMajorVersion = $latestApiDiff.MajorMinor.Split(".")[0]
+        $probeFeedUrl = "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet${latestMajorVersion}/nuget/v3/index.json"
```
jeffhandley: @copilot This feed URL template is hard-coded in multiple places. There should be a single place where it's defined, presumably with a function that accepts the majorVersion to plug in.

release-notes/RunApiDiff.ps1
```diff
@@ -764,22 +958,166 @@ if ($PSVersionTable.PSVersion.Major -lt 7) {
...
+    Write-Error "CurrentMajorMinor is required. Specify it explicitly or provide -CurrentNuGetFeed to auto-discover." -ErrorAction Stop
+}
+
+# Validate that previous and current versions are different
+If ($PreviousMajorMinor -eq $CurrentMajorMinor -and $PreviousPrereleaseLabel -eq $CurrentPrereleaseLabel) {
+    $previousDesc = If ($PreviousPrereleaseLabel) { "$PreviousMajorMinor-$PreviousPrereleaseLabel" } Else { "$PreviousMajorMinor GA" }
+    Write-Error "Previous and current versions are the same ($previousDesc). Ensure -PreviousNuGetFeed and -CurrentNuGetFeed point to different versions, or specify version parameters explicitly." -ErrorAction Stop
+}
 
 # True when comparing 8.0 GA with 9.0 GA
```
jeffhandley:
```suggestion
# True when comparing GA releases
```

release-notes/RunApiDiff.md
```diff
@@ -1,27 +1,87 @@
...
+| `PreviousPrereleaseLabel` | Prerelease label for the "before" version (e.g., `preview.7`, `rc.1`). Omit for GA. | Inferred from api-diffs, extracted from `PreviousVersion`, or discovered from `PreviousNuGetFeed` |
+| `CurrentMajorMinor` | The "after" .NET major.minor version (e.g., `10.0`) | Inferred from api-diffs, extracted from `CurrentVersion`, or discovered from `CurrentNuGetFeed` |
+| `CurrentPrereleaseLabel` | Prerelease label for the "after" version (e.g., `preview.7`, `rc.1`). Omit for GA. | Inferred from api-diffs, extracted from `CurrentVersion`, or discovered from `CurrentNuGetFeed` |
+
+### Feed Parameters
+
+| Parameter | Description | Default |
+|---|---|---|
+| `CurrentNuGetFeed` | NuGet feed URL for downloading "after" packages | Constructed from `CurrentMajorMinor`: `https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet{MAJOR}/nuget/v3/index.json` |
+| `PreviousNuGetFeed` | NuGet feed URL for downloading "before" packages | Constructed from `PreviousMajorMinor`: `https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet{MAJOR}/nuget/v3/index.json` |
```
ericstj: As I mentioned before, we may want to use the dotnet-public feed.  I don't think we will always publish the final released build to the nightly feeds.  I don't think we'll publish if it was an internal only build, or if it was a stable build.  Those only go to nuget.org, then get mirrored back to dotnet-public.

Another option here would be to try to use the major version feed, then try public, then fail.

ericstj (approved):
Minor suggestion on feeds.  

I like all the reduction of parameters by inferring and searching.  It makes this much easier to use.

I wish the nuget searching/downloading didn't require so much code here.  We could use https://learn.microsoft.com/en-us/nuget/reference/powershell-reference but I don't want to feature-creep this PR.

release-notes/RunApiDiff.md
jeffhandley: Oh, I understand now. I thought you just meant using the "dotnet" "public" feed (`dnceng/public`), but you meant the actual `dotnet-public` feed at `https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json`. This explanation helped me understand the spirit of using the packages that make it out to nuget.org and then back in to a filtered feed.

PR #10263, opened 2026-02-13
Fix Aspire link in the roadmap

danroth27:


danroth27:
@maddymontaquila Should Aspire still be included in the .NET Roadmap? Or should it be removed at this point?

PR #10267, opened 2026-02-17
Add libraries.md for .NET 11 Preview 2

jamesmontemagno:
Tagging for review:

- [x] @adamsitnik — Process APIs section (assignee/merger on both #124264 and #124256)
- [x] @stephentoub — System.Text.Json section (assignee/coauthor/merger on #123940)
- [x] @iremyux — Tar archive section (assignee/merger on #123407)
- [x] @kasperk81 — Tar archive section (community author of #123407)

This PR deletes the libraries-release-notes skill, in favor of #10282 adding a general-purpose release-notes skill. That skill was loaded locally and used to generate the updated release notes here.

### DIFF
--- a/release-notes/11.0/preview/preview2/fsharp.md
+++ b/release-notes/11.0/preview/preview2/fsharp.md
@@ -1,0 +1,110 @@
+# F# in .NET 11 Preview 2 - Release Notes
+
+Here's a summary of what's new in F# in this Preview 2 release:
+
+- [New language feature: Simplified DIM interface hierarchies (preview)](#simplified-dim-interface-hierarchies)
+- [Compiler performance: Overload resolution caching (preview)](#overload-resolution-caching)
+- [FSharp.Core: Big-O complexity documentation for all collection functions](#big-o-complexity-documentation)
+- [Query expression fixes for EF Core and LINQ providers](#query-expression-fixes)
+- [IDE: Find All References and Rename overhaul](#find-all-references-and-rename-overhaul)
+- [Bug fixes](#bug-fixes-and-other-improvements)
+
+## Simplified DIM interface hierarchies
+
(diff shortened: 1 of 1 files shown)


kasperk81:
Given the volume of PRs in runtime repo, manually calling out API changes will generate a lot of noise and redundant mentions. dotnet/docs already has automation to track new API surface, so this likely isn’t needed.

jeffhandley:
Thanks, @kasperk81; that makes sense. I've tuned the skill instructions such that it won't call out all API changes, but it will highlight high-value feature additions delivered through new APIs--like your `TarFile` contribution. I appreciate you reviewing this too.

release-notes/11.0/preview/preview2/libraries.md
```diff
@@ -0,0 +1,71 @@
...
+
+    public int ExitCode { get; }
+    public bool Canceled { get; }
+    public PosixSignal? Signal { get; }
+}
+```
+
+### PosixSignal.SIGKILL
+
+[dotnet/runtime#124256](https://github.com/dotnet/runtime/pull/124256) adds `PosixSignal.SIGKILL` to the `PosixSignal` enum, enabling the new process APIs to send and represent SIGKILL signals. SIGKILL cannot be caught or ignored per POSIX semantics — the OS enforces this naturally, throwing appropriate exceptions if registration is attempted. This supports `SafeChildProcessHandle.SendSignal` and `Kill` methods in the upcoming `ChildProcess` API ([dotnet/runtime#123380](https://github.com/dotnet/runtime/issues/123380)).
```
stephentoub: This feature is useless as of Preview 2 because you can't handle SIGKILL and the API to send a PosixSignal hasn't yet been added. I don't think it's worth calling out.

PR #10372, opened 2026-04-15
docs: update .NET 11 preview README release info

WeihanLi:
update .NET 11 preview README release info
(maybe we could improve the release skils to cover this but did not find related info to update, so leave it to team)


richlander (approved):

richlander:
Thanks for the fix.

PR #10373, opened 2026-04-15
[docs][release-notes] Fix broken link for dotnet 11

usta:
It was repeating top-directory path within link which was causing 404.


richlander (approved):

richlander:
Thanks for the fix!

PR #10388, opened 2026-04-30
[release-notes] .NET 11 Preview 4

danroth27:
Release notes for **.NET 11 Preview 4**.

This is the **base PR** for the milestone — it carries the shared metadata
(`README.md`, `changes.json`, `build-metadata.json`, `features.json`). Each
component file ships in its own PR targeting this branch so component teams
review their own file in isolation. See [`release-notes` skill / `pr-layout.md`](https://github.com/dotnet/core/blob/main/.github/skills/release-notes/references/pr-layout.md).

The component PRs will be linked back here as they open.

Generated end-to-end by the `release-notes` skill (dry-run promoted to real
PRs after local validation). AI-authored content; component teams should vet
for accuracy and resolve any inline `<!-- TODO -->` placeholders before
marking their PR ready for review.

PR #10389, opened 2026-04-30
[release-notes] ASP.NET Core in .NET 11 Preview 4

danroth27:
ASP.NET Core release notes for .NET 11 Preview 4.

Targets the milestone base branch [#10388](https://github.com/dotnet/core/pull/10388). Merging this PR
into `release-notes/11.0-preview4` adds `aspnetcore.md` to the milestone.

Generated end-to-end by the `release-notes` skill. AI-authored content;
please review for accuracy, resolve any inline `<!-- TODO -->` placeholders,
and then mark this PR ready for review.

cc @danroth27


release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,185 @@
...
+- [Bug fixes](#bug-fixes)
+- [Community contributors](#community-contributors)
+
+ASP.NET Core updates in .NET 11:
+
+- [What's new in ASP.NET Core in .NET 11](https://learn.microsoft.com/aspnet/core/release-notes/aspnetcore-11)
+
+## Support for the HTTP QUERY method
+
+ASP.NET Core now recognizes [HTTP QUERY](https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/) as a first-class HTTP method in routing, minimal APIs, and OpenAPI document generation ([dotnet/aspnetcore #65714](https://github.com/dotnet/aspnetcore/pull/65714)). QUERY is a proposed safe, idempotent method that lets clients send a request body when describing a search — useful when a query is too large or too structured to fit in a URL.
```
BrennanConroy: This just added HTTP QUERY to OpenAPI documents.

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,185 @@
+# ASP.NET Core in .NET 11 Preview 4 - Release Notes
+
+.NET 11 Preview 4 includes new ASP.NET Core features and improvements:
```
BrennanConroy, line 3: What did you generate this from in case you missed more changes? It's missing runtime-async being enabled for shared framework dlls.
https://github.com/dotnet/aspnetcore/pull/66449

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,185 @@
...
+}
+```
+
+The method returns `true` when the client was successfully asked to begin pausing. Clients can defer the request via the optional `onPauseRequested` callback in `CircuitStartOptions`.
+
+<!-- TODO: verify exact shape of CircuitRegistry / how a hosting app obtains a Circuit instance; PR shows the new API on Circuit but not the discovery path. -->
+
+## Virtualize keeps the viewport stable when content above it changes
+
+Preview 3 taught Blazor's `Virtualize<TItem>` to handle items with different heights. Preview 4 builds on that with viewport stability: visible items stay in place when items above the viewport change height, and when items are prepended to the collection ([dotnet/aspnetcore #65951](https://github.com/dotnet/aspnetcore/pull/65951)).
```
ilonatommy: A bit different. https://github.com/dotnet/aspnetcore/pull/65951 is about collapsing/expansion of items above the viewport and uses native browser anchoring or manual scroll for Safari/tables. Prepend moves the viewport.
(Why bot modified the milestone to p3 for this one? Based on the merge date it should have been p4. Now I'm confused if it should land in this article at all)

Prepend is fixed only by `AnchorMode` from https://github.com/dotnet/aspnetcore/pull/66521 (also in p4) and relies on detecting the change of the number of items in DOM and checking if the 1st of them changed (prepend) or stayed the same (append). For reference objects such a check requires passing additional argument to the component: `ItemComparer` which otherwise uses `EqualityComparer<T>.Default`.
Users can choose the desired behavior for the edges of lists (auto-follow / do not move the viewport) manipulating with `AnchorMode = Beginning|None|End` values.

dariatiurina (approved):

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: I generated a section for the runtime async enablement. It may have been missed because it didn't involve any API changes. I'll check to see if anything else was missed.

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: Thanks @ilonatommy! Take a look at the updated virtualization sections and see if they're now accurate.

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,218 @@
+# ASP.NET Core in .NET 11 Preview 4 - Release Notes
+
+.NET 11 Preview 4 includes new ASP.NET Core features and improvements:
```
danroth27, line 3: @javiercn Should we include the new Blazor Gateway in this list? Does it have any new user facing functionality that we want to call out?

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -137,6 +166,12 @@ Two trimming changes shrink published Blazor WebAssembly apps that don't use Ope
 
 Apps that use OpenTelemetry or hot reload aren't affected — the feature switches default on in those configurations.
 
+## Runtime-async enabled for shared framework libraries
+
+ASP.NET Core's shared-framework-only libraries are now compiled with the `runtime-async` feature on `net11.0+` ([dotnet/aspnetcore #66449](https://github.com/dotnet/aspnetcore/pull/66449), backporting [#66200](https://github.com/dotnet/aspnetcore/pull/66200)). Runtime-async lets the runtime, rather than the C# compiler, generate the state machine for `async`/`await`, which can reduce per-await allocations and improve diagnostics. This is an internal codegen change with no public API impact — apps targeting `net11.0` automatically benefit when they call into the affected ASP.NET Core libraries.
```
BrennanConroy: nit: don't need to show backport PR

BrennanConroy: I think we might want to explicitly call out that we're wanting people to try out their apps and help find any bugs that might be caused by runtime-async being enabled.

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: After investigating: my recommendation is to **not** call out Blazor Gateway in Preview 4 release notes.

**What shipped (#65982):** the `Microsoft.AspNetCore.Components.Gateway` package and its `BlazorGateway.BuildWebHost` entry point, used by the aspnetcore E2E test infrastructure as a replacement for DevServer.

**Why skip it for P4:**
- No user-observable behavior change. `dotnet new blazorwasm` + `dotnet run` still uses `Microsoft.AspNetCore.Components.WebAssembly.DevServer` (verified against `11.0.0-preview.4.26230.115`).
- The `BlazorGateway` class XML doc says *"Intended for framework test use only"*, and the Gateway package isn''t published to NuGet.org.
- No public opt-in (no `_UseBlazorGateway` MSBuild property, no template change).

**What''s actually interesting in there** (and worth a real callout once it''s user-facing): the design has `ClientApps` config sections per app, each with `PathPrefix`, `EndpointsManifest`, and a `ConfigEndpointPath`/`ConfigResponse` pair that flows JSON config to the client at a known URL. It also wires up YARP via `ReverseProxy` config + `AddServiceDiscoveryDestinationResolver`. That''s clearly the Aspire-orchestrated multi-WASM-app dev story, and it deserves a real callout when the user-facing swap and the Aspire integration land — likely P5/P6.

@javiercn please correct me if any of this is wrong or if there''s a P4-shipped surface I missed.

danroth27: Reversing course on this — we decided to call out the Blazor Gateway in Preview 4 after all. Even though the standalone Blazor WebAssembly template still references `Microsoft.AspNetCore.Components.WebAssembly.DevServer` in P4, users can opt in today by swapping the package reference, and the Gateway adds two genuinely new capabilities (YARP proxying to backends and a runtime config endpoint) that don't exist in the Dev Server.

I built an end-to-end sample (Backend + Gateway + standalone Blazor WASM Client) at [danroth27/AspNetCore11Samples/GatewayDemo](https://github.com/danroth27/AspNetCore11Samples/tree/dotnet11p4/GatewayDemo) and verified the proxying and config-endpoint flows work, then wrote up a section based on that experience. New section added in 8c2acbf7.

danroth27:
Update: pulling the Blazor Gateway section back out of the Preview 4 release notes for now. The package isn't being produced on the build feeds — root cause turned out to be that `Microsoft.NET.Sdk.Web` defaults `IsPackable` to `false`, so the Gateway project's Pack target silently no-ops despite `IsShippingPackage=true` and a custom `NuspecFile`. One-line fix is up at dotnet/aspnetcore#66579, but it won't make Preview 4. We'll bring the content back once the package actually ships.

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,238 @@
...
+```csharp
+var app = WebApplication.Create();
+
+app.MapMethods("/search", ["QUERY"], (SearchRequest request) =>
+    SearchService.Run(request));
+
+app.Run();
+```
+
+Thank you [@kilifu](https://github.com/kilifu) for this contribution!
```
BrennanConroy: rm
Repeated below

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,238 @@
...
+
+Moving the template into ASP.NET Core makes it discoverable from `dotnet new list` without a separate install step, and aligns its servicing with the rest of the web stack.
+
+## TLS handshake observability in Kestrel
+
+Two related changes make it easier to diagnose and customize TLS connections in Kestrel.
+
+`ITlsHandshakeFeature` now exposes an `Exception` property containing the exception thrown during a failed TLS handshake, so middleware and logging can record why a connection failed instead of seeing a bare `IOException` further up the stack ([dotnet/aspnetcore #65807](https://github.com/dotnet/aspnetcore/pull/65807)). The feature continues to work after the handshake fails — Kestrel snapshots the relevant fields off the underlying `SslStream` before it is disposed.
+
+The `TlsClientHelloBytesCallback` option on `HttpsConnectionAdapterOptions` was reworked as a connection middleware ([dotnet/aspnetcore #65808](https://github.com/dotnet/aspnetcore/pull/65808)). The previous callback shape is now obsolete; configure ClientHello inspection via the connection-builder pipeline instead.
```
BrennanConroy: Add snippet using new `UseTlsClientHelloListener` method?

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,238 @@
...
+- **Endpoint filters can observe parameter-binding failures.** When a minimal API endpoint has any filters or filter factories configured, the filter pipeline now runs even if parameter binding fails. Filters can read `HttpContext.Response.StatusCode == 400` and substitute their own response body. Endpoints without filters continue to short-circuit with a 400 as before ([dotnet/aspnetcore #64539](https://github.com/dotnet/aspnetcore/pull/64539)). In Development, set `RouteHandlerOptions.ThrowOnBadRequest = false` so the framework returns a 400 the filter can observe instead of throwing `BadHttpRequestException` to the developer exception page. Thank you [@marcominerva](https://github.com/marcominerva) for this contribution!
```
BrennanConroy: nit: this isn't really openapi related

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,238 @@
...
+## Smaller Blazor WebAssembly publish output
+
+Two trimming changes shrink published Blazor WebAssembly apps that don't use OpenTelemetry or hot reload:
+
+- The `ComponentsMetrics` and `ComponentsActivitySource` types are now gated behind a `[FeatureSwitchDefinition]`, so the trimmer can drop the metrics and tracing call paths from `Renderer` and friends when `System.Diagnostics.Metrics.Meter.IsSupported` is `false` (the default for trimmed apps) ([dotnet/aspnetcore #65901](https://github.com/dotnet/aspnetcore/pull/65901)).
+- `HotReloadManager` now exposes a feature-switched `IsSupported` property tied to `System.Reflection.Metadata.MetadataUpdater.IsSupported`, so the trimmer can eliminate hot-reload caches and metadata-update handler registrations across the renderer when published ([dotnet/aspnetcore #65903](https://github.com/dotnet/aspnetcore/pull/65903)).
+
+Apps that use OpenTelemetry or hot reload aren't affected — the feature switches default on in those configurations.
+
+## Runtime-async enabled for shared framework libraries
```
BrennanConroy, line 278: cc @agocke in case you want to provide input on the messaging here.

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Breaking changes
+
+- **`%2F` is preserved in HTTP/1.1 absolute-form request targets.** Previously, a request like `GET http://host/a%2Fb` was decoded to a path of `/a/b`, while `GET /a%2Fb` (origin-form) preserved the encoded slash. Both forms now resolve to `/a%2Fb`. Apps that depended on the inconsistent absolute-form behavior should update their routing or middleware to expect the encoded segment ([dotnet/aspnetcore #65930](https://github.com/dotnet/aspnetcore/pull/65930)).
+- **`HttpsConnectionAdapterOptions.TlsClientHelloBytesCallback` is obsolete.** Use the new connection middleware shape introduced in [dotnet/aspnetcore #65808](https://github.com/dotnet/aspnetcore/pull/65808). The property continues to work in 11.0 but produces an obsoletion warning.
+
+<!-- Filtered features (significant engineering work, but too niche or off-product for ASP.NET Core release notes):
```
BrennanConroy: Is this going to be removed before merging?

BrennanConroy (approved):

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: Removed the duplicate in 7722c18d.

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: Added a `UseTlsClientHelloListener` snippet in 7722c18d.

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: Good call — split into two sections (`File result types appear in OpenAPI documents` and `Endpoint filters observe parameter-binding failures`) in 7722c18d.

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: Yep — removed the filtered-features HTML comment block in 7722c18d. (Agent audit output that shouldn't ship; will follow up with the agent author to strip it earlier.)

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,246 @@
...
+    options.ListenAnyIP(5001, listenOptions =>
+    {
+        listenOptions.UseHttps();
+        listenOptions.UseTlsClientHelloListener((connection, clientHelloBytes) =>
+        {
+            // Inspect the ClientHello (SNI, ALPN, etc.) for diagnostics or routing.
+        });
+    });
+});
+```
```
danroth27, line 245: @BrennanConroy This is what copilot came up with for a snippet for UseTlsClientHelloListener. Does it look ok?

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -160,20 +159,41 @@ Two related changes make it easier to diagnose and customize TLS connections in
...
+
+```csharp
+var builder = WebApplication.CreateBuilder(args);
+
+builder.WebHost.ConfigureKestrel(options =>
+{
+    options.ListenAnyIP(5001, listenOptions =>
+    {
+        listenOptions.UseHttps();
+        listenOptions.UseTlsClientHelloListener((connection, clientHelloBytes) =>
```
BrennanConroy, line 238: https://github.com/dotnet/aspnetcore/blob/05434a9e554a75c887f204d08849fba337488dc3/src/Servers/Kestrel/Core/src/ListenOptionsHttpsExtensions.cs#L287
> This must be called before <c>UseHttps()</c>

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,246 @@
...
+        {
+            // Inspect the ClientHello (SNI, ALPN, etc.) for diagnostics or routing.
+        });
+    });
```
halter73:
```suggestion
        // This must be called before UseHttps()
        listenOptions.UseTlsClientHelloListener((connection, clientHelloBytes) =>
        {
            // Inspect the ClientHello (SNI, ALPN, etc.) for diagnostics or routing.
        })

        listenOptions.UseHttps();
    });
```

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,246 @@
...
+
+Thank you [@marcominerva](https://github.com/marcominerva) for this contribution!
+
+## Endpoint filters observe parameter-binding failures
+
+When a minimal API endpoint has any filters or filter factories configured, the filter pipeline now runs even if parameter binding fails ([dotnet/aspnetcore #64539](https://github.com/dotnet/aspnetcore/pull/64539)). Filters can read `HttpContext.Response.StatusCode == 400` and substitute their own response body. Endpoints without filters continue to short-circuit with a 400 as before. In Development, set `RouteHandlerOptions.ThrowOnBadRequest = false` so the framework returns a 400 the filter can observe instead of throwing `BadHttpRequestException` to the developer exception page.
```
halter73: Should we call out this was backported to [10.0.8](https://github.com/dotnet/aspnetcore/pull/66221)?


```suggestion
When a minimal API endpoint has any filters or filter factories configured, the filter pipeline now runs even if parameter binding fails ([dotnet/aspnetcore #64539](https://github.com/dotnet/aspnetcore/pull/64539)). Filters can read `HttpContext.Response.StatusCode == 400` and substitute their own response body.

In the `Development` environment, set `RouteHandlerOptions.ThrowOnBadRequest = false` so the framework returns a 400 the filter can observe instead of throwing `BadHttpRequestException` to the developer exception page. This is already the default in non-`Development` environments.

This change was has also been backported to 10.0.8.
```

I also don't think the short-circuit bit adds anything, because what's being short circuited? The non-existent filters? The endpoint gets skipped with or without filters.

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,246 @@
...
+- [Bug fixes](#bug-fixes)
+- [Community contributors](#community-contributors)
+
+ASP.NET Core updates in .NET 11:
+
+- [What's new in ASP.NET Core in .NET 11](https://learn.microsoft.com/aspnet/core/release-notes/aspnetcore-11)
+
+## HTTP QUERY in generated OpenAPI documents
+
+OpenAPI document generation now recognizes [HTTP QUERY](https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/) as a first-class operation type ([dotnet/aspnetcore #65714](https://github.com/dotnet/aspnetcore/pull/65714)). QUERY is a proposed safe, idempotent method that lets clients send a request body when describing a search — useful when a query is too large or too structured to fit in a URL. Routing already accepted arbitrary verb strings via `MapMethods`; this change makes QUERY endpoints show up correctly in the generated OpenAPI document so client generators and API explorers can consume them.
```
halter73:
```suggestion
OpenAPI document generation now recognizes [HTTP QUERY](https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/) as a known operation type ([dotnet/aspnetcore #65714](https://github.com/dotnet/aspnetcore/pull/65714)). QUERY is a proposed safe, idempotent method that lets clients send a request body when describing a search — useful when a query is too large or too structured to fit in a URL. Routing already accepted arbitrary verb strings via `MapMethods`; this change makes QUERY endpoints show up correctly in the generated OpenAPI document so client generators and API explorers can consume them.
```

Because you know someone is going to call us out for calling it "first-class" and not adding `MapQuery()` or `[HttpQuery]`.

halter73 (approved):

release-notes/11.0/preview/preview4/aspnetcore.md
DeagleGross: We can include an example for retrieving exception as well.

```suggestion
        listenOptions.Use(next => async context =>
        {
            await next(context);
        
            var tlsHandshakeFeature = context.Features.Get<ITlsHandshakeFeature>();
            if (tlsHandshakeFeature?.Exception is { } ex)
            {
                Console.WriteLine($"[TLS Handshake Failed] ConnectionId={context.ConnectionId}, Exception={ex.GetType().Name}: {ex.Message}");
            }
        });
        
        listenOptions.UseTlsClientHelloListener((connection, clientHelloBytes) =>
        {
            Console.WriteLine($"TLS Client Hello received on {connection.ConnectionId}, {clientHelloBytes.Length} bytes");
        });
        listenOptions.UseHttps();
    });
```

release-notes/11.0/preview/preview4/aspnetcore.md
mikekistler: The following suggestion includes the change requested by @halter73:

```suggestion
OpenAPI document generation now recognizes [HTTP QUERY](https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/) as a known operation type ([dotnet/aspnetcore #65714](https://github.com/dotnet/aspnetcore/pull/65714)). QUERY is a proposed safe, idempotent method that lets clients send a request body when describing a search — useful when a query is too large or too structured to fit in a URL. Routing already accepted arbitrary verb strings via `MapMethods`, and OpenAPI 3.2 adds a [`query` field to the Path Item Object](https://spec.openapis.org/oas/v3.2.0.html#fixed-fields-6) so this can be described in the OpenAPI document. This change makes QUERY endpoints show up correctly in the generated OpenAPI document so client generators and API explorers can consume them.
Note that `query` is only valid in an OpenAPI 3.2 document, so you'll want to set the `OpenApiVersion` in the `OpenApiOptions`. In earlier OpenAPI versions, the `query` operation is generated within a `x-oai-additionalOperations` specification extension in the Path Item Object.
```

release-notes/11.0/preview/preview4/aspnetcore.md
```diff
@@ -0,0 +1,246 @@
...
+```csharp
+var app = WebApplication.Create();
+
+app.MapMethods("/search", ["QUERY"], (SearchRequest request) =>
+    SearchService.Run(request));
+
+app.Run();
+```
+
+The method shows up in generated OpenAPI documents alongside `GET`, `POST`, and friends, so client generators and API explorers pick it up automatically.
```
mikekistler:
```suggestion
The method shows up in generated OpenAPI documents alongside `get`, `post`, and friends, so client generators and API explorers pick it up automatically.
```

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: Updated in 83d32ce0 — applied @halter73's ordering fix and @DeagleGross's combined snippet so both TLS features are demoed together.

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: Fixed in 83d32ce0 — `UseTlsClientHelloListener` is now called before `UseHttps()`, with a comment noting why.

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: Applied in 83d32ce0, combined with @DeagleGross's expansion below so the snippet demonstrates both `ITlsHandshakeFeature.Exception` and `UseTlsClientHelloListener` in one example.

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: Applied in 83d32ce0 (with the small "was has" → "has" typo fix). Dropped the short-circuit sentence and added the 10.0.8 backport note.

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: Replaced "first-class" with "known" in 83d32ce0 — folded into @mikekistler's expanded suggestion below that also covers the OpenAPI 3.2 `query` Path Item field.

danroth27: Applied in 83d32ce0 — thanks for the OpenAPI 3.2 context and the `OpenApiVersion` / `x-oai-additionalOperations` callout.

release-notes/11.0/preview/preview4/aspnetcore.md
danroth27: Applied in 83d32ce0 — lowercased to `get`/`post` to match OpenAPI spec casing.

mikekistler (commented):
Looks good. 👍 

I spent a bunch of time looking at the File results type item. I think it is correct but I think we will want to expand on this when we put this into the docs.

PR #10447, opened 2026-06-10
Add foundational security documents

GrabYourPitchforks:
Adds the first two documents we use to frame how we respond to incoming vulnerability reports:
- baseline security assumptions, which covers what we consider to be within our authority boundary
- vulnerability theory, which covers our definition of a vulnerability

There are some dead links at the moment because I need to clean up the other docs before publishing them, but IMO the dead links are ok for now. They don't meaningfully take away from the substance.


blowdart:
@GrabYourPitchforks Presumably you meant to add art and I as reviewers rather than assignees?

GrabYourPitchforks:
@blowdart yes. I fail at UI :(

blowdart (approved):

Documentation/security-foundations/README.md
```diff
@@ -0,0 +1,13 @@
+# .NET security framing guides
+
+This folder contains a set of framing documents guiding the .NET team's security posture. These guides cover how we perform security designs for our components, the contracts these components have with users, and how we assess reported vulnerabilities in these components. They are specific to .NET's unique shape of being a provider of reusable libraries and SDK components rather than a standalone deployed application.
+
+This folder also contains _prototype_ skills used to generate security design documents / threat models and to assess incoming vulnerability reports.
```
artl93, line 5: Not seeing them in this PR.

Documentation/security-foundations/baseline-security-assumptions.md
```diff
@@ -0,0 +1,279 @@
...
+
+> [!NOTE]
+>
+> This analysis is from the perspective _of the debuggee._ A debugger who attaches to a target running under a different authority or who processes a memory dump originating from a different authority **must not** make trust assumptions about the target.
+
+This extends to the contents of the dumps themselves: an operator in possession of a dump is considered fully trusted. That operator can restore the process from the snapshot and can control execution flow, including potentially exercising some of the same authority (e.g., accessing the same resources) as the original process. Because there is no general way for a component to identify and scrub all privilege-granting state from a dump, we must assume that in the extreme, an operator with access to the dump contents can exercise equivalent authority as the original process.
```
artl93, line 65: The inverse is not true though - the diagnostic process should NOT trust the dump, correct?

artl93 (approved):

Documentation/security-foundations/baseline-security-assumptions.md
GrabYourPitchforks: Depends on what robustness claims the diagnostic process wants to make.

_From the perspective of the dump itself,_ the operator of the diagnostic tool is trusted. The diagnostic tool is within the target process's authority boundary. This applies even post-hoc: the diagnostic tool is within the process's authority boundary, even before the dump has been captured, because _the possibility_ to capture a dump exists. (The vuln-theory doc discusses this a bit more as well: _potential_ is a satisfying gate.)

But authority boundaries are asymmetric.

_From the perspective of the debugging tool,_ it can choose to trust the dump, or it can choose not to trust the dump.

Documentation/security-foundations/README.md
GrabYourPitchforks: My fault. :(

I forgot to remove that line before copying the docs into the public repo. Will clean it up in post so as not to block this merge.

PR #10449, opened 2026-06-13
remove the description about the polyfill for union

WeihanLi:
remove the description about the polyfill for union since it's already included into the base library since .NET 11 Preview 4


BillWagner (approved):
I'll defer to @leecow and @rbhanda but this LGTM.

PR #10456, opened 2026-06-29
[release-notes] ASP.NET Core in .NET 11 Preview 6

danroth27:
Draft ASP.NET Core release notes for .NET 11 Preview 6. This PR targets the milestone base branch `release-notes/11.0-preview6` (see #10455).

Please review the AI-authored content for accuracy, including any code samples and `<!-- TODO -->` / `<!-- Filtered features -->` notes, then mark ready for review. 🤖 Generated with the release-notes skill.


danroth27:
Verified every feature in these notes by building and running samples against the locally installed .NET 11 Preview 6 SDK (build `26325.125`). Samples live in `c:\github\danroth27\aspnetcore11samples`. Corrections pushed in the latest commit:

**Async validation** (ran a minimal API; valid → 200, async rules → 400):
- `IAsyncValidatableObject` extends `IValidatableObject`, so the model must also implement the synchronous `Validate` method — the original sample didn't compile without it.
- `AsyncValidationAttribute` is overridden via the protected **`IsValidAsync`** (the doc said `GetValidationResultAsync`, which is the public method the framework *calls* and isn't overridable).

**Blazor Virtualize** (verified in a headless browser: `InitialIndex=500` opened at items 485–515; "Back to top" → `ScrollToIndexAsync(0)` scrolled to Product 0):
- `Virtualize.Items` requires `ICollection<T>`, so the sample now uses `List<Product>` (`IReadOnlyList<Product>` does not compile).

**`[ShortCircuit]`** (verified middleware is skipped):
- Changed the `robots.txt` example from `[ShortCircuit(404)]` to `[ShortCircuit]`. `[ShortCircuit(404)]` returns the body with a **404** status, which is wrong for robots.txt; `[ShortCircuit]` returns 200. Added a minimal-API example.

**Verified accurate as written (no change):**
- OpenAPI: document version is `3.2.0` by default; a union return type produces `anyOf: [Kitten, Puppy]` with bare-named case components.
- SignalR auth refresh: `EnableAuthenticationRefresh`, `OnAuthenticationRefresh`, and `Hub.OnAuthenticationRefreshedAsync` compile, and the `/refresh` endpoint is mapped alongside `/negotiate`.
- Breaking changes: the `IValidatableInfo` split into `IValidatableTypeInfo`/`IValidatableParameterInfo`/`IValidatablePropertyInfo` is present in the build.

Note: the local build is `26325.125`; the notes target `26328.106`. I confirmed the corrected APIs are identical across both builds.

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,172 @@
...
+}
+```
+
+The same attribute works on minimal API endpoints, and the existing `ShortCircuit()` convention continues to work unchanged:
+
+```csharp
+app.MapGet("/health", [ShortCircuit] () => "Healthy");
+```
+
+## SignalR refreshes authentication on long-lived connections
```
BrennanConroy: Might want to include a little about the client-side as well. I did a short write-up on the github issue
https://github.com/dotnet/aspnetcore/issues/5297#issuecomment-4814057160

BrennanConroy: https://github.com/dotnet/aspnetcore/pull/64098 was also done in preview6. Might be worth a small section.

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,172 @@
...
+
+public class ReservationRequest : IAsyncValidatableObject
+{
+    [Required]
+    public string Email { get; set; } = "";
+
+    public DateOnly Date { get; set; }
+
+    // IValidatableObject (synchronous) — no synchronous rules here.
+    public IEnumerable<ValidationResult> Validate(ValidationContext context) => [];
```
Youssef1313: throw instead?

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,172 @@
...
+    private Virtualize<Product> list = default!;
+    private List<Product> products = ProductCatalog.All;
+
+    private async Task GoToTop() => await list.ScrollToIndexAsync(0);
+}
+```
+
+Out-of-range indexes are clamped to the valid range. If a second `ScrollToIndexAsync` call starts while one is still in flight, the last call wins. Calling `ScrollToIndexAsync` before the first interactive render throws `InvalidOperationException`; use `InitialIndex` to set the starting position instead.
+
+## OpenAPI documents default to 3.2 and describe union types
```
Youssef1313: Do we need this to be its own heading? Or should it move under the breaking changes heading?

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,172 @@
+<!-- Verified against Microsoft.AspNetCore.App.Ref@11.0.0-preview.6.26328.106 and Microsoft.NETCore.App.Ref@11.0.0-preview.6.26328.106 -->
+# ASP.NET Core in .NET 11 Preview 6 - Release Notes
+
+.NET 11 Preview 6 includes new ASP.NET Core features and improvements:
+
+- [Minimal API validation runs async validators](#minimal-api-validation-runs-async-validators)
+- [Blazor Virtualize can scroll to an item](#blazor-virtualize-can-scroll-to-an-item)
+- [OpenAPI documents default to 3.2 and describe union types](#openapi-documents-default-to-32-and-describe-union-types)
```
BrennanConroy: I think this was the only area we made product changes for unions. But it might be worth having a section that explicitly states that unions work in ASP.NET Core, possibly cross-reference to .NET release notes which has specifics for configuring?

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,172 @@
+<!-- Verified against Microsoft.AspNetCore.App.Ref@11.0.0-preview.6.26328.106 and Microsoft.NETCore.App.Ref@11.0.0-preview.6.26328.106 -->
+# ASP.NET Core in .NET 11 Preview 6 - Release Notes
+
+.NET 11 Preview 6 includes new ASP.NET Core features and improvements:
+
+- [Minimal API validation runs async validators](#minimal-api-validation-runs-async-validators)
```
danroth27: Change to: Async validation for minimal APIs.

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: It's upgraded functionality, so I think it makes sense to cover defaulting to OpenAPI 3.2 as a feature.

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,164 @@
...
+- [Blazor Virtualize can scroll to an item](#blazor-virtualize-can-scroll-to-an-item)
+- [OpenAPI documents default to 3.2 and describe union types](#openapi-documents-default-to-32-and-describe-union-types)
```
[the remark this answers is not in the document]
danroth27: Fixed by splitting this into two shorter sections: **OpenAPI 3.2 by default** and **Unions in ASP.NET Core**. TOC anchors updated to match.

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,164 @@
...
+    private Virtualize<Product> list = default!;
+    private IReadOnlyList<Product> products = ProductCatalog.All;
+
+    private async Task GoToTop() => await list.ScrollToIndexAsync(0);
+}
+```
+
+Out-of-range indexes are clamped to the valid range. If a second `ScrollToIndexAsync` call starts while one is still in flight, the last call wins. Calling `ScrollToIndexAsync` before the first interactive render throws `InvalidOperationException`; use `InitialIndex` to set the starting position instead.
+
+## OpenAPI documents default to 3.2 and describe union types
```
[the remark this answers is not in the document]
danroth27: Split into two shorter headings: OpenAPI 3.2 by default and Unions in ASP.NET Core.

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Good call — changed to 	hrow. I verified on the final Preview 6 build that async minimal API validation still runs (the async path only calls ValidateAsync), so throwing from the sync Validate surfaces accidental synchronous validation instead of silently skipping the async rules.

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Kept it as a feature per Dan's note, but split it: OpenAPI 3.2 by default is now its own section, and union support moved into a broader Unions in ASP.NET Core section.

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Done — renamed to **Async validation for minimal APIs**.

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Done — OpenAPI 3.2 by default is now its own feature section, separate from unions.

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Added the client-side story. There's now a WithAuthenticationRefresh example on the .NET client (EnableAutoRefresh, RefreshBeforeExpiration, OnAuthenticationRefreshed, OnAuthenticationRefreshFailed), plus a note that it's the .NET client only in this preview with the JS/TS client and Azure SignalR Service in progress. Thanks for the write-up — I pulled the exact API from the checked-in source (the shipped shape is a configure lambda rather than 
ew()).

danroth27: Added a **Cancel hub invocations from the client** section for #64098. I verified the round-trip on the final Preview 6 build: passing a CancellationToken to InvokeAsync and cancelling it triggers the hub method's CancellationToken on the server (previously streaming-only).

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Done — added a broad **Unions in ASP.NET Core** section covering Minimal APIs, MVC/Razor Pages, SignalR, and Blazor, with cross-references to the C# and libraries release notes for the language feature and STJ configuration. I researched the PRs to confirm your point: OpenAPI (#67001) is the only area with intentional product code; Blazor had a null-active-case prerendering fix (#67296); the rest "just work" through System.Text.Json. Also documented the current gaps (non-body binding #66648; MessagePack/Newtonsoft hub protocols).

danroth27:
Pushed an update addressing the review feedback. Summary:

**Restructure**
- Renamed **Async validation for minimal APIs** (+ cross-link to the libraries notes; the sync `Validate` now throws for async-only types).
- Split the OpenAPI section into **OpenAPI 3.2 by default** (kept as a feature) and a broader **Unions in ASP.NET Core** section (Minimal APIs, MVC/Razor Pages, SignalR, Blazor + gaps + cross-refs).
- SignalR: added the **client-side `WithAuthenticationRefresh`** API, and a new **Cancel hub invocations from the client** section (#64098).

**Re-verified on the final Preview 6 build (`11.0.100-preview.6.26359.118`)** — every feature was rebuilt/run in [danroth27/aspnetcore11samples](https://github.com/danroth27/aspnetcore11samples): async validation (incl. the `throw` change), OpenAPI 3.2 + union `anyOf`, `[ShortCircuit]`, Virtualize (browser-tested `InitialIndex`/`ScrollToIndexAsync`), SignalR auth refresh (client compiles), and hub-invocation cancellation (round-trip: client cancel → server token fires).

**Community contributors** — vetted all 12 for Microsoft/.NET-team affiliation and removed two:
- **@jesuszarate** — company `@microsoft` (ANCM signing work). Microsoft employee.
- **@Alex-Sob** — Akvelon, Inc. (a firm that staffs contractors on the .NET/ASP.NET Core team); the merged PRs are internal analyzer work. Removed as a judgment call — let me know if you'd prefer to keep them.

The other 10 (Xebia, Syncfusion, IBM, Cloudflare, and independents) are external and were kept. @desjoerd and @sheiksyedm are public members of the `dotnet` org but work for Xebia/Syncfusion (not Microsoft), so I kept them.

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,237 @@
...
+        options.RefreshBeforeExpiration = TimeSpan.FromMinutes(1);
+        options.OnAuthenticationRefreshed = context => Task.CompletedTask;
+        options.OnAuthenticationRefreshFailed = context => Task.CompletedTask;
+    })
+    .Build();
+```
+
+## Cancel hub invocations from the client
+
+The SignalR client can now cancel a regular, non-streaming hub method invocation ([dotnet/aspnetcore #64098](https://github.com/dotnet/aspnetcore/pull/64098)). Previously only streaming invocations could be cancelled from the client. Now, when you pass a `CancellationToken` to `InvokeAsync` and cancel it, the client sends a cancellation message and the hub method's `CancellationToken` parameter is triggered on the server.
```
BrennanConroy:
```suggestion
The SignalR client can now cancel a regular, non-streaming hub method invocation ([dotnet/aspnetcore #64098](https://github.com/dotnet/aspnetcore/pull/64098)). Previously only streaming invocations could be canceled from the client. Now, when you pass a `CancellationToken` to `InvokeAsync` and cancel it, the client sends a cancellation message and the hub method's `CancellationToken` parameter is triggered on the server.
```

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,237 @@
...
+
+```csharp
+// Client — cancelling the token cancels the server-side invocation.
```
BrennanConroy:
```suggestion
// Client — canceling the token cancels the server-side invocation.
```

danroth27:
Evaluated the 12 PRs against the final Preview 6 build (`11.0.100-preview.6.26359.118`), building a sample for each candidate in [aspnetcore11samples](https://github.com/danroth27/aspnetcore11samples).

**Added (verified in the build):**
| PR | What | Added as |
|----|------|----------|
| #66585 | Auto-enabled cross-origin (CSRF) protection via `Sec-Fetch-Site`/`Origin` | Feature section (this is the *feature* behind the existing #67082 breaking-change note) |
| #67337 | Configure browser behavior from the server (`WithBrowserOptions`) | Feature section |
| #67369 | `EnvironmentView` component (env-conditional content) | Feature section |
| #67368 | `NavigationManager.GetUriWithFragment` | Folded into the EnvironmentView section |
| #66919 | `dotnet user-jwts --file` for file-based apps | Feature section |
| #65876 | Blazor Hybrid `@rendermode` no longer throws in WebView | Bug fixes |
| #63276 | `QUERY` added to hosting metrics known methods | Bug fixes |

**Already documented:** #66753 (Virtualize), #64098 (hub-invocation cancellation).

**Excluded (with reason):**
- #67130 (media components → `Microsoft.AspNetCore.Components.Media`) — the package only publishes up to `26324`; there's no build matching `26359`, so it's not cleanly shipping. Recommend holding until the packaging settles.
- #67037 (Blazor `Label`/`DisplayName`) — narrow form helpers, still actively in API review.
- #66527 (`WebApplicationFactory.ConfigureHostApplicationBuilder`) — niche integration-testing API.

Two API-shape notes worth flagging (found by building samples, not from PR bodies): `CsrfProtectionResult` ships as a **class** with `Allowed()`/`Denied()` factory methods (not an enum), and `ICsrfProtection` exposes **`ValidateAsync`** returning `ValueTask<CsrfProtectionResult>`.

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,300 @@
...
+
+Apps built with `WebApplicationBuilder` now automatically reject unsafe cross-origin requests based on the browser's `Sec-Fetch-Site` and `Origin` headers ([dotnet/aspnetcore #66585](https://github.com/dotnet/aspnetcore/pull/66585)). This lightweight cross-site request forgery (CSRF) protection is on with no configuration and applies across Minimal APIs, MVC, Razor Pages, and Blazor. Same-origin and user-initiated requests are allowed; a cross-origin request that tries to consume a form is rejected.
+
+The check runs as auto-injected middleware after routing and authentication, and records its result on `IAntiforgeryValidationFeature`. Framework form consumers (MVC, minimal API `[FromForm]`, and Blazor SSR) enforce the verdict when they read the form — see the related [breaking change](#breaking-changes). To customize or replace the policy, register an `ICsrfProtection` implementation:
+
+```csharp
+public sealed class MyCsrfProtection : ICsrfProtection
+{
+    public ValueTask<CsrfProtectionResult> ValidateAsync(HttpContext context) =>
+        ValueTask.FromResult(CsrfProtectionResult.Allowed());
```
BrennanConroy: Might want to be a little careful here since this is a security primitive. At a minimum should probably comment that this is unsafe. Also, there is a way to opt-out of the auto-injected middleware.

danroth27:
Reframed the Blazor preview API changes per the guidance that **preview-to-preview renames aren''t breaking changes** (we haven''t shipped these APIs in a stable release yet).

**New `Preview API changes` section** (migration notes for upgrading from an earlier .NET 11 preview) replaces the old `Breaking changes` section:
- `EnvironmentBoundary` → `EnvironmentView` — this component was documented back in [Preview 1](https://github.com/dotnet/core/blob/main/release-notes/11.0/preview/preview1/aspnetcore.md#environmentboundary-component), so this is a rename (dotnet/aspnetcore#67369)
- `NavigationManager.GetUriWithHash` → `GetUriWithFragment` (dotnet/aspnetcore#67368)
- Browser options reshape: `WithBrowserConfiguration`→`WithBrowserOptions`, `BrowserConfiguration`→`BrowserOptions`, `ServerBrowserOptions`→`InteractiveServerBrowserOptions`, `DisableDomPreservation`→`PreserveDom`, `CircuitInactivityTimeoutMs`→`CircuitInactivityTimeout` (dotnet/aspnetcore#67337)
- Media components moved out of `Components.Web` to a new `Components.Media` package (dotnet/aspnetcore#67130)
- Minimal API validation resolver split (`IValidatableInfo` → three interfaces; dotnet/aspnetcore#67183) — experimental API, grouped here rather than as a break

**Removed** the two Blazor feature sections that re-documented the renamed APIs — we don''t need to re-document a renamed feature, just note the rename.

**Automatic cross-origin (CSRF) protection** is now documented purely as **new functionality**; I folded the shared-verdict detail (dotnet/aspnetcore#67082) into that feature section instead of listing it as a breaking change.

Net result: there is no `Breaking changes` section this preview — nothing here breaks a shipped-stable API.

⚠️ **Media move caveat (dotnet/aspnetcore#67130):** the components are removed from `Microsoft.AspNetCore.Components.Web` in the Preview 6 build (`26359.118`), but the new `Microsoft.AspNetCore.Components.Media` package''s earliest published build is `11.0.0-preview.7.26360.111` — there is no matching preview.6 package. I documented it as "removed from Web, moving to a new package" without pointing to a specific `PackageReference`. Happy to adjust if you''d prefer different handling.

danroth27:
Reviewed the **Automatic cross-origin (CSRF) protection** section against the new draft docs ([reference](https://learn.microsoft.com/aspnet/core/security/csrf-protection?view=aspnetcore-11.0), [migration](https://learn.microsoft.com/aspnet/core/migration/antiforgery-to-csrf?view=aspnetcore-11.0)) and improved it. Every claim is verified against the Preview 6 source (`release/11.0.1xx-preview6`: `DefaultCsrfProtection`, `CsrfProtectionMiddleware`, `WebApplicationBuilder`) **and** the running `26359.118` build.

What changed:
- **Added the primary upgrade action** — allow a trusted cross-origin caller by declaring its origin in a CORS policy. The middleware reuses the endpoint''s CORS policy; `AllowAnyOrigin` is intentionally **not** trusted for writes.
- **Added opt-outs** — `.DisableAntiforgery()` / `[IgnoreAntiforgeryToken]` per endpoint, and the `DisableCsrfProtection` global config key.
- **Clarified deferred enforcement** — endpoints that never read a form (e.g. JSON APIs) aren''t rejected.
- **Corrected ordering** — "after authentication and authorization" (was "after routing and authentication").
- **Linked the new conceptual + migration docs.**

Runtime check against the installed P6 build (curl, `Sec-Fetch-Site` set manually):

| Scenario | Result |
| --- | --- |
| same-origin form POST | 200 |
| cross-site untrusted origin, minimal API `[FromForm]` | **400** |
| cross-site origin trusted via CORS default policy | 200 |
| non-browser (no `Sec-Fetch-Site`/`Origin`) | 200 |
| cross-site untrusted, JSON body (no form read) | 200 |

One nuance worth knowing: reading `HttpContext.Request.Form` **directly** is a backstop that throws `InvalidOperationException` rather than returning a clean 400. The clean 400 comes from framework form consumers (minimal API `[FromForm]`, MVC, Blazor SSR), which is exactly what the note describes.

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Applied the `canceled` spelling — thanks!

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Applied `canceling` in the comment — thanks!

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Good call — this is a security primitive. I removed the custom `ICsrfProtection` sample that returned `Allowed()` for everything (an unsafe example), and the section now documents opting out of the auto-injected middleware: `.DisableAntiforgery()` / `[IgnoreAntiforgeryToken]` per endpoint, and the `DisableCsrfProtection` config key to disable it app-wide. Custom trust logic still goes through `ICsrfProtection`.

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,260 @@
...
+- [Bug fixes](#bug-fixes)
+- [Community contributors](#community-contributors)
+
+ASP.NET Core updates in .NET 11:
+
+- [What's new in ASP.NET Core in .NET 11](https://learn.microsoft.com/aspnet/core/release-notes/aspnetcore-11)
+
+## Async validation for minimal APIs
+
+Minimal API validation now supports asynchronous validators end-to-end ([dotnet/aspnetcore #66487](https://github.com/dotnet/aspnetcore/pull/66487), [dotnet/aspnetcore #67183](https://github.com/dotnet/aspnetcore/pull/67183)). Preview 5 shipped the building blocks for asynchronous form validation in Blazor. Preview 6 adds the new asynchronous `DataAnnotations` APIs in the base libraries — `AsyncValidationAttribute`, `IAsyncValidatableObject`, and `Validator.ValidateObjectAsync` — and `Microsoft.Extensions.Validation` now runs them when an endpoint validates a request. The new base-library APIs are covered in the [.NET libraries release notes](./libraries.md#asynchronous-validation-with-dataannotations).
```
Youssef1313: fwiw, `Validator` class isn't used by Microsoft.Extensions.Validation at all. It's only used by Blazor when Microsoft.Extensions.Validation isn't wired up.

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,260 @@
...
+
+This lets a validation rule do real work, such as a database lookup or a remote API call, without blocking a thread. A model implements `IAsyncValidatableObject` and returns validation results as an `IAsyncEnumerable<ValidationResult>`. Because `IAsyncValidatableObject` extends `IValidatableObject`, also implement the synchronous `Validate` method. When a type validates asynchronously only, throw from `Validate` so it isn't silently validated through the synchronous APIs:
```
Youssef1313: Should this say "isn't silently **skipped**"?

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,260 @@
...
+public class ReservationRequest : IAsyncValidatableObject
+{
+    [Required]
+    public string Email { get; set; } = "";
+
+    public DateOnly Date { get; set; }
+
+    // IValidatableObject (synchronous) — this type validates asynchronously only.
+    public IEnumerable<ValidationResult> Validate(ValidationContext context) =>
+        throw new NotSupportedException("Validate this type with ValidateAsync.");
```
Youssef1313: The docs suggest `InvalidOperationException` instead of `NotSupportedException`.

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,260 @@
...
+
+    private async Task GoToTop() => await list.ScrollToIndexAsync(0);
+}
+```
+
+Out-of-range indexes are clamped to the valid range. If a second `ScrollToIndexAsync` call starts while one is still in flight, the last call wins. Calling `ScrollToIndexAsync` before the first interactive render throws `InvalidOperationException`; use `InitialIndex` to set the starting position instead.
+
+## OpenAPI 3.2 by default
+
+Generated OpenAPI documents now target OpenAPI 3.2 by default ([dotnet/aspnetcore #67097](https://github.com/dotnet/aspnetcore/pull/67097), [dotnet/aspnetcore #67007](https://github.com/dotnet/aspnetcore/pull/67007)). OpenAPI 3.2 is the newest version of the specification. Documents continue to generate as before; set the document version explicitly if you need to target an earlier version for tooling that hasn't adopted 3.2 yet.
```
Youssef1313: I'm having a hard time understanding what "Documents continue to generate as before" means

Youssef1313 (approved):
LGTM for the parts in my areas. Only left few comments.

Also are we intentionally missing https://github.com/dotnet/aspnetcore/pull/66527 ?

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Automatic cross-origin (CSRF) protection
+
+Apps built with `WebApplication.CreateBuilder` now automatically reject unsafe cross-origin requests based on the browser's `Sec-Fetch-Site` and `Origin` headers ([dotnet/aspnetcore #66585](https://github.com/dotnet/aspnetcore/pull/66585), [dotnet/aspnetcore #67082](https://github.com/dotnet/aspnetcore/pull/67082)). This lightweight cross-site request forgery (CSRF) protection is on with no configuration and applies across Minimal APIs, MVC, Razor Pages, and Blazor. Same-origin requests, user-initiated navigations, and non-browser clients are allowed, while a cross-origin browser request that tries to consume a form is rejected. It runs alongside the existing token-based antiforgery system.
+
+To opt an endpoint out, use `.DisableAntiforgery()` (Minimal APIs) or `[IgnoreAntiforgeryToken]` (MVC); to turn the feature off app-wide, set the `DisableCsrfProtection` configuration key. For full control over the trust decision, register a custom `ICsrfProtection` implementation.
+
```
DeagleGross, line 102: Section is good! Should we also point users to main article here?

```suggestion

For more information see [Automatic CSRF protection in ASP.NET Core](https://learn.microsoft.com/aspnet/core/security/csrf-protection?view=aspnetcore-11.0&viewFallbackFrom=aspnetcore-10.0)

```

release-notes/11.0/preview/preview6/aspnetcore.md
```diff
@@ -0,0 +1,260 @@
...
+// Requires <LangVersion>preview</LangVersion>
+public record class Dog(string Name);
+public record class Cat(int Lives);
+public union Pet(Dog, Cat);
+
+// The active case is serialized on the way out and described with anyOf in OpenAPI.
+app.MapGet("/pets/{id}", Pet (int id) => id == 0 ? new Dog("Rex") : new Cat(9));
+```
+
+For **OpenAPI**, an endpoint that returns a union is described with an `anyOf` schema listing each case type ([dotnet/aspnetcore #67001](https://github.com/dotnet/aspnetcore/pull/67001)). Unlike polymorphic types, union cases don't carry a `$type` discriminator, so a case such as `Dog` reuses the standalone `#/components/schemas/Dog` component instead of a duplicated, prefixed one. ApiExplorer detects a union through `JsonTypeInfoKind.Union`, so the schema also flows through to Swashbuckle and NSwag.
```
DeagleGross, line 150:
```suggestion
For **OpenAPI**, an endpoint that returns a union is described with an `anyOf` schema listing each case type ([dotnet/aspnetcore #67001](https://github.com/dotnet/aspnetcore/pull/67001)). Unlike polymorphic types, union cases don't carry a `$type` discriminator, so a case such as `Dog` reuses the standalone `#/components/schemas/Dog` component instead of a duplicated, prefixed one.
```

1. ApiExplorer does not detect a union via `JsonTypeInfoKind.Union`
2. "Schema also flows through to Swashbuckle and NSwag." - also not true, because Swashbuckle (e.g.) consumes ApiExplorer output (`ApiDescription`). And it is yet to be confirmed to be working (or requires a code change) on Swashbuckle side (see https://github.com/dotnet/aspnetcore/issues/66544#issuecomment-4622444161)   

I would just drop this phrase. If we want to keep info on how external providers treat unions we probably can say:
```suggestion
For **OpenAPI**, an endpoint that returns a union is described with an `anyOf` schema listing each case type ([dotnet/aspnetcore #67001](https://github.com/dotnet/aspnetcore/pull/67001)). Unlike polymorphic types, union cases don't carry a `$type` discriminator, so a case such as `Dog` reuses the standalone `#/components/schemas/Dog` component instead of a duplicated, prefixed one. Third-party generators such as Swashbuckle and NSwag do not yet recognize unions and will produce their default schema shape
```
```

File: release-notes/11.0/preview/preview6/aspnetcore.md (hunk 14 above)
Comment (guardrex) [reply release-notes/11.0/preview/preview6/aspnetcore.md]: Pinging @tdykstra for an 👁️ on this suggestion because we're mirroring this language in the [PR for ASP.NET Core *What's New* coverage (dotnet/AspNetCore.Docs #37334)](https://github.com/dotnet/AspNetCore.Docs/pull/37334).

File: release-notes/11.0/preview/preview6/aspnetcore.md (hunk 11 above)
Comment (danroth27) [reply release-notes/11.0/preview/preview6/aspnetcore.md]: Good catch. Reworded so the note no longer implies `Microsoft.Extensions.Validation` uses the `Validator` static class. It now lists the two APIs M.E.Validation actually runs (`AsyncValidationAttribute`, `IAsyncValidatableObject`) and points to the libraries release notes for `Validator.ValidateObjectAsync`.

File: release-notes/11.0/preview/preview6/aspnetcore.md (hunk 12 above)
Comment (danroth27) [reply release-notes/11.0/preview/preview6/aspnetcore.md]: Updated to `silently skipped` — thanks.

File: release-notes/11.0/preview/preview6/aspnetcore.md (hunk 13 above)
Comment (danroth27) [reply release-notes/11.0/preview/preview6/aspnetcore.md]: Changed the sample to throw `InvalidOperationException` to match the docs.

File: release-notes/11.0/preview/preview6/aspnetcore.md (hunk 14 above)
Comment (danroth27) [reply release-notes/11.0/preview/preview6/aspnetcore.md]: Reworded to make clear only the default document version changes; your generation code and configuration are unchanged. Since the docs (dotnet/AspNetCore.Docs#37334) mirror this language, I'll leave this thread open to align with whatever final wording you and @tdykstra land on. cc @guardrex

File: release-notes/11.0/preview/preview6/aspnetcore.md [hunk 17]
```diff
@@ -0,0 +1,287 @@
...
+
+@code {
+    private Virtualize<Product> list = default!;
+    private List<Product> products = ProductCatalog.All;
+
+    private async Task GoToTop() => await list.ScrollToIndexAsync(0);
+}
+```
+
+Out-of-range indexes are clamped to the valid range. If a second `ScrollToIndexAsync` call starts while one is still in flight, the last call wins. Calling `ScrollToIndexAsync` before the first interactive render throws `InvalidOperationException`; use `InitialIndex` to set the starting position instead.
```
ilonatommy, line 125:
```suggestion
Out-of-range indexes are clamped to the valid range. If a second `ScrollToIndexAsync` call starts while one is still in flight, the last call wins. If user scrolls during `ScrollToIndexAsync` call, user scroll wins. Calling `ScrollToIndexAsync` before the first interactive render throws `InvalidOperationException`; use `InitialIndex` to set the starting position instead.
```

That's the only update I would make, otherwise it looks good.

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Good catch. Since #10456 already merged, I applied this in the base PR #10455 (commit 3de5ef8a): dropped the incorrect `ApiExplorer`/`JsonTypeInfoKind.Union` claim and now note that Swashbuckle and NSwag don't yet recognize unions and produce their default schema shape. Thanks!

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Thanks! Since #10456 already merged, I applied this in the base PR #10455 (commit 3de5ef8a): added that if the user scrolls during a `ScrollToIndexAsync` call, the user's scroll wins.

release-notes/11.0/preview/preview6/aspnetcore.md
danroth27: Thanks for the suggestion! We're intentionally keeping this section free of a link to the CSRF conceptual doc. That article is expected to be refactored and moved as the feature settles, and we'd rather not ship a release-note link that later breaks. The section already covers the behavior and the opt-outs, and readers can reach the docs from the main *What's new in ASP.NET Core in .NET 11* page. Leaving the text as-is (this applies to the merged content now in #10455).

guardrex:
Let me jump in to mention that if you want to cross-link to the two Blazor node article sections, use the following two links. The Blazor node articles (and changes on the Blazor Pre6 PR) cover the `AntiforgeryToken` component, the `[RequireAntiforgeryToken]` attribute, and the `AntiforgeryStateProvider.GetAntiforgeryToken` method. Let's take up the Pre6 updates on the [Blazor Pre6 PR](https://github.com/dotnet/AspNetCore.Docs/pull/37322). For this PR if you want to include the links, here they are, and this PR won't break because these are existing sections ........

* \<xref:blazor/forms/index#antiforgery-support>
* \<xref:blazor/security/index#antiforgery-support>

UPDATE: I added those links to the Blazor section of ASP.NET Core *What's New* that pertains to the BWA template change (dropping the antiforgery middleware setup), along with the general links to CSRF middleware.

Issue #10213, opened 2026-01-18
Unexpected indent in api diff


OwnageIsMagic:
### URL(s)

https://github.com/dotnet/core/blob/6a389d2680848bb4b891dd37936caf10bbadc665/release-notes/10.0/api-diff/Microsoft.NETCore.App/10.0.0_System.Net.WebSockets.md?plain=1#L27-L29

https://github.com/dotnet/core/blob/6a389d2680848bb4b891dd37936caf10bbadc665/release-notes/10.0/api-diff/Microsoft.NETCore.App/10.0.0_System.Runtime.Intrinsics.md?plain=1#L1226-L1231

PR #10472, opened 2026-07-10
Fix missing indentation in generated API diffs

svick:
The api-diff tool had a bug where the first member of a type could lose its indentation in the generated markdown (dotnet/core#10213). This was caused by a synthesized default constructor lacking a body/semicolon, which fused it with the following member during formatting.

The tool is being fixed in dotnet/sdk#54130, but the already-generated diff files need manual correction. This re-indents the 101 affected member lines across 19 files to align with their sibling members.

Only whitespace (indentation) is changed; no API content is modified.

Contributes to https://github.com/dotnet/core/issues/10213

### DIFF
--- a/release-notes/10.0/api-diff/Microsoft.AspNetCore.App/10.0.0_Microsoft.AspNetCore.Components.md
+++ b/release-notes/10.0/api-diff/Microsoft.AspNetCore.App/10.0.0_Microsoft.AspNetCore.Components.md
@@ -51,7 +51,7 @@
 +     }
 +     public sealed class RestoreContext
 +     {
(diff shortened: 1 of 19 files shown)


jeffhandley (approved):
There are some corrections to nullability annotations on some of the APIs as well, but the updates match what's current so it's good with me to take those changes.

svick:
@jeffhandley This PR does not make any changes to nullability annotations. I have checked and `git diff -w` is empty.

Thanks for the review.

jeffhandley:
https://github.com/dotnet/core/pull/10472/changes#diff-3a7263794916137c3f544aa0b3986d611daa17c434f41cc26c751d2b6eab66c6 is what I saw, but could be I'm misinterpreting that diff.

PR #10469, opened 2026-07-07
Automate public API diff reports for in-development .NET releases

jeffhandley:
## Summary

Adds an "API Diff" automation that keeps public API diff report PRs continuously
up to date for the active, not-yet-shipped .NET majors. A scheduled orchestrator
deterministically discovers which diffs should exist from the dnceng NuGet feeds
and dispatches a per-target worker that generates the reports from real published
builds and opens/maintains a **draft** PR for each one. The automation never
merges -- humans review and merge.

This PR adds only the automation and tooling. No API diff report content is
included; the reports are produced as draft PRs by the workflows after this
merges.

## How it works

**Discovery (deterministic, no agent).** `api-diff-orchestrator.yml` is a plain
workflow whose discover job runs a C# file-based app, `api-diff-discover.cs`,
that uses the NuGet client SDK (NuGet.Protocol + NuGet.Versioning) for
SemVer-correct feed queries. `maxGA` is the highest GA major on dotnet-public;
only `maxGA+1` and `maxGA+2` are considered. It emits one target per diff, each
pinning the exact previous/current package versions and their feeds, and fans out
to the API Diff Worker workflow via `workflow_call` (max-parallel 1, matching the
worker's shared concurrency group so a queued target is never dropped).

Two tracks of diffs are produced:

- **Major-to-major (cumulative).** Current = newest `dotnet{major}` dev-feed build
  of any prerelease label; baseline = newest `(major-1)` RC-or-GA build on
  dotnet-public, advancing RC1 -> RC2 -> GA. Identity is keyed on the major and
  kept canonical (`(major-1).0-ga_to_major.0`) so one evergreen PR is reused as
  the head advances. Content at `release-notes/<major>.0/<major>.0.0/api-diff`.
- **Incremental (milestone deltas).** Head is sourced from the dev feed so a
  milestone's **draft** PR is produced as soon as its build exists (to merge
  before the milestone releases); the baseline is pinned to the previous
  milestone's **shipped** public build, so a milestone's diff holds off until its
  baseline has shipped. Content under
  `release-notes/<major>.0/preview/<milestone>/api-diff`.

**Generation + publishing (agentic worker).** The API Diff Worker workflow,
`api-diff-worker.md` (compiled to `api-diff-worker.lock.yml`), is the per-target
worker. Its host generation script, `api-diff-generate.sh`, positions the working
tree on the publish branch and runs the api-diff skill's `RunApiDiff.ps1` against
[...]

PR #10473, opened 2026-07-10
Capture preview API changes in the release-notes skill

danroth27:
## What

Updates the `release-notes` skill so **preview-to-preview API changes** (renames, reshapes, removals of APIs that haven''t shipped in a stable release) are captured and documented as migration notes, instead of being missed or mislabeled as breaking changes.

Motivated by the .NET 11 Preview 6 notes: `EnvironmentBoundary` (documented in Preview 1) was renamed to `EnvironmentView` in Preview 6, and the automatic CSRF default was initially framed as a breaking change. Preview notes are an upgrade guide from the previous preview, so these renames need migration notes — but a preview-to-preview rename is not a "breaking change" (nothing stable shipped yet).

## Changes

- **`references/editorial-rules.md`** — new "Preview API changes vs. breaking changes" section: use a dedicated `Preview API changes` section for preview renames; reserve `Breaking changes` for version-to-version breaks of shipped-stable APIs; new default behavior is a feature, not a break; check earlier previews for the original name before writing.
- **`references/format-template.md`** — add a `Preview API changes` section to the component template and update the section rule.
- **`references/api-verification.md`** — new "Capturing preview API changes" section describing the previous-preview → current-preview `dotnet-inspect diff` to surface renamed/removed public APIs.
- **`references/changes-schema.md`** — note that `breaking_changes: true` also covers preview API changes, which the writing stage routes to the `Preview API changes` section.
- **`SKILL.md`** — mention preview-API-change capture in the pipeline description.

Docs-only change to the skill; no code or release-notes content is affected.

### DIFF
--- a/.github/skills/release-notes/SKILL.md
+++ b/.github/skills/release-notes/SKILL.md
@@ -15,7 +15,7 @@ This skill is the **editorial writing stage** of the pipeline. It turns a scored
 1. `generate-changes` diffs `source-manifest.json` between VMR refs to produce `changes.json`
 2. `generate-features` reads `changes.json`, resolves revert/backout relationships, and emits `features.json` with optional scores using the shared `editorial-scoring` rubric
 3. `update-existing-branch` handles incremental reruns when a milestone branch already exists, merging deltas instead of restarting from scratch
-4. `api-diff` / `dotnet-inspect` verifies public APIs and confirms suspect features still exist in the shipped build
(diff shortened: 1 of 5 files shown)


jeffhandley:
@danroth27 I'd really like to get [Automate public API diff reports for in-development .NET releases (#10469)](https://github.com/dotnet/core/pull/10469) merged before we merge this; there is some overlap here.

PR #10494, opened 2026-07-24
Allow API diff area owner mentions

svick:
I believe the mentions in API diff PRs were not supposed to be escaped using `` ` ``. This happened, because `safe-outputs.mentions` was not set. This PR adds all current area owners to the `allowed` list.

## Summary

- allow all lead and owner handles from dotnet/runtime docs/area-owners.md
- allow the four static ASP.NET, WinForms, and WPF reviewer teams
- regenerate the compiled agentic workflow

This prevents API diff review-checklist mentions from being escaped with backticks.


jeffhandley (approved):
Great catch and change, @svick!

I revised the list of mentionees by capturing every team and individual the workflow found it could possibly mention for how it finds them, and then verified all the names against the microsoft-team.md documentation we have in this repo.

Then I refactored the safe-output out into a shared import file so that we can reuse the list across workflows if needed and have a single place to maintain the list.

tannergooding (approved):

PR #10501, opened 2026-08-02
[release-notes] .NET 11 Preview 7

danroth27:
Release notes for .NET 11 Preview 7.

This is the **base PR** for the milestone. It carries the shared metadata:
`README.md`, `changes.json`, `features.json`, `build-metadata.json`, and the API diff.

Each component's notes are reviewed in its own PR targeting this branch. As those merge,
this PR's diff becomes the full milestone.

Stays a draft until the milestone ships.

PR #10502, opened 2026-08-02
[release-notes] ASP.NET Core in .NET 11 Preview 7

danroth27:
Component release notes for `aspnetcore.md` in .NET 11 Preview 7.

Targets the milestone base branch `release-notes/11.0-preview7` (see #10501).

This content is AI-authored and needs a review pass from the component team before it goes ready-for-review.


release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+- [`PathString.StartsWithSegments` treats `\` as a segment boundary](#pathstringstartswithsegments-treats--as-a-segment-boundary)
+- [Breaking changes](#breaking-changes)
+- [Bug fixes](#bug-fixes)
+- [Community contributors](#community-contributors)
+
+ASP.NET Core updates in .NET 11:
+
+- [What's new in ASP.NET Core in .NET 11](https://learn.microsoft.com/aspnet/core/release-notes/aspnetcore-11)
+
+## Blazor Server automatic circuit pause
```
danroth27: This section heading is a bit awkward. How about: "Auto pause Blazor circuits on inactivity"?

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+
+Interactive Server circuits can now pause themselves automatically when the browser tab is hidden, releasing SignalR and server-side memory until the user comes back ([dotnet/aspnetcore #67098](https://github.com/dotnet/aspnetcore/pull/67098)). Preview 6 added [`Circuit.RequestCircuitPauseAsync`](https://github.com/dotnet/aspnetcore/issues/64886) so the server could pause a circuit on demand; Preview 7 adds a matching client-driven path that fires after a configurable inactivity delay while a page is hidden.
```
danroth27:
```suggestion
Interactive Server circuits can now pause themselves automatically when the browser tab is hidden, releasing server resources until the user comes back ([dotnet/aspnetcore #67098](https://github.com/dotnet/aspnetcore/pull/67098)). Preview 6 added [`Circuit.RequestCircuitPauseAsync`](https://github.com/dotnet/aspnetcore/issues/64886) so the server could pause a circuit on demand; Preview 7 adds a matching client-driven path that fires after a configurable inactivity delay while a page is hidden.
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+
+Configure it per app on the render mode via `AddAutoPause` on `BrowserOptions`. The behavior ships in the separate `Microsoft.AspNetCore.Components.Server.AutoPause` package, so add a reference to opt in:
```
danroth27:
```suggestion
To enable support for auto pausing circuits, add the `Microsoft.AspNetCore.Components.Server.AutoPause` and   configure it for the app via `AddAutoPause` on `BrowserOptions`:
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+    {
+        options.AddAutoPause(pause =>
+        {
+            pause.Enabled = true; // default
+            pause.HiddenDelay = TimeSpan.FromSeconds(30); // default is 2 minutes
+        });
+    });
+```
+
+Auto-pause defers itself while a pause would lose data or interrupt work in progress. The shipped checks are a text input or `contenteditable` element whose value differs from its default (including inside shadow DOM and same-origin iframes), an `<audio>` or `<video>` element that is playing unmuted, an open [Picture-in-Picture](https://developer.mozilla.org/docs/Web/API/Picture-in-Picture_API) window, a held [Web Lock](https://developer.mozilla.org/docs/Web/API/Web_Locks_API), and any in-flight circuit activity such as a running `IJSRuntime` call, a `DotNetStreamReference`/`JSStreamReference` transfer, or a queued render.
```
danroth27:
```suggestion
Auto-pause defers itself when a pause would lose data or interrupt work in progress. The shipped checks are a text input or `contenteditable` element whose value differs from its default (including inside shadow DOM and same-origin iframes), an `<audio>` or `<video>` element that is playing unmuted, an open [Picture-in-Picture](https://developer.mozilla.org/docs/Web/API/Picture-in-Picture_API) window, a held [Web Lock](https://developer.mozilla.org/docs/Web/API/Web_Locks_API), and any in-flight circuit activity such as a running `IJSRuntime` call, a `DotNetStreamReference`/`JSStreamReference` transfer, or a queued render.
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+```text
+System.InvalidOperationException: Component 'QuickGrid`1[...]' cannot be used inside a CacheView
+because its output depends on per-request state ([CacheBehavior(CacheBehavior.Throw)],
+[CacheCondition(CacheVaryBy.Query)]) that cannot be safely cached and replayed. To fix this,
+configure the CacheView to vary by Query, or move the component outside the CacheView.
+```
+
+## New Blazor analyzers
+
+Preview 7 adds five new analyzers to the Blazor SDK targeting pack. All are on by default and ship as `Warning`:
```
danroth27:
```suggestion
Preview 7 adds five new Blazor analyzers to help improve code quality. All are on by default and ship as `Warning`:
```

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27, line 1: Reduce the number of AI tells, like the excessive us of emdashes.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+| `BL0016` | JS interop call '…' is not guarded with a try/catch block | [dotnet/aspnetcore #67900](https://github.com/dotnet/aspnetcore/pull/67900) |
+
+`BL0012` flags `StateHasChanged()` calls that the framework already schedules for you — for example inside `OnInitializedAsync`, an `EventCallback` handler, or a component parameter setter — and offers a code fix that removes them. `BL0013` catches the common bug where a component reads `GetAuthenticationStateAsync()` once and then serves a stale user because it never subscribed to `AuthenticationStateChanged`. `BL0014` catches a classic Blazor bug where a `for`-loop counter is captured by a lambda used in `@onclick` or a child component's parameter, so every button ends up bound to the final iteration value. `BL0015` flags `[JSInvokable]` methods that aren't `public` — they compile but JavaScript can't call them — and offers a code fix that makes them public. `BL0016` warns when an `IJSRuntime.InvokeAsync` / `InvokeVoidAsync` call is not inside a `try`/`catch`, so a disconnected circuit or a JS exception doesn't tear down the component.
```
danroth27: Can the descriptions be incorporated into the table? Or is there a better way to format this information so that the details for each analyzer are together?

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+
+Four of these were community contributions — see the [community section](#community-contributors) at the end.
```
danroth27: I like to credit the specific contributors directly in the feature section that covers what they contributed if possible.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+```razor
+@* Preview 6 *@
+<Toast Message="@("Saved 3 items.")" />
+
+@* Preview 7 *@
+<Toast Message="Saved 3 items." />
+```
+
+One gap remains: child content between the component's tags still doesn't populate a `RenderFragment` case, so that case has to be supplied as an expression ([dotnet/razor #13200](https://github.com/dotnet/razor/issues/13200)).
+
```
danroth27: It's not clear that the ChildContent case is actually a scenario. Remove.

```suggestion
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+## Blazor SSR client-side form validation reworked
+
+Preview 5 added client-side validation for Blazor static SSR forms: adding `<DataAnnotationsValidator />` inside a static `EditForm` now runs `[Required]`, `[Range]`, `[RegularExpression]`, `[StringLength]`, and similar rules in the browser before the form posts. Preview 7 keeps that feature but reworks the implementation and the API shape after review feedback ([dotnet/aspnetcore #67324](https://github.com/dotnet/aspnetcore/pull/67324), [dotnet/aspnetcore #67855](https://github.com/dotnet/aspnetcore/pull/67855)):
```
danroth27:
```suggestion
Preview 5 added client-side validation for Blazor static SSR forms without requiring an interactive render mode. Preview 7 keeps that feature but reworks the implementation and the API shape after review feedback ([dotnet/aspnetcore #67324](https://github.com/dotnet/aspnetcore/pull/67324), [dotnet/aspnetcore #67855](https://github.com/dotnet/aspnetcore/pull/67855)):
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+    <InputText @bind-Value="Model.Email" />
+    <ValidationMessage For="() => Model.Email" />
+
+    <button type="submit">Book</button>
+</EditForm>
+```
+
+## Validation localization is built in
+
+`Microsoft.Extensions.Validation` localizes validation messages and display names without a separate package. The preview-only `Microsoft.Extensions.Validation.Localization` package and its `AddValidationLocalization<TResource>()` / `IValidationLocalizer` API are removed; localization now activates automatically as soon as an `IStringLocalizerFactory` is registered, and the lookup is emitted by the validation source generator into your assembly ([dotnet/aspnetcore #68005](https://github.com/dotnet/aspnetcore/pull/68005)).
```
danroth27:
```suggestion
`Microsoft.Extensions.Validation` now localizes validation messages and display names without a separate package. The preview-only `Microsoft.Extensions.Validation.Localization` package and its `AddValidationLocalization<TResource>()` / `IValidationLocalizer` API are removed; localization now activates automatically as soon as an `IStringLocalizerFactory` is registered, and the lookup is emitted by the validation source generator into your assembly ([dotnet/aspnetcore #68005](https://github.com/dotnet/aspnetcore/pull/68005)).
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+public sealed class DivisibleByAttribute : ValidationAttribute, IValidationMessageFormatter
+{
+    public int Divisor { get; init; }
+
+    public string FormatMessage(CultureInfo culture, string template, string displayName)
+        => string.Format(culture, template, displayName, Divisor); // {0} = name, {1} = divisor
+}
+```
+
+The same rules apply to minimal API endpoint validation, Blazor interactive form validation, and Blazor SSR client-side validation, so a message localizes identically wherever the model is used.
```
danroth27:
```suggestion
The same localization rules apply to validation for minimal APIs and Blazor, so a message localizes identically wherever the model is used.
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+
+Behind Azure SignalR the app token authenticates against the app server, and the transport uses a separate service token issued by the redirect. Before this change the client posted `/refresh` with the wrong token, ignored any refreshed transport token in the response, and lost the `tokenLifetimeSeconds` value across the redirect. The client now captures the app-token provider before transports overwrite it, sends `/refresh` with the app token, adopts a refreshed `accessToken` from the response when one is present, and preserves the negotiate hop's `tokenLifetimeSeconds` so auto-refresh scheduling keeps working past the original expiry.
+
+Self-hosted SignalR is unaffected — the redirect-only fields are optional and omitted there.
```
danroth27: This section is overstated. Work was done on the SignalR client for .NET to enable auth token refresh behind a redirecting server, like the Azure SignalR Service, but the service itself has not yet enabled this feature. This was a client only change in Preview 7. Announcing updates to Azure services is not in scope for the .NET release notes.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+
+## Consistent authorization metadata across the stack
+
+`[Authorize]` policies aren't the only shape of authorization metadata attached to an endpoint. `AuthorizationPolicy` instances used directly as metadata and `IAuthorizationRequirementData` attributes (which carry requirements without a named policy) are equally valid, and endpoint routing's `AuthorizationMiddleware` has honored all three for years. Preview 7 brings MVC filters, SignalR hub method authorization, and Blazor's `AuthorizeView` / `AuthorizeRouteView` into line so the same custom authorization attribute behaves consistently everywhere ([dotnet/aspnetcore #67765](https://github.com/dotnet/aspnetcore/pull/67765)).
```
danroth27: Tighten this paragraph up and make it more to-the-point.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+        IAuthorizationPolicyProvider policyProvider,
+        IEnumerable<object> metadata);
+}
+```
+
+You typically don't call it yourself: MVC's `AuthorizeFilter`, SignalR's hub method dispatcher, and Blazor's authorize view components all now route through it, so a single attribute that implements both `IAuthorizeData` and `IAuthorizationRequirementData` contributes to each side of the decision exactly once. The legacy non-endpoint-routing MVC path (`EnableEndpointRouting = false`) is deliberately not updated.
+
+## OpenAPI Server-Sent Events in OpenAPI 3.2
+
+Endpoints that return `SseItem<T>` are now described in the generated OpenAPI document with the OpenAPI 3.2 `itemSchema` shape for `text/event-stream` responses ([dotnet/aspnetcore #67461](https://github.com/dotnet/aspnetcore/pull/67461)). Preview 6 made [OpenAPI 3.2 the default](../preview6/aspnetcore.md#openapi-32-by-default); Preview 7 uses that headroom to describe a stream's per-event payload shape instead of falling back to a plain `string` schema.
```
danroth27: Simplify?

```suggestion
Endpoints that return `SseItem<T>` are now described in the generated OpenAPI document with the OpenAPI 3.2 `itemSchema` shape for `text/event-stream` responses ([dotnet/aspnetcore #67461](https://github.com/dotnet/aspnetcore/pull/67461)). The `itemSchema` describes a stream's per-event payload shape instead of falling back to a plain `string` schema.
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+{
+    foreach (var todo in Todos.All)
+    {
+        yield return new SseItem<Todo>(todo) { EventId = todo.Id.ToString() };
+        await Task.Delay(1000, ct);
+    }
+}
+```
+
+Route the stream through `TypedResults.ServerSentEvents`. Returning an `IAsyncEnumerable<SseItem<T>>` directly from the handler is not an SSE response — minimal APIs serialize it as a JSON array, and the document describes it that way. Note also that `ServerSentEvents` has a dedicated `IAsyncEnumerable<SseItem<T>>` overload that takes no `eventType`; passing `eventType:` alongside `SseItem<T>` items instead binds `ServerSentEvents<T>(IAsyncEnumerable<T>, string?)` with `T` inferred as `SseItem<Todo>`, which serializes the whole envelope into `data`. Use `TypedResults.ServerSentEvents(items, eventType: "todo")` with a plain `IAsyncEnumerable<Todo>` when one event name covers the stream.
```
danroth27: Tighten and simplify.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+
+    await next(context);
+});
+```
+
+Kestrel returns the binding from `SslStream.TransportContext.GetChannelBinding`. IIS and HTTP.sys return it from the request; on HTTP.sys the new `HttpSysOptions.Authentication.HardeningLevel` (defaults to `Medium`) controls whether the OS is asked to enforce channel binding, and setting it to `Strict` now fails startup if the OS cannot apply that hardening rather than silently degrading ([dotnet/aspnetcore #67720](https://github.com/dotnet/aspnetcore/pull/67720)).
+
+## Kestrel and IIS hardening
+
+A cluster of security fixes closed protocol-level gaps in Kestrel, HTTP/2, HTTP/3, and the ASP.NET Core Module.
```
danroth27: I don't think these are security fixes in the traditional sense. This is defense in depth work to harden the stack.




```suggestion
A cluster of fixes closed protocol-level gaps in Kestrel, HTTP/2, HTTP/3, and the ASP.NET Core Module.
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+        // Compare cbt against the token the client presented during authentication.
+    }
+
+    await next(context);
+});
+```
+
+Kestrel returns the binding from `SslStream.TransportContext.GetChannelBinding`. IIS and HTTP.sys return it from the request; on HTTP.sys the new `HttpSysOptions.Authentication.HardeningLevel` (defaults to `Medium`) controls whether the OS is asked to enforce channel binding, and setting it to `Strict` now fails startup if the OS cannot apply that hardening rather than silently degrading ([dotnet/aspnetcore #67720](https://github.com/dotnet/aspnetcore/pull/67720)).
+
+## Kestrel and IIS hardening
```
danroth27: These are all just bug fixes that should just be included in the "Bug fixes" section.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+
+**Zstandard request decompression window cap** — the request decompression middleware caps the zstd decoder window at 8 MB (RFC 9659's ceiling) so a small compressed payload cannot force a 128 MB decoder allocation. Well-behaved clients that emit a ≤ 2 MB window are unaffected ([dotnet/aspnetcore #67688](https://github.com/dotnet/aspnetcore/pull/67688)).
+
+**CONNECT close-on-reject** — HTTP/1.1 `CONNECT` requests that Kestrel rejects with a 3xx-or-higher status now close the connection instead of keeping it alive, per [RFC 9931 §8](https://www.rfc-editor.org/rfc/rfc9931#section-8) ([dotnet/aspnetcore #67929](https://github.com/dotnet/aspnetcore/pull/67929)). HTTP/2, HTTP/3, and accepted (2xx) CONNECTs are unaffected.
+
+**Multipart headers limit enforced across buffer boundaries** — the `MultipartHeadersLengthLimit` enforcement now spans buffered reads so a peer can no longer split a huge header across chunks to slip past the limit ([dotnet/aspnetcore #67840](https://github.com/dotnet/aspnetcore/pull/67840)).
+
+## Rewrite middleware collapses scheme-relative slashes
```
danroth27: This is just a bug fix that should be included in the Bug fixes section.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+
+Four of these were community contributions — see the [community section](#community-contributors) at the end.
+
+Because all five default to `Warning`, upgrading an existing app from Preview 6 will surface new warnings in code that previously built clean. Two cases are worth calling out:
+
+- `BL0012` fires on the common pattern of calling `StateHasChanged()` at the top of an `async` event handler before an `await`. The call really is redundant — `ComponentBase` re-renders as soon as the handler yields — so the code fix is safe. If the handler relies on the UI updating before a long synchronous block, keep the `await Task.Yield()`; that, not `StateHasChanged()`, is what lets the render reach the browser.
+- `BL0016` also fires on reusable wrapper and library code that deliberately lets `JSException` / `JSDisconnectedException` propagate to its caller. Wrapping those calls in a `try`/`catch` that swallows the exception would be wrong; suppress the diagnostic with a justification instead.
+
+## Blazor `Virtualize` API renames and improvements
```
danroth27: Should the API changes in this section just be covered in the Breaking changes section? The improvement that uses the equality comparer seems like it could stay.

```suggestion
## Blazor `Virtualize` no longer requires an ItemComparer
```

danroth27: If these API changes are too long to fit in the Breaking changes section, I guess they can stay as is. But consider moving them for consistency.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+@* Preview 6 *@
+<Toast Message="@("Saved 3 items.")" />
+
+@* Preview 7 *@
+<Toast Message="Saved 3 items." />
+```
+
+One gap remains: child content between the component's tags still doesn't populate a `RenderFragment` case, so that case has to be supplied as an expression ([dotnet/razor #13200](https://github.com/dotnet/razor/issues/13200)).
+
+## Blazor SSR client-side form validation reworked
```
danroth27: More preview-to-preview API changes. Consider if these should just be moved to the Breaking changes section.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,488 @@
...
+
+Preview 5 added client-side validation for Blazor static SSR forms: adding `<DataAnnotationsValidator />` inside a static `EditForm` now runs `[Required]`, `[Range]`, `[RegularExpression]`, `[StringLength]`, and similar rules in the browser before the form posts. Preview 7 keeps that feature but reworks the implementation and the API shape after review feedback ([dotnet/aspnetcore #67324](https://github.com/dotnet/aspnetcore/pull/67324), [dotnet/aspnetcore #67855](https://github.com/dotnet/aspnetcore/pull/67855)):
+
+- Rules are now delivered through a single inert per-form carrier element instead of MVC-style `data-val-*` attributes on every input.
+- `DataAnnotationsValidator.EnableClientValidation` was replaced by `DisableClientValidation` (default `false`, so client validation is on out of the box). A global `RazorComponentsServiceOptions.DisableClientValidation` opt-out turns rule emission off for every form.
```
danroth27: Switching the default to having client-side validation enabled seems worth keeping called out as a feature/improvement.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Applied. The heading and TOC entry are now "Auto pause Blazor circuits on inactivity."

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Applied as suggested.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Applied, with "package" added after the package ID so the sentence reads grammatically.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Applied as suggested.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Applied as suggested.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Addressed across the file. The em-dash count is now zero; the Breaking changes bullets use a bold lead sentence instead, and the remaining prose uses commas, semicolons, or separate sentences.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Done. The table now has "What it detects" and "Contribution" columns, so each analyzer's explanation, contributor, and PR are together. The two upgrade-specific caveats remain below the table.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Done. Each analyzer row now credits its contributor directly: @MayaKirova, @kdinev, @skrustev, and @damyanpetev.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Removed.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Applied as suggested. I also moved the preview-to-preview API renames from this section into Breaking changes, while keeping the implementation change and enabled-by-default behavior in the feature section.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Applied as suggested.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Applied as suggested.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Corrected. The section is now scoped to the SignalR .NET client change, and explicitly says Azure SignalR Service has not enabled the feature yet. It no longer announces a service update.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Tightened to one implementation sentence: the client preserves the app-token provider, adopts the refreshed transport token, and retains tokenLifetimeSeconds across the redirect.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Applied as suggested.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Tightened. The guidance is now four short sentences covering the direct-return JSON behavior, the SseItem overload, and the eventType alternative.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Superseded by the structural change requested in the adjacent thread: the whole section moved into Bug fixes, so it no longer characterizes these as security fixes.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Done. All six PRs moved into a "Server and middleware hardening" group in Bug fixes, one PR link per bullet.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Done. Both Rewrite middleware PRs moved into the "Server and middleware hardening" bug-fix group, one PR link per bullet.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Done. The Preview 6 API renames moved into Breaking changes. The improvement remains as a short feature section under "Blazor Virtualize no longer requires an ItemComparer." This also addresses the follow-up about consistency; the rename set fits cleanly in one Breaking changes bullet.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Done. The preview-to-preview client-validation API changes moved into Breaking changes.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Kept and emphasized. The section is now titled "Blazor SSR client-side form validation is enabled by default," and the body retains the per-form and app-wide opt-out guidance.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,441 @@
...
+
+`CacheView` supports absolute (`ExpiresAfter`, `ExpiresOn`) and sliding (`ExpiresSliding`) expiration, and vary-by dimensions for query string, route, header, cookie, authenticated user (`VaryByUser`), current culture (`VaryByCulture`), and an arbitrary `VaryBy` string. `CacheKey` disambiguates multiple boundaries under the same parent. Caching is skipped for non-GET requests, when `Enabled="false"`, and inside streaming SSR.
+
+By default `CacheView` uses an in-memory store bounded by `RazorComponentsServiceOptions.CacheViewSizeLimit` (default 100 MB). If the app registers `HybridCache`, `CacheView` picks it up from DI automatically, which gives a two-tier local/distributed cache with no other code change:
+
+```csharp
+builder.Services.AddHybridCache();
+
+builder.Services.AddRazorComponents()
+    .AddInteractiveServerRenderMode();
```
kotlarmilos: Should this be AddInteractiveServerComponents instead?

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,441 @@
...
+- **`UseWebAssemblyDebugging` obsolete, DevServer package deprecated.** `WebAssemblyNetDebugProxyAppBuilderExtensions.UseWebAssemblyDebugging` is now `[Obsolete("ASPDEPR011")]`, and the `Microsoft.AspNetCore.Components.WebAssembly.DevServer` NuGet package is marked deprecated in favor of the new Blazor Gateway host, which ships as the `Microsoft.AspNetCore.Components.Gateway` package and a `Microsoft.AspNetCore.Components.Gateway.Cli` .NET tool ([dotnet/aspnetcore #67990](https://github.com/dotnet/aspnetcore/pull/67990)). The Gateway serves a standalone WebAssembly app with HTTPS redirection, HSTS, health-check endpoints, and telemetry configurable through a `Gateway` configuration section. The Blazor Web App templates no longer call `UseWebAssemblyDebugging`; Visual Studio, VS Code, and `dotnet run` attach the WebAssembly debugger automatically. Remove `app.UseWebAssemblyDebugging()` from your `Program.cs` if you added it manually ([dotnet/aspnetcore #67861](https://github.com/dotnet/aspnetcore/pull/67861), [dotnet/aspnetcore #67862](https://github.com/dotnet/aspnetcore/pull/67862)).
```
kotlarmilos: dotnet run doesn't attach debugger
```suggestion
- **`UseWebAssemblyDebugging` obsolete, DevServer package deprecated.** `WebAssemblyNetDebugProxyAppBuilderExtensions.UseWebAssemblyDebugging` is now `[Obsolete("ASPDEPR011")]`, and the `Microsoft.AspNetCore.Components.WebAssembly.DevServer` NuGet package is marked deprecated in favor of the new Blazor Gateway host, which ships as the `Microsoft.AspNetCore.Components.Gateway` package and a `Microsoft.AspNetCore.Components.Gateway.Cli` .NET tool ([dotnet/aspnetcore #67990](https://github.com/dotnet/aspnetcore/pull/67990)). The Gateway serves a standalone WebAssembly app with HTTPS redirection, HSTS, health-check endpoints, and telemetry configurable through a `Gateway` configuration section. The Blazor Web App templates no longer call `UseWebAssemblyDebugging`; Visual Studio and VS Code attach the WebAssembly debugger automatically. Remove `app.UseWebAssemblyDebugging()` from your `Program.cs` if you added it manually ([dotnet/aspnetcore #67861](https://github.com/dotnet/aspnetcore/pull/67861), [dotnet/aspnetcore #67862](https://github.com/dotnet/aspnetcore/pull/67862)).
```

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Yes. Corrected to AddInteractiveServerComponents. AddInteractiveServerRenderMode remains on the endpoint mapping earlier in the file; this snippet is service registration and now matches the working Preview 7 sample.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Applied as suggested. Visual Studio and VS Code attach automatically; dotnet run does not.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+
+## `wasm-tools` uses Emscripten 6
+
+The .NET 11 `wasm-tools` workload now uses Emscripten 6.0.3, upgraded from Emscripten 5.0.6 in Preview 6 ([dotnet/emsdk #1789](https://github.com/dotnet/emsdk/pull/1789), [dotnet/emsdk #1788](https://github.com/dotnet/emsdk/pull/1788)). Blazor developers don't typically interact with Emscripten directly, but it provides the compiler toolchain used for WebAssembly AOT compilation and native dependencies.
+
+Emscripten 6 updates the underlying native toolchain and libraries, including musl libc, libpng, compiler-rt, and libunwind. Generated output can rely on WebAssembly mutable globals and sign-extension instructions without a Babel transpilation step. It also adds capabilities for native code and toolchain authors, including Memory64 through `-m64`, FMA intrinsics, wasm-bindgen integration, and real dynamic libraries by default. Streaming Fetch operations now cap chunks at 8 MB so large downloads don't arrive as a single allocation in WebAssembly memory.
```
akoeplinger: most of that isn't usable by blazor so I'd make this a bit more generic
```suggestion
Emscripten 6 updates the underlying native toolchain and libraries, including musl libc, llvm/clang, and libunwind. These updates bring in multiple years of optimizations across a wide range of areas from these upstream projects.
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+- **Removed long-obsolete MVC APIs.** `CompatibilityVersion`, `IMvcBuilder.SetCompatibilityVersion` / `IMvcCoreBuilder.SetCompatibilityVersion`, `MvcCompatibilityOptions`, `ConfigureCompatibilityOptions<TOptions>`, the `ImageTagHelper` constructor that took `IWebHostEnvironment` and `TagHelperMemoryCacheProvider`, `ImageTagHelper.HostingEnvironment`, `ImageTagHelper.Cache`, and the `ModelMetadataIdentity.ForProperty(Type, string, Type)` overload have been removed after multiple releases of obsolete-with-warning. Use the `PropertyInfo`-based `ModelMetadataIdentity.ForProperty` overload; delete `SetCompatibilityVersion` calls (the "current" behavior is the only supported one) ([dotnet/aspnetcore #67077](https://github.com/dotnet/aspnetcore/pull/67077)).
```
Youssef1313: Regarding compatibility switches that was removed, I **think** the API previously had no effect.

I tried to look through the history, but I don't have full context on everything that happened, but looks like those compatibility switches were cleaned up as part of https://github.com/dotnet/aspnetcore/issues/4866.

danroth27: @Youssef1313 Is any edit to the release notes needed here then? Do you want me to remove mention of the compat switch removal?

Youssef1313: > (the "current" behavior is the only supported one)

I guess dropping this part is clearer. Because even with the obsolete APIs, they were not restoring older behaviors. They did at some point in the past, but on the more recent versions they didn't (probably since the APIs were marked obsolete they started doing nothing).

danroth27:
```suggestion
- **Removed long-obsolete MVC APIs.** `CompatibilityVersion`, `IMvcBuilder.SetCompatibilityVersion` / `IMvcCoreBuilder.SetCompatibilityVersion`, `MvcCompatibilityOptions`, `ConfigureCompatibilityOptions<TOptions>`, the `ImageTagHelper` constructor that took `IWebHostEnvironment` and `TagHelperMemoryCacheProvider`, `ImageTagHelper.HostingEnvironment`, `ImageTagHelper.Cache`, and the `ModelMetadataIdentity.ForProperty(Type, string, Type)` overload have been removed after multiple releases of obsolete-with-warning. Use the `PropertyInfo`-based `ModelMetadataIdentity.ForProperty` overload; delete `SetCompatibilityVersion` calls ([dotnet/aspnetcore #67077](https://github.com/dotnet/aspnetcore/pull/67077)).
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+}
+```
+
+The same localization rules apply to validation for minimal APIs and Blazor, so a message localizes identically wherever the model is used.
+
+`Microsoft.Extensions.Validation` also drops its `[Experimental("ASP0029")]` markers in Preview 7 ([dotnet/aspnetcore #67634](https://github.com/dotnet/aspnetcore/pull/67634)). If you suppressed `ASP0029` to use `[ValidatableType]` or `AddValidation()`, you can remove the suppression.
+
+## SignalR .NET client supports auth refresh after redirects
+
+Preview 7 updates the SignalR .NET client so [authentication refresh](../preview6/aspnetcore.md#signalr-authentication-refresh) works when negotiate redirects to another server ([dotnet/aspnetcore #67612](https://github.com/dotnet/aspnetcore/pull/67612), contributed by [@MoChilia](https://github.com/MoChilia)). This client change enables support for redirecting servers such as Azure SignalR Service, but Azure SignalR Service has not enabled the feature yet.
```
BrennanConroy, line 247: >contributed by [@MoChilia](https://github.com/MoChilia))

This is a dev from Azure SignalR Service, not sure we count that as a community member

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+- [Bug fixes](#bug-fixes)
+- [Community contributors](#community-contributors)
+
+ASP.NET Core updates in .NET 11:
+
+- [What's new in ASP.NET Core in .NET 11](https://learn.microsoft.com/aspnet/core/release-notes/aspnetcore-11)
+
+## Auto pause Blazor circuits on inactivity
+
+Interactive Server circuits can now pause themselves automatically when the browser tab is hidden, releasing server resources until the user comes back ([dotnet/aspnetcore #67098](https://github.com/dotnet/aspnetcore/pull/67098)). Preview 6 added [`Circuit.RequestCircuitPauseAsync`](https://github.com/dotnet/aspnetcore/issues/64886) so the server could pause a circuit on demand; Preview 7 adds a matching client-driven path that fires after a configurable inactivity delay while a page is hidden.
```
ilonatommy: > Preview 7 adds a matching client-driven path 

I don't know if I would call it this way, at first I understood we added a client methods for pausing that are symmetric to `Circuit.RequestCircuitPauseAsync` which would be https://github.com/dotnet/aspnetcore/blob/a728d9836edf4f4d73d77c8b4355e4b9d57a94cf/src/Components/Web.JS/src/Platform/Circuits/CircuitManager.ts#L367. But these methods was already in the framework for some time. Shortly: it made me confused.

My proposal:
```
Preview 6 added [`Circuit.RequestCircuitPauseAsync`](https://github.com/dotnet/aspnetcore/issues/64886) so the server could pause a circuit on demand. Preview 7 added a package that can help auto-detect the moment when `Circuit.RequestCircuitPauseAsync` should be requested. The detection runs on hidden pages and provides two points of configuration: timeout and client callbacks that should run before the pause happens.
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+    {
+        options.AddAutoPause(pause =>
+        {
+            pause.Enabled = true; // default
+            pause.HiddenDelay = TimeSpan.FromSeconds(30); // default is 2 minutes
+        });
+    });
+```
+
+Auto-pause defers itself when a pause would lose data or interrupt work in progress. The shipped checks are a text input or `contenteditable` element whose value differs from its default (including inside shadow DOM and same-origin iframes), an `<audio>` or `<video>` element that is playing unmuted, an open [Picture-in-Picture](https://developer.mozilla.org/docs/Web/API/Picture-in-Picture_API) window, a held [Web Lock](https://developer.mozilla.org/docs/Web/API/Web_Locks_API), and any in-flight circuit activity such as a running `IJSRuntime` call, a `DotNetStreamReference`/`JSStreamReference` transfer, or a queued render.
```
ilonatommy: > Auto-pause defers itself when a pause would lose data or interrupt work in progress.

True but I had a hard time to understand. 

```
Auto-pause is deferred by default in some situations when a pause could cause data loss or interrupt work in progress.
```
Feel free to find a middle ground if any of my suggestions doesn't look like an improvement.

> The shipped checks are a text input or `contenteditable` element whose value differs from its default (including inside shadow DOM and same-origin iframes)...

```
These situations are e.g. a text input or `contenteditable` element whose value differs from its default (including inside shadow DOM and same-origin iframes)...
```

ilonatommy (commented):
Auto-pause has a few comments, Virtualize and QuckGrid approved.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+    <DataAnnotationsValidator />
+
+    <InputText @bind-Value="Model.Email" />
+    <ValidationMessage For="() => Model.Email" />
+
+    <button type="submit">Book</button>
+</EditForm>
+```
+
+## Validation localization is built in
```
Youssef1313, line 199: @oroztocil to review this

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+- **`Microsoft.AspNetCore.Grpc.Swagger` package removed.** The OpenAPI-for-gRPC-transcoding package is deleted. Use ASP.NET Core's built-in OpenAPI document generation on the transcoded HTTP endpoints instead ([dotnet/aspnetcore #67919](https://github.com/dotnet/aspnetcore/pull/67919)).
```
Youssef1313: @JamesNK to confirm if this is viable.

Youssef1313 (approved):

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+    public int Divisor { get; init; }
+
+    public string FormatMessage(CultureInfo culture, string template, string displayName)
+        => string.Format(culture, template, displayName, Divisor); // {0} = name, {1} = divisor
+}
+```
+
+The same localization rules apply to validation for minimal APIs and Blazor, so a message localizes identically wherever the model is used.
+
+`Microsoft.Extensions.Validation` also drops its `[Experimental("ASP0029")]` markers in Preview 7 ([dotnet/aspnetcore #67634](https://github.com/dotnet/aspnetcore/pull/67634)). If you suppressed `ASP0029` to use `[ValidatableType]` or `AddValidation()`, you can remove the suppression.
```
oroztocil: This should not be a side-note under Validation localization. I propose adding a dedicated section:

## `Microsoft.Extensions.Validation` is no longer experimental

`Microsoft.Extensions.Validation` drops its `[Experimental("ASP0029")]` markers in Preview 7 ([dotnet/aspnetcore #67634](https://github.com/dotnet/aspnetcore/pull/67634)). Applications that suppressed `ASP0029` to use `[ValidatableType]`, `AddValidation()`, or `ValidationOptions`, can remove the suppression.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+
+```razor
+@* Preview 6 *@
+<Toast Message="@("Saved 3 items.")" />
+
+@* Preview 7 *@
+<Toast Message="Saved 3 items." />
+```
+
+## Blazor SSR client-side form validation is enabled by default
```
oroztocil: Nothing changed about client-side form validation being enabled by default. We just switched the flag controlling it from `EnableClientValidation { get; set; } = true;` to `DisableClientValidation { get; set; } = false;`.

The global setting `RazorComponentsServiceOptions.DisableClientValidation` is new, that is correct.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+
+Preview 5 added client-side validation for Blazor static SSR forms without requiring an interactive render mode. Preview 7 keeps that feature but reworks the implementation and the API shape after review feedback ([dotnet/aspnetcore #67324](https://github.com/dotnet/aspnetcore/pull/67324), [dotnet/aspnetcore #67855](https://github.com/dotnet/aspnetcore/pull/67855)):
+
+- Rules are now delivered through a single inert per-form carrier element instead of MVC-style `data-val-*` attributes on every input.
+- Client-side validation is enabled by default. Set `DataAnnotationsValidator.DisableClientValidation="true"` for one form, or set `RazorComponentsServiceOptions.DisableClientValidation` to disable rule emission for the app.
+
+```razor
```
oroztocil: The code example here does not add much value. It is just plain Blazor form as you would see in a .NET 10 app. At least show the use of the opt-out flag - or probably just remove the code example.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+<Toast Message="Saved 3 items." />
+```
+
+## Blazor SSR client-side form validation is enabled by default
+
+Preview 5 added client-side validation for Blazor static SSR forms without requiring an interactive render mode. Preview 7 keeps that feature but reworks the implementation and the API shape after review feedback ([dotnet/aspnetcore #67324](https://github.com/dotnet/aspnetcore/pull/67324), [dotnet/aspnetcore #67855](https://github.com/dotnet/aspnetcore/pull/67855)):
+
+- Rules are now delivered through a single inert per-form carrier element instead of MVC-style `data-val-*` attributes on every input.
+- Client-side validation is enabled by default. Set `DataAnnotationsValidator.DisableClientValidation="true"` for one form, or set `RazorComponentsServiceOptions.DisableClientValidation` to disable rule emission for the app.
+
```
oroztocil: New behavioral aspect that can be mentioned here:


```suggestion
- The client-side engine now applies the same CSS classes as Blazor's interactive validation: `valid` and `invalid` on inputs, with `modified` once the user edits a field, `validation-message` on message elements, and `validation-summary-errors` or `validation-summary-valid` on the summary.
- Client-side rules are emitted only for fields the server would also validate. Validation on the server is authoritative and client-side validation is a UX improvement, so a field that the server ignores (for example a nested property in an app that has not called `AddValidation()`) no longer gets a client-side rule that suggests otherwise.
```

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,450 @@
...
+    <InputText @bind-Value="Model.Email" />
+    <ValidationMessage For="() => Model.Email" />
+
+    <button type="submit">Book</button>
+</EditForm>
+```
+
+## Validation localization is built in
+
+`Microsoft.Extensions.Validation` now localizes validation messages and display names without a separate package. The preview-only `Microsoft.Extensions.Validation.Localization` package and its `AddValidationLocalization<TResource>()` / `IValidationLocalizer` API are removed; localization now activates automatically as soon as an `IStringLocalizerFactory` is registered, and the lookup is emitted by the validation source generator into your assembly ([dotnet/aspnetcore #68005](https://github.com/dotnet/aspnetcore/pull/68005)).
```
oroztocil: The correct PR is #67987

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: If they're not directly on the .NET team, we still want to acknowledge and thank them for their contributions.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: I assume you're referring to this statement: "Use ASP.NET Core's built-in OpenAPI document generation on the transcoded HTTP endpoints instead". That may be copilot making an incorrect assumption that this is possible. I'll double check. The text may just need to be removed.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Ah, makes sense. So, this is just an API change, not a behavior change. We should probably just move this then to the breaking changes list.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Updated the wording to describe the Preview 7 package as detecting when to request a pause after a configurable inactivity delay, instead of calling it a new client-driven path.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Updated the wording to clarify that auto-pause is deferred by default when pausing could cause data loss or interrupt work, and introduced the checks as examples of those situations.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Confirmed from dotnet/aspnetcore#67919 that built-in OpenAPI document generation is not a replacement for this package. Removed the unsupported replacement guidance.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Moved this into a dedicated section. I scoped the wording to ValidatableTypeAttribute and SkipValidationAttribute because #67634 stabilizes those two attributes; AddValidation and ValidationOptions were already stable.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Removed the claim that client-side validation became enabled by default. The API polarity and related API changes remain in Breaking changes, while the feature section now covers only the actual Preview 7 behavioral improvements.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Removed the generic form example.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Added both behavioral improvements, including the complete interactive-style CSS classes and the server-authoritative rule emission behavior with the nested-property example.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Updated the localization citation to the implementation PR, #67987.

release-notes/11.0/preview/preview7/aspnetcore.md
Youssef1313: Thanks. Yeah I was exactly referring to this part as I thought there is no replacement but I wasn't 100% sure. Thanks for confirming!

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,442 @@
...
+            out ReadOnlyMemory<byte> cbt))
+    {
+        // Compare cbt against the token the client presented during authentication.
+    }
+
+    await next(context);
+});
+```
+
+Kestrel returns the binding from `SslStream.TransportContext.GetChannelBinding`. IIS and HTTP.sys return it from the request; on HTTP.sys the new `HttpSysOptions.Authentication.HardeningLevel` (defaults to `Medium`) controls whether the OS is asked to enforce channel binding, and setting it to `Strict` now fails startup if the OS cannot apply that hardening rather than silently degrading ([dotnet/aspnetcore #67720](https://github.com/dotnet/aspnetcore/pull/67720)).
```
halter73:
```suggestion
Kestrel returns the binding from `SslStream.TransportContext.GetChannelBinding`. IIS and HTTP.sys return it from the request; on HTTP.sys the new `HttpSysOptions.HttpAuthenticationHardeningLevel` (defaults to `Medium`) controls whether the OS is asked to enforce channel binding, and setting it to `Strict` fails startup if the OS cannot apply that hardening rather than silently degrading like the default `Medium` hardening level does ([dotnet/aspnetcore #67720](https://github.com/dotnet/aspnetcore/pull/67720)).
```

It's `HttpSysOptions.HttpAuthenticationHardeningLevel` not `HttpSysOptions.Authentication.HardeningLevel`.

"now" seems weird to add since `HardeningLevel.Strict` is completely new. We're not chasing the behavior. The contrast is from the default `Medium` level, not from some previous `Strict` behavior we never had.

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,442 @@
...
+              $ref: '#/components/schemas/Todo'
+            event: { type: string }
+            id: { type: string }
+```
+
+If the event payload is a discriminated union (a preview C# 14 feature), OpenAPI also emits the union's case names as an `enum` on the `event` field.
+
+## TLS channel-binding token access from `ITlsConnectionFeature`
+
+Applications that authenticate with Kerberos, NTLM, or another SSPI-based scheme over TLS can now read the connection's channel binding token to defend against relay attacks ([dotnet/aspnetcore #67436](https://github.com/dotnet/aspnetcore/pull/67436)):
```
halter73:
```suggestion
Applications using TLS can now read the connection's channel binding token to defend against relay attacks ([dotnet/aspnetcore #67436](https://github.com/dotnet/aspnetcore/pull/67436)):
```

You should be able to get the token no matter if/how your application handles user authentication. That should have no bearing on whether `TryGetChannelBindingBytes` works or not.

It's particularly useful for Windows auth, and IIS/HTTP.Sys should leverage it automatically when doing Windows auth. However, I don't think we've updated the `NegotiateHandler` to leverage this when running on Kestrel yet. @DeagleGross do we have an issue tracking this?

halter73 (approved):

dariatiurina (approved):

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,442 @@
...
+- [@kdinev](https://github.com/dotnet/aspnetcore/pulls?q=is%3Apr+is%3Amerged+author%3Akdinev+milestone%3A11.0-preview7)
+- [@KitKeen](https://github.com/dotnet/aspnetcore/pulls?q=is%3Apr+is%3Amerged+author%3AKitKeen+milestone%3A11.0-preview7)
+- [@kobihikri](https://github.com/dotnet/aspnetcore/pulls?q=is%3Apr+is%3Amerged+author%3Akobihikri+milestone%3A11.0-preview7)
+- [@mahdiaghtaee](https://github.com/dotnet/aspnetcore/pulls?q=is%3Apr+is%3Amerged+author%3Amahdiaghtaee+milestone%3A11.0-preview7)
+- [@marcominerva](https://github.com/dotnet/aspnetcore/pulls?q=is%3Apr+is%3Amerged+author%3Amarcominerva+milestone%3A11.0-preview7)
+- [@martincostello](https://github.com/dotnet/aspnetcore/pulls?q=is%3Apr+is%3Amerged+author%3Amartincostello+milestone%3A11.0-preview7)
+- [@MayaKirova](https://github.com/dotnet/aspnetcore/pulls?q=is%3Apr+is%3Amerged+author%3AMayaKirova+milestone%3A11.0-preview7)
+- [@medhatiwari](https://github.com/dotnet/aspnetcore/pulls?q=is%3Apr+is%3Amerged+author%3Amedhatiwari+milestone%3A11.0-preview7)
+- [@oroztocil](https://github.com/dotnet/aspnetcore/pulls?q=is%3Apr+is%3Amerged+author%3Aoroztocil+milestone%3A11.0-preview7)
```
oroztocil: Not a community contributor (unless I missed some news)

release-notes/11.0/preview/preview7/aspnetcore.md
```diff
@@ -0,0 +1,442 @@
...
+
+```csharp
+builder.Services.AddValidation(options =>
+{
+    // Resolve every model's keys from one shared resource file.
+    options.LocalizerProvider = (_, factory) => factory.Create(typeof(ValidationMessages));
+
+    // Supply a key by convention when an attribute has no ErrorMessage, for example
+    // [Range] => "RangeAttribute_Error".
+    options.MessageKeyProvider = ctx => $"{ctx.ValidatorType.Name}_Error";
```
oroztocil: Should we publicize this API if we decided during API review to drop it in favor of non-configurable built-in conventions? RC1 will ship the final shape.

kotlarmilos (approved):

release-notes/11.0/preview/preview7/aspnetcore.md
DeagleGross: Should we be more specific explaining options?

```suggestion
Kestrel returns the binding from `SslStream.TransportContext.GetChannelBinding`. IIS and HTTP.sys return it from the request; On HTTP.sys, `HttpSysOptions.HttpAuthenticationHardeningLevel` controls Extended Protection and CBT exposure:
- `Legacy` disables CBT validation and exposure;
- `Medium` (the default) exposes the CBT and validates it when supplied but tolerates its absence;
- `Strict` requires CBT for authenticated requests and fails startup if the OS cannot apply the configuration
```

release-notes/11.0/preview/preview7/aspnetcore.md
DeagleGross: We did not; thanks for reminding about it! Here is the issue: https://github.com/dotnet/aspnetcore/issues/68315

DeagleGross (approved):

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27:
```suggestion
```

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Combined both suggestions. The text now uses the correct `HttpSysOptions.HttpAuthenticationHardeningLevel` property and explains the observable semantics of `Legacy`, `Medium`, and `Strict`. I also kept the startup distinction from #67720 precise: `Strict` fails startup if the OS configuration cannot be applied, while `Legacy` and `Medium` log that configuration failure and continue. This avoids implying that `Strict` previously degraded, while still explaining how it differs from the default `Medium` behavior.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Removed the authentication-scheme restriction from the introduction. `TryGetChannelBindingBytes` is available independently of how the application handles authentication. The section still explains the relay-attack use case without implying that only SSPI-based applications can retrieve the token.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Removed @oroztocil from Community contributors because this was an internal contribution.

release-notes/11.0/preview/preview7/aspnetcore.md
danroth27: Agreed. Removed the `ValidationOptions.MessageKeyProvider` sample and related prose. API review approved removing that temporary Preview 7 customization in favor of built-in key fallback conventions for the final shape. The section retains `LocalizerProvider` and `IValidationMessageFormatter`, which remain in the approved API.

oroztocil (approved):

PR #10516, opened 2026-08-02
Encode .NET 11 Preview 7 retrospective into the release-notes skills

danroth27:
Folds the lessons from generating and validating the .NET 11 Preview 7 release notes back into the release-notes skills, so future milestones apply the same provenance, runtime-verification, editorial, and PR-management checks.

## What prompted this

Four documented claims in the Preview 7 ASP.NET Core notes were wrong, and none were catchable by static managed API verification alone:

| Claim | Reality | Why static API verification missed it |
| --- | --- | --- |
| `Blazor.pause.waitFor(fn)` | Does not exist | JavaScript surface area is invisible to a managed API check |
| `.RequireAntiforgeryToken()` | Does not exist | There was no symbol to verify |
| `EnableClientValidation` | The shipped API is `DisableClientValidation`, with inverted polarity | A name-only check doesn't verify defaults or meaning |
| A documented sample | Compiled but returned HTTP 500 | `[PersistentState]` silently required a public property |

The errors were found by building and running maintained samples against the actual milestone build.

### Runtime verification

- Adds the new `validate-code-samples` skill, filling the runtime-verification stage that `api-diff-validation` already referenced.
- Documents selecting the milestone's matching SDK feature-band or preview build through the latest-builds table linked from `dotnet/sdk`, confirming VMR provenance with `productCommit-*.json` and `build-metadata.json`, and installing it in a scoped location instead of relying on a machine-wide preview SDK.
- Requires compiling every sample and exercising documented defaults, polarity, JavaScript/browser APIs, endpoints, runtime behavior, and public feature reachability.
- Adds guidance for recording evidence and deciding whether to fix the notes, update a stale sample, or drop an unsupported claim.
- Wires `validate-code-samples` into the main release-notes workflow and fixes the link from `api-diff-validation`.

### Shipped-change and milestone provenance

- Adds a milestone cross-check to `generate-changes` for repositories that maintain per-preview milestones: `dotnet/sdk`, `dotnet/aspnetcore`, `dotnet/runtime`, and `dotnet/efcore`.
- Keeps `changes.json` as the authoritative shipped-change manifest. Milestones are strong but not conclusive supporting evidence: they are normally applied by automation, but assignments can be changed, incorrect, stale, or followed by a revert.
- Requires independent verification before adding anything discovered through a milestone sweep and distinguishes repositories that don't maintain per-preview milestones from repositories with no shipped changes.
- Adds the corresponding provenance rule to API verification: proving that an API exists in the current build does not prove that it changed in this milestone. Features discovered while testing must be traced back to the current milestone before being documented.
- Calls out the limits of static verification for JavaScript APIs, defaults and polarity, and runtime-only behavior.

### Editorial structure

- Defines the Breaking changes section in preview notes as the upgrade guidance from the previous preview, covering renamed APIs, new analyzer diagnostics, changed defaults, and other changes that would otherwise cause build errors, warnings, or silent behavior changes.
- Keeps removed workarounds with the corresponding bug fix rather than misclassifying them as breaking changes.
- Updates the Bug fixes format to group fixes by namespace or feature area, exactly one level deep.
- Requires one PR-linked bullet per fix, using a cleaned-up PR or issue title as the link text, with no bundled links or trailing citation/prose. Test-only, CI, and infrastructure fixes remain excluded.

### PR creation, ownership, and merge flow

- Documents pushing the milestone base branch before opening component PRs and checking that each component branch is based directly on the base commit, avoiding missing base branches and noisy component diffs.
- Requires validating owner mappings through the repository assignees endpoint and re-reading each created PR because GitHub can silently drop non-assignable users.
- Documents that `gh pr edit` is unusable in `dotnet/core` because of the Projects (classic) GraphQL failure, including title/body edits as well as assignee changes, and provides REST alternatives.
- Makes component owners responsible for merging their approved PR into the milestone base branch and requires that responsibility to be stated when owners are notified.

## Notes

- Supersedes the `release-notes/pr-owner-merges` branch, whose commit is included here; that branch can be deleted.
- `.github/CODEOWNERS` separately lists `/release-notes/ @rbhanda @leecow @victorisr`. That is review ownership rather than per-component ownership, so it is deliberately unchanged.

### DIFF
--- a/.github/skills/generate-changes/SKILL.md
+++ b/.github/skills/generate-changes/SKILL.md
@@ -155,14 +155,14 @@ Rules for using it:
   legitimately appear in `changes.json`, so the counts will not match and are not meant to.
 - **Only the missing direction matters.** What is worth acting on is a PR in the milestone that
   describes a user-facing change and has no corresponding entry in the notes.
-- **Verify every addition independently before writing it up.** Milestone membership is a label
-  applied by hand, so it proves someone intended the PR for this release — not that the change
-  actually shipped in it. A PR can be milestoned and then reverted, retargeted to a later preview,
-  or milestoned in error. Before promoting anything found this way, confirm it appears in
-  `changes.json` and exists in the build (see
+- **Verify every addition independently before writing it up.** Milestones are applied by
+  automation, so they are usually right — but they can be changed or applied incorrectly by hand
+  afterwards, and a milestoned PR can still be reverted. Treat a milestone as reliable evidence of
+  where to look and weak evidence that the change shipped. Before promoting anything found this
+  way, confirm it appears in `changes.json` and exists in the build (see
   [`api-verification.md`](../release-notes/references/api-verification.md) and
-  [`validate-code-samples`](../validate-code-samples/SKILL.md)). The milestone tells you where to
-  look; it is not evidence on its own.
+  [`validate-code-samples`](../validate-code-samples/SKILL.md)). The expected outcome is that it
+  checks out; the point is to catch the occasional one that does not.
 - **Confirm the milestone exists before relying on its absence.** Repos without `11.x` milestones
   will return zero results, which means "not tracked here", not "nothing shipped".
 
--- a/.github/skills/generate-changes/SKILL.md
+++ b/.github/skills/generate-changes/SKILL.md
@@ -140,16 +140,12 @@ gh api -X GET search/issues \
   --jq '.total_count'
 ```
 
-Milestone discipline varies by repo, so this check only applies where it is actually maintained.
-Measured for 11.0-preview7:
+Milestone discipline varies by repo, so this check only applies for the repos where it is actually maintained:
 
-| Repo | Merged PRs in milestone | Entries in `changes.json` | Use it? |
-| ---- | ----------------------- | ------------------------- | ------- |
-| `dotnet/sdk` | 177 | 199 | Yes |
-| `dotnet/aspnetcore` | 176 | 205 | Yes |
-| `dotnet/runtime` | 583 | 692 | Yes |
-| `dotnet/efcore` | 77 | 107 | Yes, with weaker coverage |
-| `dotnet/roslyn`, `dotnet/razor` | no `11.x` milestones exist | 108 (roslyn) | No |
+- dotnet/sdk
+- dotnet/aspnetcore
+- dotnet/runtime
+- dotnet/efcore
 
 Rules for using it:
 
@@ -159,6 +155,14 @@ Rules for using it:
   legitimately appear in `changes.json`, so the counts will not match and are not meant to.
 - **Only the missing direction matters.** What is worth acting on is a PR in the milestone that
   describes a user-facing change and has no corresponding entry in the notes.
+- **Verify every addition independently before writing it up.** Milestone membership is a label
+  applied by hand, so it proves someone intended the PR for this release — not that the change
+  actually shipped in it. A PR can be milestoned and then reverted, retargeted to a later preview,
+  or milestoned in error. Before promoting anything found this way, confirm it appears in
+  `changes.json` and exists in the build (see
+  [`api-verification.md`](../release-notes/references/api-verification.md) and
+  [`validate-code-samples`](../validate-code-samples/SKILL.md)). The milestone tells you where to
+  look; it is not evidence on its own.
 - **Confirm the milestone exists before relying on its absence.** Repos without `11.x` milestones
   will return zero results, which means "not tracked here", not "nothing shipped".
 
--- a/.github/skills/generate-changes/SKILL.md
+++ b/.github/skills/generate-changes/SKILL.md
@@ -158,8 +158,8 @@ Rules for using it:
 - **Verify every addition independently before writing it up.** Milestones are applied by
   automation, so they are usually right — but they can be changed or applied incorrectly by hand
   afterwards, and a milestoned PR can still be reverted. Treat a milestone as reliable evidence of
-  where to look and weak evidence that the change shipped. Before promoting anything found this
-  way, confirm it appears in `changes.json` and exists in the build (see
+  where to look and strong but not conclusive evidence that the change shipped. Before promoting
+  anything found this way, confirm it appears in `changes.json` and exists in the build (see
   [`api-verification.md`](../release-notes/references/api-verification.md) and
   [`validate-code-samples`](../validate-code-samples/SKILL.md)). The expected outcome is that it
   checks out; the point is to catch the occasional one that does not.
--- a/.github/skills/generate-changes/SKILL.md
+++ b/.github/skills/generate-changes/SKILL.md
@@ -126,4 +126,40 @@ The output file must follow the shared schema documented in [changes-schema.md](
 - stable `id` values in `repo@shortcommit` format
 - same authoritative source of truth used by later skills
 
+## Milestone cross-check
+
+`changes.json` is derived from a VMR source-manifest diff. That makes it authoritative for *what
+flowed into the build*, but it is a commit-shaped view, and a feature can be easy to overlook in it.
+For the repos that maintain preview milestones, sweep the milestone as a **second, independent view**
+of the same release and reconcile anything that looks like a user-facing feature but never made it
+into the notes.
+
+```bash
+gh api -X GET search/issues \
+  -f q="repo:dotnet/aspnetcore is:pr is:merged milestone:11.0-preview7" \
+  --jq '.total_count'
+```
+
+Milestone discipline varies by repo, so this check only applies where it is actually maintained.
+Measured for 11.0-preview7:
+
+| Repo | Merged PRs in milestone | Entries in `changes.json` | Use it? |
+| ---- | ----------------------- | ------------------------- | ------- |
+| `dotnet/sdk` | 177 | 199 | Yes |
+| `dotnet/aspnetcore` | 176 | 205 | Yes |
+| `dotnet/runtime` | 583 | 692 | Yes |
+| `dotnet/efcore` | 77 | 107 | Yes, with weaker coverage |
+| `dotnet/roslyn`, `dotnet/razor` | no `11.x` milestones exist | 108 (roslyn) | No |
+
+Rules for using it:
+
+- **Supplement, never replace.** `changes.json` stays the source of truth. The milestone is a
+  prompt to go back and look, not an alternative manifest.
+- **The milestone is a subset.** It excludes infrastructure and dependency-flow PRs that
+  legitimately appear in `changes.json`, so the counts will not match and are not meant to.
+- **Only the missing direction matters.** What is worth acting on is a PR in the milestone that
+  describes a user-facing change and has no corresponding entry in the notes.
+- **Confirm the milestone exists before relying on its absence.** Repos without `11.x` milestones
+  will return zero results, which means "not tracked here", not "nothing shipped".
+
 Once `changes.json` exists, the next step is usually `generate-features`.
--- a/.github/skills/release-notes/references/api-verification.md
+++ b/.github/skills/release-notes/references/api-verification.md
@@ -116,6 +116,32 @@ You do NOT need to verify:
 - General concepts (e.g., "Zstandard compression" as a concept vs `ZstandardCompressionProvider` as a type)
 - CLI flags (e.g., `dotnet test --artifacts-path`)
 
+## Verify the change shipped in *this* milestone
+
+API verification answers "does this exist in the build". It does not answer "did this change in this
+release" — and those come apart in a way that is easy to miss.
+
+A behavior observed in the Preview 7 build was traced to a fix that shipped in **Preview 3**. It was
+real, it was verifiable, and documenting it in the Preview 7 notes would have been wrong. Anything
+still present from an earlier release will verify perfectly against the current build.
+
+The rule: **every documented change must trace to a PR in this milestone**, not merely be observably
+true in the build.
+
+1. **`changes.json` membership is the primary test.** If a change has no entry, it did not flow into
+   this milestone. Confirm before writing it up.
+2. **Check the PR's merge date and milestone** where the repo maintains them —
+   `dotnet/aspnetcore`, `dotnet/runtime`, `dotnet/sdk`, and `dotnet/efcore` do;
+   `dotnet/roslyn` and `dotnet/razor` do not (see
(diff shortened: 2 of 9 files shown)


.github/skills/generate-changes/SKILL.md
```diff
@@ -126,4 +126,40 @@ The output file must follow the shared schema documented in [changes-schema.md](
...
+Milestone discipline varies by repo, so this check only applies where it is actually maintained.
+Measured for 11.0-preview7:
+
+| Repo | Merged PRs in milestone | Entries in `changes.json` | Use it? |
+| ---- | ----------------------- | ------------------------- | ------- |
+| `dotnet/sdk` | 177 | 199 | Yes |
+| `dotnet/aspnetcore` | 176 | 205 | Yes |
+| `dotnet/runtime` | 583 | 692 | Yes |
+| `dotnet/efcore` | 77 | 107 | Yes, with weaker coverage |
+| `dotnet/roslyn`, `dotnet/razor` | no `11.x` milestones exist | 108 (roslyn) | No |
```
danroth27:
```suggestion
Milestone discipline varies by repo, so this check only applies for the repos where it is actually maintained:

- dotnet/sdk
- dotnet/aspnetcore
- dotnet/runtime
- dotnet/efcore

File: .github/skills/generate-changes/SKILL.md [hunk 2]
```diff
@@ -126,4 +126,40 @@ The output file must follow the shared schema documented in [changes-schema.md](
...
+
+Rules for using it:
```
danroth27, line 150: Add a rule stating that additions from merged PRs in the corresponding milestone must be separately verified to be in the release. PRs may have been incorrectly added to the milestone.

.github/skills/release-notes/references/api-verification.md
```diff
@@ -116,6 +116,32 @@ You do NOT need to verify:
...
+## Verify the change shipped in *this* milestone
+
+API verification answers "does this exist in the build". It does not answer "did this change in this
+release" — and those come apart in a way that is easy to miss.
+
+A behavior observed in the Preview 7 build was traced to a fix that shipped in **Preview 3**. It was
+real, it was verifiable, and documenting it in the Preview 7 notes would have been wrong. Anything
+still present from an earlier release will verify perfectly against the current build.
+
+The rule: **every documented change must trace to a PR in this milestone**, not merely be observably
```
danroth27: Saying "in this milestone" might be misinterpreted to mean a corresponding GitHub milestone, but not all repos use milestones for each preview release.

```suggestion
The rule: **every documented change must trace to a PR associated with this milestone**, not merely be observably
```

.github/skills/release-notes/references/component-mapping.md
```diff
@@ -24,6 +24,23 @@ Each release notes file gets its own per-component branch named `release-notes/{
 
 The agent assigns each component PR to its default assignee(s) when opening the PR (`gh pr create --assignee ...`), so the right team sees it in their review queue.
 
+### Check that the owners are still current
+
+This table is maintained by hand and goes stale as people change roles. In a single milestone, three
+of the rows were wrong: one owner had retired, one had moved off the component, and one component had
+been handed to someone new.
+
+Before opening the PRs, confirm each assignee can actually be assigned:
```
danroth27:
```suggestion
This table is maintained by hand and goes stale as people change roles. Before opening the PRs, confirm each assignee can actually be assigned:
```

.github/skills/release-notes/references/format-template.md
```diff
@@ -89,6 +89,34 @@ Known component docs links:
 4. **Feature ordering** — highest customer impact first
 5. **Breaking changes near the end** — low-score entries with `breaking_changes: true` usually belong in a short section before Bug fixes, not as full feature sections
 6. **Preview feature callout** — when a feature is listed in `release-notes/features.json`, start its section with the standard blockquote callout from that file
+7. **Call out what breaks an upgrade from the previous preview** — see below
```
danroth27:
```suggestion
7. **Preview upgrade instructions** — see below
```

.github/skills/release-notes/references/format-template.md
```diff
@@ -89,6 +89,34 @@ Known component docs links:
 4. **Feature ordering** — highest customer impact first
 5. **Breaking changes near the end** — low-score entries with `breaking_changes: true` usually belong in a short section before Bug fixes, not as full feature sections
```
danroth27: Do these breaking changes include preview-to-preview breaking changes? Or do they only include version-to-version breaking changes? If they include preview-to-preview breaking changes, then the preview upgrade instructions should probably be covered here.

.github/skills/validate-code-samples/SKILL.md
```diff
@@ -0,0 +1,143 @@
...
+| Failure | What static verification reported | What running it showed |
+| ------- | --------------------------------- | ---------------------- |
+| `Blazor.pause.waitFor(fn)` | Nothing - JS surface is invisible to `dotnet-inspect` | The API does not exist. The served `blazor.web.js` exposes only `pauseCircuit` / `resumeCircuit`; the real hook is a circuit handler callback |
+| `.RequireAntiforgeryToken()` | Nothing - inferred from a PR title, never compiled | Does not exist. A reflection sweep of the shipped assembly found only `DisableAntiforgery` |
+| `EnableClientValidation` | Member found, so the draft looked verified | Renamed to `DisableClientValidation` **and the polarity inverted**. A name-only check actively confirmed the wrong claim |
+| Documented sample page | Built cleanly | Returned HTTP 500 - the attribute silently requires a *public* property |
+
+The pattern: **a name is not a behavior.** Static verification answers "does this symbol exist",
+which is a much weaker question than "does this documented claim hold".
+
```
danroth27: We don't need this historical context in the skill.

```suggestion
```

.github/skills/generate-changes/SKILL.md
danroth27: Added as a rule under `Rules for using it`: milestone membership is a hand-applied label, so it shows someone *intended* the PR for this release, not that the change shipped in it. A PR can be milestoned and then reverted, retargeted to a later preview, or milestoned in error. Anything found via the sweep now has to be confirmed in `changes.json` and in the build before it's written up, with links to `api-verification.md` and `validate-code-samples`.

.github/skills/release-notes/references/component-mapping.md
danroth27: Applied.

.github/skills/release-notes/references/api-verification.md
danroth27: Applied. Also changed the same phrasing in bullet 3 of that section for consistency, since it had the identical ambiguity.

.github/skills/generate-changes/SKILL.md
danroth27: Applied. Dropped the measured table in favor of the plain repo list — the counts were a point-in-time measurement that would go stale, and the list is the part that's actually actionable.

.github/skills/validate-code-samples/SKILL.md
danroth27: Removed. Checked first that nothing was lost — the paragraph immediately above already states the substance (static verification can't see JavaScript APIs, can't tell you a default value, and can't tell you whether a documented sequence of calls works), so deleting the table doesn't drop any guidance.

.github/skills/release-notes/references/format-template.md
danroth27: Good question, and checking it changed the answer. They **do** include preview-to-preview breaking changes.

I verified against the existing notes rather than assuming. The Preview 4 ASP.NET Core breaking changes section lists [dotnet/aspnetcore #65930](https://github.com/dotnet/aspnetcore/pull/65930) (`%2F` preserved in absolute-form request targets) and [dotnet/aspnetcore #65808](https://github.com/dotnet/aspnetcore/pull/65808) (`TlsClientHelloBytesCallback` obsoleted) — both milestoned `11.0-preview4` and merged during that preview. So each preview's Breaking changes section is already scoped to what changed *in that milestone*, which is exactly the preview-to-preview upgrade story.

So I folded the upgrade guidance into the Breaking changes section instead of keeping a parallel one, and rule 5 now points at it. Two notes on the merge:

- The section now says explicitly to write it for the reader upgrading from the previous preview, rather than as an abstract list of incompatibilities with the last GA.
- **Removed or replaced workarounds** was the one item that didn't fit — "your workaround is no longer needed" is upgrade-relevant but isn't a breaking change. I pulled it out of the list and it now says to call it out with the fix in Bug fixes.

This supersedes your suggestion on rule 7 in the comment just above, since rule 7 is gone. Happy to restore a separate section if you'd rather keep the upgrade guidance distinct from breaking changes.

.github/skills/release-notes/references/format-template.md
danroth27: Superseded by the fold described in the reply on rule 5's comment below — rule 7 no longer exists, since the preview upgrade guidance is now part of the Breaking changes section. Rule 5 carries the pointer instead. Let me know if you'd prefer a separate section after all and I'll restore this with your wording.

.github/skills/generate-changes/SKILL.md
danroth27: Corrected — thanks, I had the mechanism wrong. I'd written that milestone membership is "a label applied by hand", which overstates it and would push people toward treating every milestoned PR as suspect.

Now reads: milestones are applied by automation, so they are usually right, but they can be changed or applied incorrectly by hand afterwards, and a milestoned PR can still be reverted. A milestone is reliable evidence of *where to look* and weak evidence that the change *shipped*. The rule ends with "the expected outcome is that it checks out; the point is to catch the occasional one that does not", so the verification step reads as confirming the expected result rather than expecting a problem.

Checked the other skill files for the same mischaracterization — the only remaining "maintained by hand" is in `component-mapping.md`, which is accurate since that owner table really is hand-maintained.

danroth27: Re-audited this before resolving. I made one further wording correction: because milestones are applied by automation and are usually right, the skill now calls them "strong but not conclusive evidence" that a change shipped, rather than "weak evidence." The independent changes.json and build checks remain required to catch the occasional incorrect assignment or revert.

.github/skills/release-notes/references/api-verification.md
```diff
@@ -116,6 +116,32 @@ You do NOT need to verify:
...
+The rule: **every documented change must trace to a PR associated with this milestone**, not merely be observably
+true in the build.
+
+1. **`changes.json` membership is the primary test.** If a change has no entry, it did not flow into
+   this milestone. Confirm before writing it up.
+2. **Check the PR's merge date and milestone** where the repo maintains them —
+   `dotnet/aspnetcore`, `dotnet/runtime`, `dotnet/sdk`, and `dotnet/efcore` do;
+   `dotnet/roslyn` and `dotnet/razor` do not (see
+   [`generate-changes`](../../generate-changes/SKILL.md)). For repos without milestones,
+   `changes.json` membership is the only available provenance signal.
```
jeffhandley, line 137: We're starting to scatter info about different repos throughout all these instructions. I still favor the original approach I'd taken with the release-notes skill where the top-level skill was ignorant about any repo-specific processes, and we had optional extensibility for each repo to provide its repo-specific process info.

I don't think we should hold this PR for that feedback, and I know that @richlander still has open work for more comprehensive changes to this whole process--I'd like to see the next round of changes incorporate that kind of separation of concerns.

.github/skills/release-notes/references/pr-layout.md
```diff
@@ -15,15 +15,50 @@ The set of components and their release notes files is defined in [`component-ma
...
+
+`gh pr edit` fails against `dotnet/core` with a Projects (classic) GraphQL deprecation error and
+leaves the PR unchanged. This affects the whole command, not just one flag — `--add-assignee`,
+`--remove-assignee`, and `--body` / `--body-file` all fail the same way. It exits non-zero, but the
+error text is about Projects rather than about what you were trying to change, so it is easy to
+mistake for a warning. Use the REST endpoints:
+
+```bash
+gh api -X POST   repos/dotnet/core/issues/{number}/assignees -f "assignees[]=<user>"
+gh api -X DELETE repos/dotnet/core/issues/{number}/assignees -f "assignees[]=<user>"
```
jeffhandley, line 43: This is surprising to me. I assume this is based on the evidence you saw.

.github/skills/validate-code-samples/SKILL.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+### Other sources
+
+- **<https://github.com/dotnet/dotnet/blob/main/docs/builds-table.md>** — the latest-build table in
+  the VMR repo. Use it when there is no `build-metadata.json` yet, or to sanity-check that the
+  version you derived is really the current build. It also documents the `dotnet11` NuGet feed
+  needed for runtime packs in self-contained scenarios.
+- **<https://release.dot.net>** — the human-facing list of builds the team installs from. Fine for a
+  person, but prefer a URL derived from `build-metadata.json` for anything scripted.
```
jeffhandley: https://release.dot.net/ requires authentication - I don't think we can guide the skill to use it at all, and I recommend we leave it out of the skill entirely so that there's no chance of it being referenced in any way.

.github/skills/release-notes/references/api-verification.md
danroth27: Agreed. I reviewed the current release-notes workflow for repository-specific logic and will report the findings separately. Per this feedback, I am not holding this PR or attempting the separation here; a follow-up redesign should move repository policies behind an extension point instead of continuing to distribute them through the shared workflow.

.github/skills/release-notes/references/pr-layout.md
danroth27: Confirmed directly today with a no-op edit: `gh pr edit 10516 --repo dotnet/core --title <current-title>` exits 1 with the Projects (classic) `repository.pullRequest.projectCards` GraphQL error and leaves the PR unchanged. Since this workflow needs assignee and body edits, I kept the REST workaround.

.github/skills/validate-code-samples/SKILL.md
danroth27: Removed the `release.dot.net` reference entirely. The skill now starts from the Installing the SDK section in `dotnet/sdk`, follows its latest-builds table, and selects the column matching the milestone feature band or preview branch.

jeffhandley (approved):

PR #10535, opened 2026-08-20
Document August 2026 WPF known issue

rbhanda:
## Summary

- document the August 2026 WPF printing and PDF/XPS known issue for .NET 8, 9, and 10
- provide the temporary `runtimeconfig.json` AppContext switch workaround
- warn that the workaround disables protections introduced in the August 2026 update

## Validation

- markdownlint passes for each newly added section
- markdown link checks pass for all three changed files
- diff whitespace check passes with the existing CRLF format in the .NET 10 file

### DIFF
--- a/release-notes/10.0/known-issues.md
+++ b/release-notes/10.0/known-issues.md
@@ -2,6 +2,28 @@
 
 You may encounter some known issues, which may include workarounds, mitigations, or expected resolution timeframes. Watch this space for any known issues in .NET 10.0.
 
+## WPF printing and PDF/XPS generation may fail with certain fonts
+
+After installing the August 2026 .NET update, some WPF applications may fail with a `System.IO.FileFormatException` when printing or generating PDF/XPS content that uses certain fonts, including Calibri.
+
+### Workaround for WPF printing and PDF/XPS generation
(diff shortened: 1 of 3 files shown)


ToreDemant:
Are there a similar entry for .Net Framework that has same issue after the August update?

TaraOverfield:
> Are there a similar entry for .Net Framework that has same issue after the August update?

Yes, .NET and .NET Framework are both impacted by the same issue.  We're both publishing known issues with nearly the same message.  The only difference is the opt out mechanism which applies specifically to the product being used.

ToreDemant:
> > Are there a similar entry for .Net Framework that has same issue after the August update?
> 
> Yes, .NET and .NET Framework are both impacted by the same issue. We're both publishing known issues with nearly the same message. The only difference is the opt out mechanism which applies specifically to the product being used.

Found it:
https://learn.microsoft.com/en-us/dotnet/framework/release-notes/2026/08-11-august-cumulative-update

ToreDemant:
@rbhanda / @TaraOverfield shouldn't this PR be completed so that the knowledge becomes available for all?

PR #10542, opened 2026-08-28
[release-notes] .NET Libraries in .NET 11 RC 1

danroth27:
## Summary

Add the .NET Libraries release notes for .NET 11 RC 1. This component PR targets the RC 1 base branch so the component team can review and refine it independently.

The RC 1 release branch is still moving; the base metadata and this note will be refreshed against the final tag before publication.


tannergooding:
This breaking change actually shipped in P7, but the doc was only just made recently due to a customer report, so likely worth calling out in the release notes: https://github.com/dotnet/docs/issues/55743

PR #10550, opened 2026-08-28
[release-notes] Windows Forms in .NET 11 RC 1

danroth27:
## Summary

Add the Windows Forms release notes for .NET 11 RC 1. This component PR targets the RC 1 base branch so the component team can review and refine it independently.

The RC 1 release branch is still moving; the base metadata and this note will be refreshed against the final tag before publication.


KlausLoeffelmann (approved):
I...I have nothing to add.
It's spot on!

PR #10565, opened 2026-09-08
Add Blazor AI components to .NET 11 RC1 notes

danroth27:
## Summary

- Add the complete `Microsoft.AspNetCore.Components.AI` feature coverage back to the .NET 11 RC1 ASP.NET Core release notes after #10566 removes it temporarily.
- Document package installation, streaming chat, structured rich text, client and server tool rendering, generated typed handlers, approval flows, persistent conversation threads, activities, and typed, shared, and predictive UI state.

## Package and runtime validation

The package was used directly in `danroth27/AgenticUI` on .NET 11 RC1 after removing the vendored Components.AI source. Browser exercises covered streaming chat, rich-text snapshots, generated typed tool blocks, UI actions, approval and rejection, conversation restoration, typed/shared state, activity updates, and predictive-state acceptance and rollback.

The testing also confirmed that custom renderers belong in `ChatPage.MessageListContent`, generated tool handlers require `options.AddGeneratedToolBlocks()`, conversation restoration is explicit through `RestoreAsync`, and unresolved predictive state rolls back when the turn ends.

## Blocked on package publication

Keep this PR in draft until the intended package is available from the shipping feed and signed. Before marking it ready, update the package version if necessary and repeat the clean-cache restore and signature checks.

This PR is stacked on #10566 so its diff represents the complete release-note section that will be restored.

PR #10566, opened 2026-09-08
Remove Blazor AI components from .NET 11 RC1 notes

danroth27:
## Summary

- Remove the `Microsoft.AspNetCore.Components.AI` feature section from the .NET 11 RC1 ASP.NET Core release notes.
- Remove the corresponding table-of-contents entry and screenshot.

The package version intended for RC1 is not yet available on the shipping `dotnet12` feed. The full, package-validated release notes are preserved in #10565 and can be restored after the package is promoted and signed.


danroth27:
@guardrex We've hit an issue with getting the Components.AI package published, so I'm removing the related release notes for now. Once we get the package sorted, I have #10565 queued up to add back the content.

PR #10571, opened 2026-09-08
Remove unavailable affected-test workflow from RC1 notes

danroth27:
## Summary

Remove the affected-test workflow from the .NET 11 RC1 SDK release notes. The SDK includes experimental command-line orchestration, but the separately distributed Microsoft.Testing.Platform extension and its storage contract are not available for RC1, so the documented commands do not provide a usable end-to-end workflow.

The SDK work remains recorded in the filtered-features comment for future release-note generation.

## Validation

- verified a normal Microsoft.Testing.Platform test app passes with .NET SDK `11.0.100-rc.1.26427.112`
- verified the SDK exposes `--collect-test-map` and `--affected-tests` only when `DOTNET_CLI_ENABLE_AFFECTED_TESTS` is enabled
- verified both commands fail with `Unknown option` when executed without the unpublished extension
- confirmed the microsoft/testfx rollout documentation says the extension package and public storage contract are not available yet
- ran Markdown linting, link checking, and `git diff --check`

### DIFF
--- a/release-notes/11.0/preview/rc1/sdk.md
+++ b/release-notes/11.0/preview/rc1/sdk.md
@@ -100,20 +100,6 @@ option, and `--no-banner` appears in the command help
 ([dotnet/sdk #55376](https://github.com/dotnet/sdk/pull/55376) and
 [dotnet/sdk #55412](https://github.com/dotnet/sdk/pull/55412)).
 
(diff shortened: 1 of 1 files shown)


baronfel (approved):
Rationale is solid 👍

PR #10580, opened 2026-09-13
Fix table formatting in 5.0-supported-os.md

Linlccc:
Corrected formatting in the supported OS table for Linux.

before
image

after
image


mairaw (approved):
Good catch. Thank you @Linlccc

